PR #27, opened 2026-03-24
Build and publish nanoidp docker images to GitHub Container Registry

nicholasdehnen:
**Summary**: This PR adds a workflow which builds the nanoidp Dockerfile and publishes the container to the GitHub Container Registry.

Additionally, I changed the Dockerfile slightly to install and use curl for the healthcheck. The current healthcheck fails when using Podman, which runs CMD through `/bin/sh -c` and therefore breaks on the semicolon. There may be ways around that using `__import__(..)` magic, but I figured switching to curl was worth it as it also gets rid of the python interpreter [startup overhead](https://pythondev.readthedocs.io/startup_time.html) on every healthcheck, and is a reasonably small addition (~450kB).

**Why?** I would like to use NanoIDP in my compose files without having to clone the repo and build the Docker images myself.

**See also:**

- Workflow runs: <https://github.com/nicholasdehnen/nanoidp/actions>
- Container registry: <https://github.com/nicholasdehnen/nanoidp/pkgs/container/nanoidp>

Thanks 🙂

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,43 @@
+name: Publish to GHCR
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  publish-ghcr:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Docker Buildx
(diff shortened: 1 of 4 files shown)


cdelmonte-zg (approved):
Thanks for this clean and well-structured contribution, @nicholasdehnen! The GHCR workflow, the curl healthcheck fix, and the multi-arch support are all great additions. 🙏

PR #29, opened 2026-03-26
fix: allow authentication flow without authentication headers

guenhter:
When authentication-code flow is used with PKCE, libraries like authlib are not sending the Authentication header, as there is no secret to send. The client_id in this case is also usually available in the body, not in the header.

For that reason, the auth header validation is disabled in this particular case to make the PKCE authentication flow work.

I've successfully tested this fix with authlib manually.

### DIFF
--- a/src/nanoidp/routes/oauth.py
+++ b/src/nanoidp/routes/oauth.py
@@ -262,9 +262,25 @@ def token():
     audit = get_audit_log()
     req_info = _get_request_info()
 
-    # Check client authentication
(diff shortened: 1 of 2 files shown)


cdelmonte-zg (approved):
Great fix, thanks @guenhter! PKCE public client flow without auth header is correctly handled and well tested. A couple of minor things we might improve later (handling missing client_id in body, client_id mismatch between header and body) but nothing blocking. Merging!

PR #30, opened 2026-03-26
chore: handle token endpoint edge cases

guenhter:
Follow up of https://github.com/cdelmonte-zg/nanoidp/pull/29


cdelmonte-zg:
Clean follow-up, thanks @guenhter! Both edge cases handled correctly with good test coverage.

PR #31, opened 2026-04-27
feat: support env vars in config files

### LINKED ISSUE #28: Port config
Hi,

first of all: thx for this project. I love it as for testing this is exactly what I was looking for.

I'm keen also contributing here, but before I do this, I'd like to clarify a few things.
In this case, it is about port config:

if the application is started with a new different port like 8001, it is not sufficient to just make "--port 8001", as the `curl http://localhost:8005/.well-known/openid-configuration` still returns port `8000` everywhere.  That is obviously because of the port config in the config as well.

Also, when using docker this is another challenge: `docker container run -p 8001:8000 ...` so when ports are mapped, obviously port `8001` is nowhere configured. 

It would be nice, if the nanoidp just check on which port it was called on, and then the discovery endpoint is then called, it just returns the port where it is called from, and not the port configured.

What do you think?

guenhter:
This allows to use environment variables in the config file values like this:

`settings.yaml`

```
server:
  port: ${PORT:8000}
  ...
oauth:
  issuer: http://localhost:${PORT:8000}
...
```

* At load time of the configs, the content is checked for `${...}` patterns and replaced with the actual value.

Fixes https://github.com/cdelmonte-zg/nanoidp/issues/28

---

mainly written by AI (Claude Sonnet 4.6) in multiple iterations based on (human) review comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Environment variable substitution in `settings.yaml` using `${NAME}` / `${NAME:default}` syntax
+- `PORT` env var honoured in the Docker image via shell expansion in `CMD`
+
 ## [1.3.3] - 2026-04-22
   
 ### Fixed
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,13 +35,14 @@ RUN mkdir -p /app/keys
 # Environment variables
 ENV PYTHONUNBUFFERED=1
 ENV NANOIDP_CONFIG_DIR=/app/config
+ENV PORT=8000
 
-# Expose port
-EXPOSE 8000
+# Expose port (default 8000; override PORT env var for non-standard ports)
+EXPOSE ${PORT}
 
 # Health check
 HEALTHCHECK --interval=30s --timeout=3s --start-period=5s --retries=3 \
-    CMD curl -fsSL http://localhost:8000/api/health || exit 1
+    CMD curl -fsSL http://localhost:${PORT}/api/health || exit 1
 
-# Run the application
-CMD ["nanoidp", "--host", "0.0.0.0", "--port", "8000"]
+# Run the application (PORT env var is passed to nanoidp; defaults to 8000)
+CMD ["sh", "-c", "exec nanoidp --host 0.0.0.0 --port ${PORT}"]
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -1,9 +1,9 @@
 server:
   host: 0.0.0.0
-  port: 8000
+  port: ${PORT:8000}
   debug: false
 oauth:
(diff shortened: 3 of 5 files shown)


cdelmonte-zg (changes requested):
Thanks for picking this up, @guenhter — nice clean implementation and the test coverage in `test_expand_env_vars.py` is genuinely thorough (love the pattern-boundary cases). A few changes I'd like to see before we merge; should be quick to address.

### 1. Scope: `_expand_env_vars` should not run on `users.yaml` (or, if it does, please warn on unresolved vars)

The PR currently calls `_expand_env_vars` in both `_load_settings` (config.py:221) and `_load_users` (config.py:300). Issue #28 only asks for the port case, so widening to `users.yaml` is scope creep — and combined with the `${NAME}` → `""` fallback, it's a real foot-gun: if someone writes `password: ${ADMIN_PWD}` and forgets to export `ADMIN_PWD`, the user ends up with an empty password silently.

Two acceptable resolutions:
- **Preferred:** drop the call from `_load_users` and keep the feature limited to `settings.yaml`. The CHANGELOG entry should then be narrowed accordingly.
- **Alternative:** if you want to keep credential-from-env as an explicit feature, document it in the README and emit `logger.warning("env var %s not set and no default provided", name)` in `_expand_env_vars` whenever a `${NAME}` (no default) resolves to empty. Silent empty-string is the dangerous part, not the feature itself.

### 2. Dockerfile `CMD` lost exec form

The change to:
```
CMD nanoidp --host 0.0.0.0 --port ${PORT}
```
moves us from exec form to shell form, which means `nanoidp` no longer runs as PID 1 and `SIGTERM` from `docker stop` won't propagate to it cleanly (you'll wait for the 10s grace period and then get SIGKILL'd). Suggested fix that keeps env-var expansion *and* exec semantics:
```
CMD ["sh", "-c", "exec nanoidp --host 0.0.0.0 --port ${PORT:-8000}"]
```
The `exec` is what hands PID 1 over to `nanoidp`, and `${PORT:-8000}` keeps a sane default if someone runs the image without setting it.

### 3. PR description: `PORT` is not a special variable

Small thing, but the description says "`PORT` is treated as a special variable". It isn't — it's just shell expansion in the Dockerfile and a regular `${PORT:8000}` substitution in `settings.yaml`. No Python code treats `PORT` differently from any other env var. Could you tweak the wording so future readers don't go hunting for special-case logic that doesn't exist?

---

Nothing here is a technical blocker — implementation is solid and the regex/test work is good. Mostly scope and a couple of safety nits. Happy to re-review as soon as these are in.

guenhter:
Oh, true. I'll change that in a second. Thx

guenhter:
Done.

I don't use `${PORT:-8000}` in the Dockerfile as it anyway has the `ENV PORT=8000` line which defaults then to the port `8000` if not defined.

cdelmonte-zg (approved):
All three points addressed cleanly — scope correctly narrowed to settings.yaml, Dockerfile CMD restored to exec semantics via sh -c, description tightened. Verified locally: unit tests green, end-to-end scenarios (default fallback, PORT override, and the security regression check that users.yaml no longer expands env vars) all behave as expected. Nice work, thanks!

guenhter:
Appreciate your effort. Thx for merging

Issue #72, opened 2026-07-07
Cover wizard.py (currently 0%)


cdelmonte-zg:
`wizard.py` is the only module with zero test coverage (87/87 lines uncovered); `services/yaml_writer.py` is the runner-up at 60%. Both cap how far the CI coverage threshold (#71) can be ratcheted.

The wizard is interactive, so tests need its prompts driven programmatically (monkeypatched `input`/`prompt` or extracting the question flow from the I/O). Acceptance: wizard-generated config round-trips through `Settings`/users loading, and the threshold in #71 gets a bump in the same PR.

PR #74, opened 2026-07-07
ci: enforce coverage threshold and fail on Codecov upload errors

### LINKED ISSUE #71: Enforce a coverage threshold in CI and fail on Codecov upload errors
Coverage is measured in CI (3.12 leg) but nothing acts on it: no `--cov-fail-under`, and the Codecov upload has `fail_ci_if_error: false`, so a silent upload failure looks green.

Measured today: **73%** total (662 tests).

## Proposal

- `pytest --cov=nanoidp --cov-fail-under=70` in the coverage step — 3 points of headroom below current reality, so the gate catches regressions without flaking; ratchet upward as coverage improves.
- `fail_ci_if_error: true` on the codecov action.

Threshold ratcheting works like the mypy ratchet: the number only goes up.

cdelmonte-zg:
Closes #71 — first execution PR of the [CI quality gates](https://github.com/cdelmonte-zg/nanoidp/milestone/4) milestone.

- `--cov-fail-under=70` on the 3.12 coverage step. Measured today: 72.88% across 662 tests, so the gate has ~3 points of headroom against flakiness while still catching real regressions. Per the issue, the threshold only ratchets upward (a comment in the workflow says so); #72 (covering `wizard.py`, currently 0%) is the natural next bump.
- ~~`fail_ci_if_error: true` on the Codecov upload~~ → **the Codecov upload step is removed.** Turning the flag on surfaced that the upload has failed on every run since inception ("Token required - not valid tokenless upload"): the repo has no `CODECOV_TOKEN` secret, no `codecov.yml`, no badge — Codecov was never set up, and `fail_ci_if_error: false` hid that. A step that pretends to publish coverage nowhere is exactly what #71 exists to stop. Coverage enforcement now lives entirely in CI; if dashboards are ever wanted, re-add the action with a real token. (Details in the PR comment.)

Verified locally: the full suite passes the gate (`Required test coverage of 70% reached`).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,11 +42,12 @@ jobs:
       - name: Run tests with coverage
         if: matrix.python-version == '3.12'
         run: |
-          pytest tests/ --cov=nanoidp --cov-report=xml --cov-report=term-missing
+          # Threshold ratchets upward only (#71); measured 73% when introduced.
+          pytest tests/ --cov=nanoidp --cov-report=xml --cov-report=term-missing --cov-fail-under=70
 
       - name: Upload coverage to Codecov
         if: matrix.python-version == '3.12'
         uses: codecov/codecov-action@v4
         with:
           files: ./coverage.xml
-          fail_ci_if_error: false
+          fail_ci_if_error: true
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,11 +43,4 @@ jobs:
         if: matrix.python-version == '3.12'
         run: |
           # Threshold ratchets upward only (#71); measured 73% when introduced.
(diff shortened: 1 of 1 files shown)


cdelmonte-zg:
CI surfaced something better than what the PR set out to do: with `fail_ci_if_error: true` on, the Codecov upload failed with **"Token required - not valid tokenless upload"** — and a check of the repo shows no `CODECOV_TOKEN` secret, no `codecov.yml`, no badge. The upload has failed silently on **every run since inception**; `fail_ci_if_error: false` was hiding it.

Rather than wiring up a service that was never actually configured, the second commit removes the dead upload step entirely — `--cov-fail-under=70` in CI is the gate, with no external dependency. If Codecov dashboards are ever wanted, it's an account on codecov.io + a `CODECOV_TOKEN` secret + re-adding the action (with `fail_ci_if_error: true`, this time for real).

PR #77, opened 2026-07-07
mypy ratchet tranche 3: src/ fully annotated, flag flipped globally

### LINKED ISSUE #70: Ratchet mypy to disallow_untyped_defs module by module until src/ is clean
Follow-up to the gradual baseline from #55. Measured today (mypy 2.2.0-era config, `--disallow-untyped-defs`): **83 errors in 13 of 19 modules** — 6 modules already pass.

| Module | Errors |
|---|---|
| `routes/ui.py` | 23 |
| `config.py` | 18 |
| `routes/oauth.py` | 11 |
| `routes/api.py` | 11 |
| `services/crypto.py` | 9 |
| `routes/saml.py` | 9 |
| `services/auth_code.py` | 4 |
| `services/audit.py` | 4 |
| `mcp_server.py` | 3 |
| `app.py` | 3 |
| `wizard.py` | 2 |
| `services/token.py` | 2 |
| `__main__.py` | 2 |

## Mechanism

Per-module `[[tool.mypy.overrides]]` with `disallow_untyped_defs = true`, growing the list as modules are annotated — a module added to the list never leaves it (ratchet). When all of `src/` is in, flip the flag globally and drop the overrides.

## Suggested tranches

1. The already-passing 6 modules + the small tail (`__main__`, `token`, `wizard`, `app`, `mcp_server`, `audit`, `auth_code` — 20 errors total)
2. `crypto`, `saml`, `oauth`, `api` (40 errors)
3. `config.py`, `routes/ui.py` (41 errors)

Annotations only — no behavioral changes; any real bug mypy uncovers gets its own issue (that's how the stdio-entrypoint crash was found in #56).

cdelmonte-zg:
Closes #70 — final tranche, completing the [Typing strictness](https://github.com/cdelmonte-zg/nanoidp/milestone/3) milestone.

- Annotates the last 32 signatures: the nine `config.py` loaders/savers (`-> None`, all of them), and the 23 `routes/ui.py` views (`-> ResponseReturnValue`, same as the other route modules).
- Flips `disallow_untyped_defs = true` **globally** in `[tool.mypy]` and deletes the per-module ratchet override — the ratchet's end state per the issue. Any new unannotated def anywhere in `src/` now fails CI.

Unlike tranche 2, no cascade fixes were needed: `check_untyped_defs` had already been checking these bodies, so the annotations were purely declarative.

Verified: `mypy src` clean (19 files), `ruff check .` clean, 662 tests pass, coverage 73% ≥ the 70% gate.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,40 +106,15 @@ ignore = [
 ]
 
 [tool.mypy]
-# Gradual-adoption baseline (#55): catch real type errors in src/ without
-# demanding full annotation coverage yet. Tighten over time (e.g. enable
-# disallow_untyped_defs per module) as annotations improve.
+# Baseline from #55, ratcheted to full annotation coverage in #70: every def
+# in src/ is annotated and stays that way.
 python_version = "3.10"
 files = ["src"]
+disallow_untyped_defs = true
 check_untyped_defs = true
 warn_redundant_casts = true
 warn_unused_ignores = true
 no_implicit_optional = true
-
-[[tool.mypy.overrides]]
-# Typing ratchet (#70): modules here require fully annotated defs. The list
-# only grows; when all of src/ is in, flip disallow_untyped_defs globally
-# and delete this block. Still out: config, routes.ui.
-module = [
-    "nanoidp",
-    "nanoidp.__main__",
-    "nanoidp.app",
-    "nanoidp.exceptions",
-    "nanoidp.mcp_server",
-    "nanoidp.wizard",
-    "nanoidp.routes",
-    "nanoidp.routes.api",
-    "nanoidp.routes.oauth",
-    "nanoidp.routes.saml",
-    "nanoidp.services",
-    "nanoidp.services.audit",
-    "nanoidp.services.auth_code",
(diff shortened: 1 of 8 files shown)


cdelmonte-zg:
Both taken, generalized to all of `src/` (same symmetry argument):

- `_parse_audiences -> list[str]`, and with it the other two bare generics a grep turned up: `_check_admin_secret(arguments: dict[str, Any])` in `mcp_server.py` and `_device_codes: dict[str, Any]` in `routes/oauth.py` (the values are heterogeneous — grant-state dicts plus `"user:<code>" -> device_code` back-references — so explicit `Any` is the honest type; a comment now says so).
- `-> None` on **all 27** `__init__` that mypy accepted thanks to annotated parameters (`ConfigManager` included; the others: 24 in `exceptions.py`, `YamlWriter`, `AuditLog`).

mypy/ruff/662 tests green again. After this, `grep` finds no bare generics and no defs without a return type in `src/` — if `disallow_any_generics` is ever wanted, the ground is ready.

Issue #68, opened 2026-07-07
oauth21 security profile: PKCE required, S256-only, rotation on, no password grant


cdelmonte-zg:
The building blocks shipped in 2.2.0: `require_pkce` (#47), `refresh_token_rotation` (#46), S256-only + plain rejection in `stricter-dev`, refresh token `client_id` binding (#56). What's missing is the profile that bundles them into "nanoidp at its strictest, aligned with draft OAuth 2.1" (draft-ietf-oauth-v2-1-15).

## Proposal

A third profile, `--profile oauth21` (alongside `dev` and `stricter-dev`), which enforces:

- PKCE required on the authorization code flow, S256 only (§4.1.1, §7.5.2)
- Refresh token rotation on (§4.3.1; RFC 9700 §4.14.2 family revocation already implemented)
- `password` grant disabled — OAuth 2.1 removes it (it does not appear in the draft's grant types)
- Registered `redirect_uris` mandatory for clients used with `/authorize`, exact matching (#67 is the prerequisite)

**Discovery must reflect the active profile** (principle 2: metadata never lies): `grant_types_supported`, `code_challenge_methods_supported`, and anything else profile-dependent are emitted per-profile via the shared builder in `services/discovery.py` — the same mechanism that already switches `code_challenge_methods_supported` for `stricter-dev`.

Open question to settle in the PR: whether `oauth21` implies the `stricter-dev` runtime hardening (bcrypt, CORS, rate limiting) or stays orthogonal to it — leaning orthogonal, since 2.1 is about protocol behavior, not deployment hardening.

Per principle 5: MCP exposure (`get_settings` reporting the profile), e2e coverage, and docs in the same PR.

Depends on #67.

Issue #35, opened 2026-05-25
Friendlier handling of malformed additional_audiences in settings.yaml


cdelmonte-zg:
Follow-up from the code review of #32.

### Problem
Config load passes `client_data.get("additional_audiences", [])` straight into the Pydantic `OAuthClient` model (`List[str]`). A natural single-value form like:

```yaml
additional_audiences: api://my-api   # scalar instead of a list
```

or non-string items (`additional_audiences: [123]`) raise a Pydantic `ValidationError`. `_load_settings` is not wrapped in try/except, so `init_config()` propagates and the **server fails to start / reload**.

This is the same failure class as other mistyped settings, but the new list field is an easy footgun (people will write a single value without a list).

### Options
- Coerce a scalar string into a single-element list at load, and/or
- Catch and surface a clear, actionable error (which client, which field) instead of a raw stack trace at startup.

### Refs
`src/nanoidp/config.py` (`_load_settings`, ~line 241). Note the MCP path already validates via `_normalize_audiences`; YAML load should be at least as friendly.

cdelmonte-zg:
Addressed in #38 (branch `fix/issue-32-followups`).

PR #78, opened 2026-07-07
feat: registered redirect_uris with exact matching on /authorize

### LINKED ISSUE #67: Client redirect_uris registration with exact matching on /authorize
Today `/authorize` accepts any syntactically valid http(s) `redirect_uri` (see the comment in `routes/oauth.py` — "basic validation - in production should check against registered URIs"), and clients in `settings.yaml` have no `redirect_uris` field at all.

OAuth 2.1 (draft-ietf-oauth-v2-1 §1.4.1, following RFC 9700 §2.1) requires redirect URIs to be compared using **exact string matching** against pre-registered values. This is also a behavior clients genuinely need to test (a mis-registered callback is one of the most common integration failures).

## Proposal

- Optional `redirect_uris: [...]` list on each client in `settings.yaml` (web UI + MCP `create_client`/`update_client` too).
- When a client has registered URIs, `/authorize` enforces exact string match and rejects mismatches with `invalid_request` **without redirecting** (RFC 6749 §3.1.2.4: never redirect to an unvalidated URI).
- When the list is absent, current permissive behavior stays (principle 3: hardening is opt-in). The upcoming `oauth21` profile will require registration.

Per principle 5: MCP exposure, `examples/test_agent.py` coverage, and docs land in the same PR.

cdelmonte-zg:
Closes #67 — first feature PR of the [OAuth 2.1 profile](https://github.com/cdelmonte-zg/nanoidp/milestone/1) milestone, and the prerequisite for #68.

## Behavior

- `OAuthClient` gains an optional `redirect_uris: [...]` list. When **non-empty**, `/authorize` enforces **exact string matching** (RFC 6749 §3.1.2.3, OAuth 2.1 §4.1.1): no prefix, host, path or trailing-slash normalization.
- A mismatch answers `400 invalid_request` **directly** — per §3.1.2.4 the error is never delivered by redirecting to the unvalidated URI. The rejection is audit-logged.
- When the list is absent/empty, the permissive dev behavior is unchanged (principle 3: hardening is opt-in). The upcoming `oauth21` profile (#68) will make registration mandatory.

## Ships whole (principle 5)

- **Config**: YAML load with the same scalar-string coercion as `additional_audiences` (#35) — the coercer is generalized (`_coerce_client_str_list`), with `_coerce_additional_audiences` kept as a wrapper since tests import it. Round-trips through `ConfigManager.save()` and the YAML writer.
- **Web UI**: textarea on the client form (one URI per line), preserved across edit and secret-regeneration.
- **MCP**: `create_client`/`update_client` schemas + handlers, `redirect_uris` in client dicts; `_normalize_audiences` generalized the same way. Empty list via `update_client` removes the restriction.
- **e2e**: new `test_redirect_uri_exact_matching` in `examples/test_agent.py`, backed by a `registered-client` fixture in the default config (same pattern as `multi-aud-client`).
- **Docs**: the book's configuration reference (example + semantics paragraph).
- **Tests**: 18 new (680 total) — exact match accepted, 7 mismatch shapes (trailing slash, port, scheme, host, path, query, prefix+suffix) rejected without a `Location` header, enforcement on the POST login leg, YAML load/scalar coercion/persistence round-trip, MCP create/update/clear/invalid-type.

## Verified

`mypy src` clean, `ruff` clean, 680 unit tests green, `mdbook build` green — and a **full e2e run against a live server: 41/41**, including the new test (registered URI accepted, `…/callbackevil` rejected with 400 and no redirect, `demo-client` still permissive).

### DIFF
--- a/book/src/reference/configuration.md
+++ b/book/src/reference/configuration.md
@@ -61,6 +61,11 @@ oauth:
       additional_audiences:     # optional; makes the ID Token "aud" an array
         - "https://api.example.com"
         - "urn:service:billing"
+    - client_id: "registered-client"
+      client_secret: "secret"
+      description: "Client whose redirect_uri is pinned"
+      redirect_uris:            # optional; when set, /authorize enforces
+        - "http://localhost:3000/callback"  # exact string matching
 
 saml:
   entity_id: "http://localhost:8000/saml"
@@ -76,6 +81,13 @@ authority_prefixes:
 logging:
   verbose_logging: true  # Include usernames/client_ids in logs (default: true)
 ```
+
+**Registered redirect URIs** — a client with a non-empty `redirect_uris`
+list gets exact-string matching on `/authorize` (RFC 6749 §3.1.2.3,
+OAuth 2.1 §4.1.1): no prefix, host or path normalization, and a mismatch
+is answered with `400 invalid_request` directly — never by redirecting to
+the unvalidated URI (§3.1.2.4). Clients without the field keep accepting
+any syntactically valid URI, the permissive dev default.
 
 The SAML options (`strict_binding`, `sign_responses`, `c14n_algorithm`)
 are covered in detail in [SAML options](saml.md). Security-related
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -19,6 +19,11 @@ oauth:
     additional_audiences:
     - https://api.example.com
     - urn:service:billing
+  - client_id: registered-client
+    client_secret: registered-secret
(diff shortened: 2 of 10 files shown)


cdelmonte-zg:
Good catch — and there is a second native-app aspect it drags along: for loopback redirects RFC 8252 §7.3 requires **ignoring the port** in the match (native apps bind an ephemeral port), so pure exact matching would need that exception on top of the optional netloc for custom schemes (§7.1). Filed as #81 in the OAuth 2.1 milestone, low priority: for the current web-client profile the assumption is documented there, not accidental.

Issue #56, opened 2026-06-11
Review follow-ups: refresh token client binding, atomic rotation + families, implicit-plain PKCE bypass, require_pkce persistence, honest scope response


cdelmonte-zg:
Findings from the owner's review of the 2026-06-11 merge block (#50–#54). Five real issues, in priority order:

1. **Refresh token not bound to its client** — the refresh token carries user/scope/auth_time but not the issuing `client_id`; any other configured client can spend it, and the refreshed ID Token gets the *requesting* client's `aud`, violating RFC 9700 §4.14 and OIDC Core §12.2 (aud must match the original).
2. **Rotation race + no families** — check (`jti in _revoked_tokens`) and revoke (post-issuance `add`) are not atomic: two concurrent refreshes of the same token can both succeed. The "no lock needed" comment on `_revoked_tokens` stopped being true when rotation landed. Also, reuse of a consumed token only rejects that token; RFC 9700 §4.14.2 expects the whole family (including the live descendant) to be revoked.
3. **stricter-dev PKCE `plain` bypass** — `plain` is only rejected when passed explicitly; per RFC 7636 an omitted `code_challenge_method` *defaults to plain*, and the verifier honors that, so `code_challenge=<verifier>` with no method sails through while discovery claims S256-only. Unsupported methods must also be rejected at the authorization endpoint (§4.4.1).
4. **`require_pkce` not persisted** — `update_settings(require_pkce=true)` + `save_config` + `reload_config` silently reverts to false: the field is neither loaded from nor written to settings.yaml (unlike `refresh_token_rotation`).
5. **Token response always says `scope: "openid"`** — hardcoded in `create_token()`, even for non-openid grants and narrowed refreshes; RFC 6749 §5.1 requires reporting the actually granted scope.

Acceptance: fixes for all five with tests covering two-client refresh, concurrent double-refresh (exactly one 200), family revocation on reuse, omitted-method PKCE in stricter-dev, require_pkce save/reload roundtrip, and scope values in responses.

PR #80, opened 2026-07-07
ci: e2e workflow — live-server agent run + MCP stdio smoke test

### LINKED ISSUE #79: CI: end-to-end workflow — live-server agent run + MCP stdio smoke test
CI today runs only pytest. Two real e2e surfaces are never exercised on CI:

1. **HTTP e2e**: `examples/test_agent.py` (41 tests against a live server) is run by hand. Nothing in CI would catch a regression that only shows up over real HTTP (session handling, redirects, content types).
2. **MCP stdio**: nothing anywhere drives the real `nanoidp-mcp` server over stdio — unit tests call `_execute_tool` directly. This is exactly how the stdio entrypoint crashed at startup for its entire existence without a test noticing (found by mypy in #56, not by a test).

## Proposal

A new `e2e.yml` workflow (PRs + pushes to main) with two jobs:

- **http-e2e**: install, start `python -m nanoidp` in the background, poll `/api/health` until ready, run `examples/test_agent.py`; dump the server log on failure.
- **mcp-e2e**: a new `examples/mcp_smoke_test.py` that launches the MCP server as a subprocess via the `mcp` client library, performs the handshake, asserts on `list_tools` (24 tools) and calls a few read-only tools (`list_users`, `get_oidc_discovery`, `get_keys_info`) through the real dispatch path.

The smoke test doubles as a usable example for driving nanoidp's MCP server programmatically.

cdelmonte-zg:
Closes #79 (CI quality gates milestone).

Two new CI jobs covering what pytest structurally cannot:

- **http-e2e** — boots `python -m nanoidp`, polls `/api/health` until ready, runs the full `examples/test_agent.py` (41 checks over real HTTP: every grant, SAML both bindings, key rotation, REST API). Server log dumped on failure.
- **mcp-e2e** — new `examples/mcp_smoke_test.py` launches the real MCP server over **stdio** via the `mcp` client library: handshake, `list_tools` asserted at 24, then `list_users` / `get_oidc_discovery` / `get_keys_info` through the actual dispatch path. This is the regression guard for the class of bug found in #56, where the stdio entrypoint crashed at startup for its entire existence because unit tests bypass the transport via `_execute_tool`. The script doubles as a usable example for driving nanoidp's MCP server programmatically.

Verified locally: smoke test passes (handshake, 24 tools, all calls OK) and a full agent run against a live server is 41/41.

Note for reviewers: independent of #78 — whichever merges first, the other stays green (the agent's new redirect-URI test and its config fixture travel together in #78).

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -13,6 +13,9 @@ on:
 jobs:
   http-e2e:
     runs-on: ubuntu-latest
+    # The agent has per-request timeouts, but a wedged server or a hung
+    # login flow would otherwise burn the 360-minute Actions default.
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v4
 
@@ -45,6 +48,9 @@ jobs:
 
   mcp-e2e:
     runs-on: ubuntu-latest
+    # stdio has no request timeouts: if the server starts but hangs in
+    # initialize/list_tools/call_tool, only this job timeout saves the run.
+    timeout-minutes: 5
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,61 @@
+# End-to-end checks that the unit suite structurally cannot cover (#79):
+# the HTTP agent drives a real server over the network, and the MCP smoke
+# test drives the real stdio server — the transport whose startup crash
+# once went unnoticed because unit tests call _execute_tool directly (#56).
+name: E2E
+
+on:
+  pull_request:
+    branches: [main]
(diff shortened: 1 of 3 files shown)


cdelmonte-zg:
Done: `timeout-minutes: 5` on mcp-e2e and `10` on http-e2e, with a comment explaining why (stdio has no per-request timeouts: a server that starts but hangs in `initialize` would only be saved by the job timeout).

On `EXPECTED_TOOLS = 24`: yes, the intent is exactly to force a deliberate update — the docs advertise the tool count, so growing or shrinking the MCP surface must consciously touch both (metadata never lies). The script now carries a comment saying so.

Issue #81, opened 2026-07-07
Native-app redirect URIs: custom schemes and loopback variability (RFC 8252)


cdelmonte-zg:
From the #78 review (non-blocking). Two assumptions in `/authorize` are web-client-shaped:

1. The syntactic validation requires `scheme` **and** `netloc`, so a private-use scheme URI like `com.example.app:/oauth2redirect` (RFC 8252 §7.1) is rejected before registration matching even runs.
2. Exact string matching is correct for web and custom-scheme URIs, but loopback redirects are the deliberate exception: RFC 8252 §7.3 requires the authorization server to allow a **variable port** on `http://127.0.0.1:{any}/path` (and `[::1]`), because native apps bind an ephemeral port.

Neither matters for the current web-client profile, and stays out of #68's first cut. If native clients become a testing target:

- relax the syntactic check to "absolute URI" (scheme required, netloc optional when a path is present);
- keep exact matching everywhere except registered loopback URIs, where the port component is ignored per §7.3.

Both changes are RFC-citable and additive; until then, the web-only assumption is a documented limitation rather than an accident.

Issue #46, opened 2026-06-11
Optional refresh token rotation, to let clients test rotation handling


cdelmonte-zg:
## Problem

The `refresh_token` grant returns the same refresh token indefinitely. Many real IdPs rotate refresh tokens on every use (recommended by the OAuth 2.0 Security BCP, mandatory for public clients in OAuth 2.1), and client libraries have dedicated logic for storing the rotated token. nanoidp currently can't exercise that path.

The in-memory `_revoked_tokens` JTI set (`src/nanoidp/routes/oauth.py:672-673`) already provides the building block for invalidating the old token.

## Suggested fix

Add a setting (e.g. `oauth.refresh_token_rotation: true|false`, default false to preserve current behavior) that, when enabled:

- issues a new refresh token on every `refresh_token` grant,
- adds the consumed token's `jti` to the revoked set (so reuse fails — lets developers test reuse-detection handling too).

Fits naturally after #39 (refresh must first carry scope/ID Token correctly).

## Acceptance criteria

- [ ] Rotation off by default; behavior unchanged
- [ ] With rotation on: new refresh token returned, old one rejected on reuse
- [ ] Documented in README settings reference; MCP `update_settings` supports it

PR #82, opened 2026-07-07
feat: oauth21 security profile — draft OAuth 2.1 protocol strictness

### LINKED ISSUE #68: oauth21 security profile: PKCE required, S256-only, rotation on, no password grant

cdelmonte-zg:
Closes #68 — completes the core of the [OAuth 2.1 profile](https://github.com/cdelmonte-zg/nanoidp/milestone/1) milestone (#81, native-app URIs, remains as the follow-up).

## The profile

`--profile oauth21`, alongside `dev` and `stricter-dev`:

| Behavior | Citation |
|---|---|
| PKCE required on the authorization code flow | draft-ietf-oauth-v2-1 §4.1.1 |
| S256 only (`plain` rejected, like stricter-dev) | §7.5.2 |
| Refresh token rotation forced on | §4.3.1 (family revocation from #46/#56 already in place) |
| `password` grant removed — rejected with 400 | RFC 6749 §5.2 |
| Registered `redirect_uris` mandatory at `/authorize` | §2.3, on top of #67's exact matching |

**Deliberately orthogonal to `stricter-dev`** (the open design question in the issue): `oauth21` hardens the *protocol*, `stricter-dev` the *runtime* (bcrypt, CORS, rate limiting). The docs table now says exactly that.

## Design: derived properties, one source of truth

The routes and the shared discovery builder consume four new `Settings` properties (`pkce_required`, `pkce_plain_allowed`, `rotation_enabled`, `password_grant_enabled`) instead of raw fields. Consequences:

- a profile means the same thing whether it arrives via `--profile` or `settings.yaml`;
- **metadata never lies**: under `oauth21`, `password` disappears from `grant_types_supported` and `code_challenge_methods_supported` is `["S256"]` — the same builder serves HTTP and MCP, so the surfaces cannot drift;
- `stricter-dev` behavior is **byte-identical** to before (its PKCE requirement still comes from the CLI setting `require_pkce=True`, as today — asserted by a dedicated test).

## Ships whole

- **MCP**: `get_settings` now reports `security_profile`; discovery parity is automatic via the shared builder (asserted by test).
- **e2e**: `test_agent.py --oauth21` runs a dedicated suite (discovery honesty, password grant rejected, no-PKCE/plain/unregistered-client rejected, S256+registered accepted); the E2E workflow boots a **second server** with `--profile oauth21` on port 8001 and runs it.
- **Docs**: security guide's profile tables (three profiles, feature comparison incl. protocol rows) + README one-liner.
- **Tests**: 17 new (697 total) — validator, derived properties per profile, discovery honesty both ways, token endpoint (password rejected / client_credentials unaffected), authorize strictness, MCP parity.

## Verified

`mypy` clean, `ruff` clean, 697 unit tests green, coverage 73% ≥ gate, `mdbook build` green — and the oauth21 e2e suite run locally against a real `--profile oauth21` server: **3/3**.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -42,9 +42,26 @@ jobs:
       - name: Run e2e agent
         run: python examples/test_agent.py
 
+      - name: Start oauth21 server
+        run: |
+          PORT=8001 python -m nanoidp --profile oauth21 > server-oauth21.log 2>&1 &
+          for i in $(seq 1 30); do
+            curl -sf http://localhost:8001/api/health && exit 0
+            sleep 1
+          done
+          echo "oauth21 server did not become healthy" >&2
+          cat server-oauth21.log >&2
+          exit 1
+
+      - name: Run oauth21 profile suite
+        run: python examples/test_agent.py --oauth21 --url http://localhost:8001
+
       - name: Server log on failure
         if: failure()
-        run: cat server.log
+        run: |
+          cat server.log
+          echo "===== oauth21 server ====="
+          cat server-oauth21.log 2>/dev/null || true
 
   mcp-e2e:
     runs-on: ubuntu-latest
--- a/README.md
+++ b/README.md
@@ -82,7 +82,8 @@ The full documentation lives at
 NanoIDP is a **development/testing tool** and must NOT be used in
 production. Defaults favor convenience (plaintext passwords in config,
 permissive CORS, open redirects); hardening is opt-in via the
-`stricter-dev` profile and explicit settings. The
+`stricter-dev` (runtime) and `oauth21` (draft OAuth 2.1 protocol
+strictness) profiles and explicit settings. The
 [Security guide](https://cdelmonte-zg.github.io/nanoidp/guides/SECURITY.html)
 draws the line precisely.
 
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -12,12 +12,15 @@ By design, NanoIDP prioritizes developer convenience over security hardening. It
 
 ## Security Profiles
 
-NanoIDP supports two security profiles to balance convenience with basic security controls:
+NanoIDP supports three security profiles to balance convenience with basic security controls:
 
 | Profile | Description |
 |---------|-------------|
 | `dev` (default) | Maximum convenience for development: plaintext passwords, permissive CORS, no rate limiting |
-| `stricter-dev` | Semi-hardened mode: bcrypt passwords, restricted CORS, rate limiting, debug mode blocked |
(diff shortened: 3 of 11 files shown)


cdelmonte-zg:
Spot on: the PR description was ahead of the code. `security_profile` existed only as a model default plus the CLI mutation — a YAML-declared profile was **silently ignored** on load and dropped on save. Fixed in the second commit:

- `_load_settings()` reads the top-level `security_profile` key (default `dev`);
- `save()` persists it when non-default (no churn on existing files);
- the stricter-dev runtime mutations in `create_app` now key off the **effective** profile (CLI or YAML), so `security_profile: stricter-dev` in YAML finally means what it says — before the fix it meant nothing, so nobody can depend on the old behavior;
- the CLI `--profile` still wins over YAML.

5 new tests (`TestYamlProfileParity`): YAML load → behavior, stricter-dev from YAML with runtime hardening, CLI-wins-over-YAML, save round-trip, default omission. 702 total, all green.

On `timeout-minutes`: already there — `http-e2e: 10` and `mcp-e2e: 5`, added on main from your #80 review; they don't show in this PR's diff because it only touches the steps. The 10-minute budget comfortably covers the second server too (the oauth21 suite runs in ~5s).

Issue #83, opened 2026-07-07
Single persistence path for the YAML config


cdelmonte-zg:
`ConfigManager.save()` and `services/yaml_writer.py` both serialize the same YAML files, with the client-entry building duplicated line for line. This drift has already bitten twice:

- #32: UI saves went through the writer, which didn't know about `additional_audiences` — silently dropped from disk.
- #78: adding `redirect_uris` required touching **both** serializers; forgetting one would have reintroduced the same class of bug (and `security_profile` in #82 is only in `ConfigManager.save()` — the drift is live today).

This is principle 4 ("shared builders so equivalent surfaces cannot drift") applied to persistence, exactly like the shared discovery builder fixed #40.

## Proposal

One serialization function per document (`settings.yaml`, `users.yaml`) in a single module; `ConfigManager.save()` and `YamlWriter` both delegate to it (or `ConfigManager.save()` delegates to the writer outright). A client entry is built in exactly one place.

Strictly behavior-invariant: byte-identical output for the current fields, asserted by round-trip tests. Any field found to differ between the two paths today gets surfaced in the PR description, not silently changed.

PR #92, opened 2026-07-07
feat: opt-in verification of signed AuthnRequests, both bindings

### LINKED ISSUE #69: Opt-in verification of signed AuthnRequests (both bindings) with SP certificate registration
`routes/saml.py` explicitly skips AuthnRequest signature verification today (the NOTE around line 420: Redirect-binding `Signature`/`SigAlg` query parameters are accepted and ignored). SPs that sign their requests — a common hardened configuration — currently cannot test that nanoidp actually validates them.

## Proposal

- SP certificate registration in `settings.yaml` (`saml.sp_certificates`, PEM; start global, per-SP keyed by issuer if needed).
- New opt-in setting `saml.want_authn_requests_signed: true` which:
  - **HTTP-Redirect binding**: verifies the deflate-encoded query-string signature over `SAMLRequest`/`RelayState`/`SigAlg` (SAML 2.0 Bindings §3.4.4.1), rejecting unsigned or invalid requests with 400.
  - **HTTP-POST binding**: verifies the embedded `<ds:Signature>` (SAML 2.0 Core §5), same rejection semantics.
- **Metadata never lies** (principle 2): `WantAuthnRequestsSigned="true"` is advertised in `/saml/metadata` if and only if the setting is on.
- Default stays off (principle 3) — unsigned requests keep working out of the box.

Same XXE-safe lxml configuration as the existing response-signing path; signature verification via signxml, reusing `services/crypto.py` where possible.

Per principle 5: e2e coverage in `examples/test_agent.py` (signed-request happy path + tampered-signature rejection under both bindings) and docs in the same PR. MCP exposure: settings toggle only (protocol surface stays HTTP, per the scoped parity principle).

cdelmonte-zg:
Closes #69 — completes the [SAML hardening](https://github.com/cdelmonte-zg/nanoidp/milestone/2) milestone, and with it the last big feature on the vision board.

## Behavior

```yaml
saml:
  want_authn_requests_signed: true
  sp_certificates:
    - /path/to/sp-cert.pem
```

With the flag on, nanoidp **requires and verifies** AuthnRequest signatures and rejects unsigned/invalid requests with `400` (audit-logged):

| Binding | What is verified | Citation |
|---|---|---|
| HTTP-Redirect | Query-string signature over the URL-encoded `SAMLRequest[&RelayState]&SigAlg` fragment **exactly as transmitted** — verification runs on the raw query string, since re-encoding would break it. `rsa-sha256`/`rsa-sha512`/legacy `rsa-sha1`. | Bindings §3.4.4.1 |
| HTTP-POST | Enveloped `<ds:Signature>` via signxml | Core §5 |

- **Metadata never lies**: `WantAuthnRequestsSigned="true"` is advertised if and only if enforcement is on.
- **Default off** (principle 3): unsigned requests and ignored `Signature`/`SigAlg` params behave exactly as before — all 710 pre-existing tests pass untouched.
- **Fails closed**: flag on with no registered certificates rejects rather than waves through.
- **Verification point**: where the request *enters*. The POST-binding inline-login leg re-verifies (the signature travels inside the XML); the Redirect leg can't — the query signature doesn't survive the login form roundtrip — so it relies on the entry check, documented in the code.

## Ships whole (principle 5)

- **Settings**: model fields, YAML load, persistence through the shared serializer (both writers, courtesy of #83).
- **Web UI**: toggle + SP-certificates field on the settings page.
- **MCP**: `get_settings`/`update_settings` cover both fields (settings surface only — the protocol surface stays HTTP, per the scoped parity principle).
- **Tooling**: `examples/gen_sp_keypair.py` generates a test SP keypair (documented in the book — useful to anyone testing a signing SP).
- **e2e**: `test_agent.py --saml-signed` signs real requests with the SP key and asserts metadata + accept/reject under both bindings; the E2E workflow runs it against a **third server** booted with verification enabled and a generated keypair.
- **Docs**: new "Signed AuthnRequests" section in the book's SAML reference.
- **Tests**: 16 new (726 total) — signed/unsigned/tampered/wrong-key/unsupported-SigAlg/no-certs per binding, full login flow with re-verification, metadata both ways, defaults-off, settings round-trip.

## Verified

`mypy` clean (25 files), `ruff` clean, **726 tests green, coverage 76%** (+1) — plus, against live servers: the standard agent **41/41** (defaults untouched) and the signed suite **3/3** with verification enforced.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -56,12 +56,45 @@ jobs:
       - name: Run oauth21 profile suite
         run: python examples/test_agent.py --oauth21 --url http://localhost:8001
 
+      - name: Prepare signed-SAML config
+        run: |
+          python examples/gen_sp_keypair.py --out e2e-saml
+          cp -r config e2e-saml/config
+          python - <<'EOF'
+          import yaml
+          path = "e2e-saml/config/settings.yaml"
+          with open(path) as f:
+              doc = yaml.safe_load(f)
+          doc.setdefault("saml", {})["want_authn_requests_signed"] = True
+          doc["saml"]["sp_certificates"] = ["e2e-saml/sp-cert.pem"]
+          with open(path, "w") as f:
+              yaml.safe_dump(doc, f, sort_keys=False)
+          EOF
+
+      - name: Start signed-SAML server
+        run: |
+          PORT=8002 NANOIDP_CONFIG_DIR=e2e-saml/config python -m nanoidp > server-saml.log 2>&1 &
+          for i in $(seq 1 30); do
+            curl -sf http://localhost:8002/api/health && exit 0
+            sleep 1
+          done
+          echo "signed-SAML server did not become healthy" >&2
+          cat server-saml.log >&2
+          exit 1
+
+      - name: Run signed-AuthnRequests suite
+        run: |
+          python examples/test_agent.py --saml-signed --url http://localhost:8002 \
+            --sp-key e2e-saml/sp-key.pem --sp-cert e2e-saml/sp-cert.pem
+
       - name: Server log on failure
         if: failure()
         run: |
           cat server.log
           echo "===== oauth21 server ====="
           cat server-oauth21.log 2>/dev/null || true
+          echo "===== signed-SAML server ====="
+          cat server-saml.log 2>/dev/null || true
 
   mcp-e2e:
     runs-on: ubuntu-latest
--- a/book/src/reference/saml.md
+++ b/book/src/reference/saml.md
@@ -58,6 +58,38 @@ When `sign_responses: true` (default), responses include:
 When `sign_responses: false`, responses are sent without any signature
 elements.
 
+## Signed AuthnRequests
+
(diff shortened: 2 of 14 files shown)


cdelmonte-zg:
Bypass closed — thanks, it was real and the diagnosis was exact: the `saml_original_verb=GET` leg trusted hidden form fields alone, so a forged POST skipped verification entirely (and with an authenticated session went straight to response generation).

Fixed as you proposed, exact-match variant: a successful Redirect verification stores `{SAMLRequest, RelayState}` in the session, and the login leg is admitted **only** for byte-identical values — otherwise 400, fail closed (including the case where no verification ever happened). Each newly verified GET overwrites the binding; replaying the same identical request within the session is admitted by construction (it is indistinguishable from re-verifying the same valid signature).

Four regression tests in `TestRedirectLegBinding`: forged leg with no prior verification → 400; forged leg with an authenticated session → 400 and no `SAMLResponse`; swapped request after a legitimate GET → 400; the legitimate full flow → 200 with `SAMLResponse`. Suite at 730; the signed e2e suite re-run live: 3/3.

Issue #103, opened 2026-07-08
Docs: explain where profile/email claims come from (UserInfo vs tokens)


cdelmonte-zg:
Follow-up from #101.

## Problem

The discovery document advertises `scopes_supported: [openid, profile, email, offline_access]` and a `claims_supported` list including `email`, `email_verified`, `preferred_username`, `name`, ... (`src/nanoidp/services/discovery.py`), but nothing in the docs explains **where those claims are actually served from**.

This trips up users (see #101): they request `scope=openid email`, expect an `email` claim inside the ID Token (or access token), and don't find it - because nanoidp returns those claims from the **UserInfo endpoint**, not embedded in the tokens. That is standard OIDC behaviour for the authorization code flow ([OIDC Core §5.4](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims)), but it is currently undocumented.

## Suggested changes

- In the token / endpoints reference (`book/src/reference/tokens.md`, `book/src/reference/endpoints.md`), state explicitly:
  - The ID Token and access token carry `sub`, `iss`, `aud`, `exp`, `iat`, `auth_time`, `roles`, `tenant`, `authorities`, etc.
  - `profile`/`email` scope claims (`email`, `email_verified`, `preferred_username`, ...) are returned from `GET /userinfo` using the access token, not from the tokens themselves.
- Add a short `/userinfo` request/response example.
- Cross-link from the discovery/`claims_supported` description so the advertised claims point to where they are served.

## Related

- #102 (scope-based claim filtering on `/userinfo`) - the docs should reflect the final behaviour once that lands.
- Possible future feature: optionally embedding standard claims in the ID Token (the feature gap noted in #101). Not covered by this docs issue.

cdelmonte-zg:
cc @axc450 (reported in #101) - feel free to follow along or chime in here.

Issue #102, opened 2026-07-08
/userinfo returns profile/email claims regardless of granted scopes


cdelmonte-zg:
Follow-up from #101.

## Problem

The `/userinfo` endpoint returns `email`, `email_verified`, `preferred_username`, `roles`, `tenant`, etc. for any valid access token, **without checking which scopes were actually granted**. Per [OIDC Core §5.4](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims), the claims associated with the `profile`, `email`, `address`, and `phone` scopes should only be returned when the corresponding scope was granted.

Today `scope` never reaches the claim-selection logic at UserInfo: the response is built unconditionally in `src/nanoidp/routes/oauth.py` (~lines 974-988).

## Suggested direction

- Thread the granted scope through to `/userinfo` (it is available in the access token / grant context).
- Gate the optional claims by scope:
  - `email` scope → `email`, `email_verified`
  - `profile` scope → `preferred_username`, `name`, ... (whatever we choose to support)
  - `sub` is always returned.
- Keep it aligned with what `claims_supported` / `scopes_supported` advertise in the discovery document (`src/nanoidp/services/discovery.py`).

## Notes

- This is a spec-compliance / privacy hardening item, not a security hole (the token is still authenticated).
- Should ship with tests covering "scope granted → claim present" and "scope not granted → claim absent".

cdelmonte-zg:
cc @axc450 (reported in #101) - feel free to follow along or chime in here.

Issue #101, opened 2026-07-08
How to add claims to ID token?


axc450:
Sorry if I missed it in the docs, but from `.well-known/openid-configuration` I can see a list of supported claims, but how would I have these returned in ID tokens?

For example, I want to see the claim `email` in the token. I thought maybe I could request this by `scope=openid email` but this does not seem to include the user's email in any of the tokens returned.

cdelmonte-zg:
You are right, this isn't covered clearly in the docs, so the confusion is fair.

The short answer is that at the moment the email claim isn't embedded in any of the tokens. Following the standard OIDC behaviour for the authorization code flow, you retrieve it from the UserInfo endpoint using the access token:

```
GET /userinfo
Authorization: Bearer <access_token>
```

which returns:

```json
{
  "sub": "admin",
  "email": "admin@example.org",
  "email_verified": true,
  "preferred_username": "admin",
  "roles": ["..."],
  "tenant": "..."
}
```

If your client specifically needs email inside the ID Token, that's a feature gap. The standards-aligned way to ask for that is the OIDC `claims` request parameter (OIDC Core §5.5), which we don't support yet: I'm opening for this a new issue #104.

Another point is that /userinfo currently returns email regardless of the granted scopes, I'm opening for this a new issue #102.

I am also going to update the docs to state explicitly where profile/email claims come from, tracked in #103.

Hope that unblocks you in the meantime: call /userinfo with the access token and you'll get the email.

axc450:
Brilliant, thank you!

cdelmonte-zg:
Glad that helped! Following up now that it all shipped in **v2.4.0**.

Your original ask, `email` inside the ID Token, is now supported. There are two ways to get `email`, depending on where you need it:

**1. From `/userinfo`** (the standard place for the authorization code flow). Request `scope=openid email`, then call UserInfo with the access token:

```
GET /userinfo
Authorization: Bearer <access_token>
```

A note on profiles: under the default `dev` profile UserInfo returns these claims regardless of scope. Under the `stricter-dev` and `oauth21` profiles they are gated by the granted scope (`email` needs the `email` scope, `preferred_username` needs `profile`), so requesting `scope=openid email` is the portable way to get it in every profile.

**2. Inside the ID Token**, which is what you were after. Use the OIDC `claims` request parameter (OIDC Core §5.5) at `/authorize`:

```
GET /authorize?response_type=code&client_id=...&redirect_uri=...&scope=openid
[...]

PR #106, opened 2026-07-08
Support the OIDC claims request parameter (#104)

### LINKED ISSUE #104: Support the OIDC `claims` request parameter to deliver claims in the ID Token (OIDC Core §5.5)
Follow-up from #101.

## Context

In #101 a user expected the `email` claim inside the ID Token when requesting `scope=openid email`. Per [OIDC Core §5.4](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims), that is **not** the standard behaviour for the authorization code flow: scope-based `profile`/`email`/`address`/`phone` claims are served from the UserInfo endpoint (which nanoidp already does). They are only placed in the ID Token when `response_type=id_token` (no access token issued).

The **standards-aligned** way for a client to request that a specific claim be delivered in the ID Token is the [`claims` request parameter (OIDC Core §5.5)](https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter), which nanoidp does not currently support.

## Proposal

Support the OIDC `claims` request parameter at `/authorize`, e.g.:

```
claims={"id_token":{"email":null,"email_verified":null},"userinfo":{"name":null}}
```

Scope of work:

- Parse the `claims` parameter (JSON) on the authorization request; validate structure, ignore/deny unknown claims gracefully.
- Honour the `id_token` member: include the requested claims in the ID Token when the user/config can supply them.
- Honour the `userinfo` member for `/userinfo` (composes with the scope-based gating from #102).
- Support at least the `null` (voluntary) form; `{"essential":true}` and `{"value":...}`/`{"values":[...]}` can be a later refinement.
- Advertise support via `claims_parameter_supported: true` in the discovery document (`src/nanoidp/services/discovery.py`).
- Tests: requested claim present in the ID Token; unrequested claim absent; malformed `claims` handled cleanly.

## Notes

- Optional OIDC feature; many minimal IdPs skip it, so keeping `priority: low`.
- A lighter-weight, non-standard alternative (a config flag that always embeds standard claims in the ID Token, à la Auth0/Keycloak) is spec-compatible but not spec-required; noting it here as an explicitly rejected default in favour of the standard mechanism.
- Per the project convention, any change here must also update the MCP server and `examples/test_agent.py` where relevant.

cdelmonte-zg:
Closes #104. Follow-up from #101 / #102.

## What

`/authorize` now accepts the OIDC [`claims` request parameter](https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter) (OIDC Core §5.5) - the standards-aligned way to ask for specific claims in the ID Token, which is what the reporter in #101 originally expected from `scope=openid email`.

```
claims={"id_token":{"email":null,"email_verified":null},"userinfo":{"name":null}}
```

- **`id_token` member** → the requested claims are resolved from the user and embedded in the ID Token.
- **`userinfo` member** → carried on the access token (`req_userinfo_claims`) and returned by `/userinfo`, composing with the #102 scope gating (so a client can get `email` from UserInfo even under a stricter profile that would otherwise gate it out).

## Behaviour

- Voluntary (`null`) form honoured; `essential`/`value` refinements are parsed but not yet acted on (§5.5.1).
- Claims are resolved via a shared `resolve_user_claim()` helper (email, email_verified, preferred_username, roles, tenant, identity_class, entitlements, and any user attribute). **Unresolvable names are skipped**, and **protocol claims (`auth_time`/`at_hash`/`azp`/…) are never overwritten** (`setdefault`).
- **Malformed `claims`** is ignored with a warning rather than failing the authorization flow.
- Discovery advertises `claims_parameter_supported: true`.
- MCP `generate_token` gains an `id_token_claims` argument (convention: MCP + E2E kept in sync).

## Scope / follow-ups

- Scoped to the **authorization code** grant (where the `claims` parameter lives per spec).
- Requested claims are **not yet persisted across a refresh** - the refreshed ID Token won't carry them. Noted as a deliberate follow-up to keep this PR focused.

## Tests

- New `tests/test_claims_request_parameter.py` (21 tests): parameter parsing (voluntary/essential/malformed/unknown members), `resolve_user_claim`, id_token member via the full auth-code flow (present/absent/unresolvable/protocol-claim-safe/malformed), userinfo member composing with #102 gating, discovery advertisement.
- Extended the E2E `test_authorization_code_pkce` to request `email` via `claims` and assert it lands in the ID Token.
- Full suite green (778 unit), ruff + mypy clean, E2E dev 41/41.

## Docs

- #103 will document the `claims` parameter alongside where `profile`/`email` claims come from.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   granted scope (RFC 9068 §2.2.3), letting resource endpoints reason about it.
   Set authoritatively in `TokenService.create_token`, so a caller-supplied
   `extra_claims` cannot override it.
+- **OIDC `claims` request parameter** (#104, OIDC Core §5.5): `/authorize`
+  accepts a `claims` parameter to request specific claims in the ID Token
+  (`id_token` member) or from UserInfo (`userinfo` member), e.g.
+  `claims={"id_token":{"email":null}}`. Requested claims are resolved from the
+  user and added when available (voluntary form, §5.5.1); protocol claims are
+  never overwritten and unresolvable names are skipped. Malformed input is
+  ignored with a warning rather than failing the flow. Discovery advertises
+  `claims_parameter_supported: true`, and the MCP `generate_token` tool gains an
+  `id_token_claims` argument. Scoped to the authorization code grant; the
+  requested claims are not yet persisted across a refresh.
 
 ### Changed
 - **`/userinfo` gates `email`/`profile` claims by granted scope** (#102, OIDC
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `claims_parameter_supported: true`, and the MCP `generate_token` tool gains an
   `id_token_claims` argument. Scoped to the authorization code grant; the
   requested claims are not yet persisted across a refresh.
+  `TokenService.create_token` now strips `scope`/`req_userinfo_claims` from a
+  caller-supplied `extra` before setting them authoritatively, so the `/token`
+  `extra` parameter can never smuggle scope-gated claims past `/userinfo`
+  (closes a spoofing gap in the #102 scope handling too).
 
 ### Changed
 - **`/userinfo` gates `email`/`profile` claims by granted scope** (#102, OIDC
--- a/examples/test_agent.py
+++ b/examples/test_agent.py
@@ -369,7 +369,10 @@ class NanoIDPTestAgent:
                 "scope": "openid profile",
                 "state": state,
                 "code_challenge": challenge,
-                "code_challenge_method": "S256"
+                "code_challenge_method": "S256",
+                # OIDC `claims` request parameter (§5.5, #104): ask for the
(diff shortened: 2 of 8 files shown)


cdelmonte-zg:
Good catch - fixed in `223b0ca`.

You're right: `create_token` set `scope`/`req_userinfo_claims` only when the authoritative value was present, but merged the caller-supplied `extra` first, so `extra={"req_userinfo_claims":["email"]}` survived and `/userinfo` honoured it past the #102 gating. The `/token` `extra` parameter is arbitrary caller JSON, so this was reachable.

The fix pops both reserved claims after the `extra` merge and before setting them authoritatively:

```python
if extra_claims:
    extra.update(extra_claims)

for reserved in ("scope", "req_userinfo_claims"):
    extra.pop(reserved, None)

if scope:
    extra["scope"] = scope
if userinfo_claims:
    extra["req_userinfo_claims"] = userinfo_claims
```

While there I found the **same latent gap for `scope`**, pre-existing from #102: `if scope: extra["scope"] = scope` only protects when `scope` is truthy, so `extra={"scope":"openid email"}` on a grant with no scope (e.g. client_credentials) would smuggle a scope claim and defeat the gating. Popping `scope` too closes that.

Added two hardening tests (stricter-dev): both spoof attempts leave the claim off the token and `/userinfo` does not return `email`. Full suite green (780), ruff + mypy clean.

PR #107, opened 2026-07-08
Docs: where profile/email claims come from + claims parameter (#103)

### LINKED ISSUE #103: Docs: explain where profile/email claims come from (UserInfo vs tokens)

cdelmonte-zg:
Closes #103. Follow-up from #101, documenting the behaviour shipped in #102 and #104.

## What

Explains the point of confusion from #101: the standard OIDC `profile` / `email` claims (`email`, `email_verified`, `preferred_username`, ...) are **served from `GET /userinfo`** using the access token, **not embedded in the ID Token or access token** (OIDC Core §5.4).

- **`book/src/reference/tokens.md`**: new "Where do the `email` / `profile` claims come from?" section with a `/userinfo` request/response example; documents the #102 scope gating (dev permissive; `stricter-dev`/`oauth21` gate `email` behind the `email` scope and `preferred_username` behind `profile`; custom claims always returned); new "Requesting claims in the ID Token (`claims` parameter)" section for the #104 feature. The ID Token section now points here.
- **`book/src/reference/endpoints.md`**: note on `/userinfo` cross-linking to the claims explanation.
- **`book/src/guides/token-requests.md`**: a UserInfo section with a curl example.

## Notes

- Docs only - no code changes.
- `mdbook build` is clean; all internal anchor links resolve (verified against the generated HTML).

### DIFF
--- a/book/src/guides/token-requests.md
+++ b/book/src/guides/token-requests.md
@@ -65,6 +65,20 @@ curl -X POST 'http://localhost:8000/token' \
   -d 'grant_type=urn:ietf:params:oauth:grant-type:device_code&device_code=YOUR_DEVICE_CODE'
 ```
 
+## UserInfo
+
+The `email` / `profile` claims (`email`, `email_verified`,
+`preferred_username`, ...) are returned here, using the access token - not
+embedded in the ID Token. See
+[Tokens and claims](../reference/tokens.md#where-do-the-email--profile-claims-come-from)
+for the details and for the `claims` request parameter that can put a
+claim inside the ID Token.
+
+```bash
+curl 'http://localhost:8000/userinfo' \
+  -H 'Authorization: Bearer YOUR_ACCESS_TOKEN'
+```
+
 ## Token introspection
 
 ```bash
--- a/book/src/reference/endpoints.md
+++ b/book/src/reference/endpoints.md
@@ -18,6 +18,10 @@
 curl examples for every grant are in
 [Requesting tokens](../guides/token-requests.md).
 
+The standard OIDC `profile` / `email` claims (`email`, `email_verified`,
+`preferred_username`, ...) are served from `GET /userinfo`, not embedded
+in the tokens - see [Tokens and claims](tokens.md#where-do-the-email--profile-claims-come-from).
+
 ## SAML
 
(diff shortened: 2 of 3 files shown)


cdelmonte-zg:
Good point - fixed in `ea263d4`.

You're right that `name` isn't a claim NanoIDP resolves (it's not in `resolve_user_claim` nor in `claims_supported`); listing it set a false expectation. Changed:

> (`email`, `email_verified`, `preferred_username`, `name`, ...)

to:

> (`email`, `email_verified`, `preferred_username`; plus NanoIDP-specific and custom claims where configured)

`mdbook build` still clean, anchors unchanged.

Issue #110, opened 2026-07-08
claims parameter can overwrite registered ID Token claims (iss/sub/aud/exp/...) via a colliding user attribute


cdelmonte-zg:
Found in a post-release review of 2.4.0 (#104).

## Problem

`TokenService.create_token` resolves the OIDC `claims` `id_token` member with `resolve_user_claim` and adds each value with `id_extra.setdefault(name, value)`. The in-code comment states protocol claims are never overwritten, but that guard is incomplete:

- `setdefault` only protects keys already present in `id_extra` (`token_use`, `auth_time`, `at_hash`, `azp`).
- The **registered JWT claims** (`iss`, `sub`, `aud`, `exp`, `iat`, `nbf`, `jti`) are set later by `CryptoService.create_jwt`, which does `payload.update(extra)` **after** setting them (`crypto.py:328`). So any key in `extra`/`id_extra` overrides them.

`resolve_user_claim` has a catch-all `if name in user.attributes` branch. So if a user is configured with an attribute whose name collides with a registered claim, a client can surface/overwrite it:

```
# user attributes: {"aud": "https://attacker.example", "exp": 9999999999}
GET /authorize?...&scope=openid&claims={"id_token":{"aud":null,"exp":null}}
```

→ the issued ID Token's `aud` / `exp` are taken from the user attribute, defeating `_resolve_id_token_audience` and/or the real expiry.

The value comes from admin-configured attributes (not attacker-chosen), so exploitability is limited, but it is a **new surface opened by #104** and the documented guarantee does not hold.

## Suggested fix

Refuse reserved claim names at the resolver boundary (single choke point), e.g. `resolve_user_claim` returns `(False, None)` for names in a `_RESERVED_CLAIMS` set: the registered claims `{iss, sub, aud, exp, iat, nbf, jti}` plus the protocol claims `{token_use, auth_time, at_hash, azp, nonce, scope, req_userinfo_claims}`. None of these are legitimately requestable via the `claims` parameter, so skipping them is safe and protects both the ID Token and `/userinfo` paths. Fix the inaccurate comment too.

## Related hardening (same root)

The anti-spoofing `extra.pop("scope"/"req_userinfo_claims")` in `create_token` is a hand-maintained denylist that must be kept in sync with every future authoritative claim. Worth considering a single "authoritative claims are written last, unconditionally" mechanism so this class of overwrite is structurally impossible rather than name-by-name.

PR #114, opened 2026-07-18
Persist OIDC claims requests across refresh and close the #104 coverage gaps

### LINKED ISSUE #112: Persist OIDC claims-parameter requests across token refresh
Follow-up from the post-release review of #104 (flagged independently by all three review passes).

## Problem

The OIDC `claims` request is captured on the authorization code and threaded into `create_token` as `id_token_claims` / `userinfo_claims`, but it is **not persisted in the refresh token**. On refresh, `_grant_refresh_token` returns a `_GrantOutcome` with both defaulting to `None`, so:

- the refreshed **ID Token loses** any claims the client requested via `claims={"id_token":{...}}` (e.g. `email` present on the first token, gone after refresh);
- the refreshed **access token loses `req_userinfo_claims`**, so under `stricter-dev`/`oauth21` `/userinfo` stops returning the beyond-scope claims the client had requested.

Same grant, different answer before vs after refresh. Unlike `nonce` (whose drop is documented and spec-justified at `oauth.py`), nothing documents or justifies this.

## Options

- Persist the normalized `claims` (or the two name lists) in the refresh token claims alongside `scope`/`auth_time`/`client_id`, and recover them in `_grant_refresh_token`, mirroring how scope/auth_time are already carried (OIDC Core §12.2).
- Or explicitly document that requested claims bind to the original authorization request and are intentionally not re-applied on refresh.

Priority low: the primary path (initial authorization) works; this only affects the token lifetime after a refresh.

### LINKED ISSUE #113: Test coverage + cleanup for the OIDC claims parameter (#104)
Follow-up from the post-release review of #104.

## Test coverage gaps

- **MCP `generate_token(id_token_claims=...)` is untested.** No test asserts the argument actually lands the claim in the ID Token, that unsupported names are skipped, or that it is inert without an `openid` scope. Per the repo convention (MCP + E2E kept in sync), this half of #104 is unverified.
- No end-to-end test drives a **custom user attribute** through `/authorize` into the ID Token.
- No assertion that the `essential`/`value` claim refinements are **accepted but ignored** end-to-end (only the parser is unit-tested).
- No test for the `/userinfo` **no-overwrite branch** (`if claim_name not in response`) under the permissive `dev` profile, nor for a refresh carrying claims (see the refresh-persistence issue).

## Related cleanup

- **`/userinfo` duplicates the claim mapping** now centralized in `resolve_user_claim` (`email_verified=True`, `preferred_username=user.username`, ...). The scope-gated default assembly could reuse the resolver so the two cannot diverge.
- (The anti-spoof `extra.pop(...)` denylist is already noted as related hardening in #110.)

## Possible extra

- Consider exposing `userinfo_claims` on the MCP `generate_token` tool for parity with the HTTP `claims` flow.

Priority low: quality/coverage, no user-facing defect.

cdelmonte-zg:
Closes #112, closes #113 — the two remaining follow-ups from the post-release review of #104.

## #112 — requested claims persist across token refresh

The claim names requested via the OIDC `claims` parameter are now persisted in the refresh token as `req_id_token_claims` / `req_userinfo_claims`, alongside the already-persisted `scope`, `auth_time`, `client_id` and `rt_family`, and recovered by `_grant_refresh_token`:

- a refreshed **ID Token keeps** the claims the client requested via `claims={"id_token":{...}}` (same answer before and after refresh, OIDC Core §12.2);
- a refreshed **access token keeps `req_userinfo_claims`**, so `/userinfo` keeps honouring the `userinfo` member under `stricter-dev`/`oauth21`;
- persistence survives **rotation** (the rotated refresh token re-carries the names);
- **legacy refresh tokens** (minted before this change) carry neither claim and refresh exactly as before;
- both names are **reserved**: `resolve_user_claim` refuses them (extending the #110 choke point) and `create_token` strips them from caller-supplied `extra`, so they cannot be spoofed onto an access token.

`nonce` remains the deliberate, documented exception (binds the original authentication request only).

## #113 — coverage + cleanup

- **Cleanup:** `/userinfo`'s default claim assembly (scope-gated standard claims + nanoidp-specific claims) now goes through `resolve_user_claim`, the same resolver backing the `claims` parameter, so the two mappings cannot diverge. Behavior-invariant (locked by an invariance test).
- **MCP parity:** `generate_token` gains `userinfo_claims`, mirroring the HTTP `userinfo` member.
- **New tests** (22 added across the PR, suite at 799):
  - MCP `generate_token(id_token_claims)`: claims land in the ID Token, unresolvable names are skipped, inert without an `openid` scope;
  - MCP `userinfo_claims`: stamped on the access token and honoured by `/userinfo`;
  - custom user attribute driven through `/authorize` into the ID Token;
  - `essential`/`value` refinements accepted but ignored end-to-end (a `value` constraint never overrides the real value);
  - `/userinfo` no-overwrite branch (call-numbered sentinel resolver proves the scope-gated value is never re-resolved);
  - claims persistence across refresh: full authorize→token→refresh→refresh flow, stricter-dev `/userinfo` honouring after refresh, legacy tokens without the claims.
- **E2E agent:** new `Claims across refresh` test (refreshes the auth-code grant's token, asserts the requested `email` is still in the refreshed ID Token and an unrequested attribute is not).

## Verification

- Full suite: **799 passed**; ruff and mypy clean; coverage 79% (gate 75).
- Live run: `python -m nanoidp` + `examples/test_agent.py` → **42/42**; manual positive (claims survive refresh via the real authorize flow, `/userinfo` returns the requested attribute on the refreshed access token) and negative (legacy refresh token stays clean) checks against the live server.

Docs: `book/src/reference/tokens.md` documents the refresh persistence (and the `nonce` exception), `mcp.md` documents the new tool argument, CHANGELOG updated under Unreleased.

## Review hardening (second round)

A full self-review of this PR (8 finder angles + adversarial verification, findings posted in the comment below) led to two follow-up commits:

- `sanitize_claim_names` at the `create_token` choke point and in the `/userinfo` loop: malformed requested-claims values from hand-crafted tokens can no longer fail issuance after the refresh token was consumed, nor 500 `/userinfo`.
- MCP `generate_token` validates `id_token_claims`/`userinfo_claims` with `_normalize_str_list` (clean error on non-list input).
- `_AUTHORITATIVE_CLAIMS` shared constant ties the `extra` strip list to `_RESERVED_CLAIMS`.
- The scope-narrowing semantics are now explicit (code comment + docs) and pinned by a dedicated test: a claims request survives a narrowed refresh, including through rotation.
- Docs: refresh guide links the persistence semantics; stale "authorization code grant only" sentence amended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `userinfo` member for the refreshed access token. Refresh tokens minted
   before this change carry neither claim and refresh as before. Both names are
   reserved: they cannot be requested via the `claims` parameter nor injected
-  through the `/token` `extra` parameter.
+  through the `/token` `extra` parameter. Requested-claims values are
+  sanitized at the token service (`sanitize_claim_names`): a hand-crafted
+  refresh or access token carrying a non-list value (or non-string entries)
+  refreshes and serves `/userinfo` cleanly instead of failing token issuance
+  after the refresh token was consumed. A claims request deliberately
+  survives scope narrowing on refresh (OIDC Core §5.5 is orthogonal to
+  scope); see the token reference docs.
 - **MCP `generate_token` gains `userinfo_claims`** (#113): parity with the
   HTTP `claims` flow's `userinfo` member; the names are stamped on the access
-  token as `req_userinfo_claims` and honoured by `/userinfo`.
+  token as `req_userinfo_claims` and honoured by `/userinfo`. Both
+  `id_token_claims` and `userinfo_claims` are now validated like
+  `additional_audiences`: a non-list value is rejected with a clean error
+  instead of being minted into the token.
 
 ### Changed
 - **`/userinfo` reuses `resolve_user_claim` for its default claim assembly**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **`claims` parameter requests persist across token refresh** (#112, OIDC
+  Core §12.2): the claim names requested via the OIDC `claims` parameter are
+  now persisted in the refresh token (`req_id_token_claims` /
+  `req_userinfo_claims`, alongside `scope` and `auth_time`), so a refreshed ID
+  Token keeps the requested claims and `/userinfo` keeps honouring the
+  `userinfo` member for the refreshed access token. Refresh tokens minted
+  before this change carry neither claim and refresh as before. Both names are
+  reserved: they cannot be requested via the `claims` parameter nor injected
+  through the `/token` `extra` parameter.
+- **MCP `generate_token` gains `userinfo_claims`** (#113): parity with the
+  HTTP `claims` flow's `userinfo` member; the names are stamped on the access
+  token as `req_userinfo_claims` and honoured by `/userinfo`.
+
+### Changed
+- **`/userinfo` reuses `resolve_user_claim` for its default claim assembly**
+  (#113): the scope-gated standard claims and the nanoidp-specific claims now
+  come from the same resolver that backs the `claims` request parameter, so
+  the two mappings cannot diverge. No behavior change.
+
 ### Fixed
 - **`claims` parameter could overwrite registered ID Token claims** (#110): a
   requested claim name that collided with a user attribute (e.g. an attribute
--- a/book/src/guides/token-requests.md
+++ b/book/src/guides/token-requests.md
@@ -37,7 +37,7 @@ curl -X POST 'http://localhost:8000/token' \
 If the original grant included the `openid` scope, the refresh re-issues
 an ID Token as well. A `scope` parameter may narrow, but never broaden,
 the originally granted scope (RFC 6749 §6). Claims requested via the
-OIDC `claims` parameter persist across the refresh — including a
+OIDC `claims` parameter persist across the refresh, including a
 narrowed one; see
 [Tokens and claims](../reference/tokens.md#requesting-claims-in-the-id-token-claims-parameter).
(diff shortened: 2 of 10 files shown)


cdelmonte-zg:
**Self-review (8-angle + adversarial verification) found 10 issues; commit c2cb7e2 addresses 9.**

Correctness (all fixed):
1. **Malformed requested-claims from forged tokens** — `req_id_token_claims`/`req_userinfo_claims` recovered from a refresh token were iterated with no type check, *after* `check_and_claim_refresh` had consumed the token: a hand-crafted refresh token (a documented first-class workflow) carrying a non-list value raised TypeError → 500 with the token burned, and a retry revoked the family. Fixed with `sanitize_claim_names` at the `create_token` choke point and in the `/userinfo` loop (a bare string no longer iterates per-character either).
2. **MCP `generate_token` claims args unvalidated** — the SDK doesn't enforce `inputSchema`; a string/int `userinfo_claims` minted successfully and misbehaved later at `/userinfo`. Now validated via `_normalize_str_list` (the `additional_audiences` precedent): clean error at the tool call.
3. **Scope narrowing vs persisted claims** — a claims request deliberately survives scope narrowing on refresh (§5.5 is orthogonal to scope). This was undocumented; now stated in `_grant_refresh_token` and in the tokens reference, so a tester of downscoping under `oauth21` isn't misled.

Cleanup (fixed): `_AUTHORITATIVE_CLAIMS` shared constant ties the `extra`-pop denylist to `_RESERVED_CLAIMS` (no drift); the `/userinfo` requested-claims loop reuses `_put`; the overstated "All claim values" comment reworded; stale "authorization code grant only" sentence in tokens.md amended; the no-overwrite test rewritten order-independent; the e2e agent declares `_authcode_refresh_token` in `__init__` instead of `getattr`.

Skipped (deliberate): consolidating the per-file `_refresh` test helpers into conftest — the suite's convention is per-file helpers.

New negative coverage: forged refresh/access tokens with malformed claims refresh and serve `/userinfo` cleanly (unit + live against a running server), MCP non-list rejection. Suite 798 green, ruff/mypy clean, live e2e 42/42.

PR #117, opened 2026-08-16
feat: Add first-class group support and optional SAML roles/groups ex…

dshvedchenko:
…port

Users gain a `groups` list modelled exactly like `roles`: loaded from and persisted to users.yaml (omitted when empty), emitted as a `groups` claim on the access token and from /userinfo, requestable in the ID Token via the OIDC `claims` parameter, advertised in claims_supported, and flattened into `authorities` with the new `groups` authority prefix (default GROUP_). Editable from the user form, shown on the users list and detail pages, exposed by /api/users, and settable through the MCP create_user/update_user tools. Users without groups behave exactly as before.

Roles and groups can now also be exported as SAML attributes. They are not standard SAML attributes and every SP expects a different name, so the new saml.export_roles / saml.export_groups toggles (both off by default, so existing behaviour is preserved) are paired with saml.roles_attr_name / saml.groups_attr_name, which default to `roles` and `groups` and fall back to those defaults when blanked. Both toggles are on the settings page and the MCP update_settings tool, and apply to the SSO assertion as well as the AttributeQuery endpoint.


cdelmonte-zg:
One early request before the full review (which will come after #118 lands, since this branch cannot pass CI until the mcp 2.0 fix is in): please drop `uv.lock` from this PR.

The project currently uses pip everywhere, CI included, and has no lock file. Committing a `uv.lock` without adopting uv across CI and the dev docs would leave a file that nothing consumes and nobody keeps in sync, going stale on the first `pyproject.toml` change. If moving to uv is worth discussing, happy to take that in a dedicated issue/PR, but it shouldn't ride along inside a feature PR.

Two more heads-ups for when you get to the rebase:

- This branch (and #119) fails CI today because main's floating `mcp>=1.0.0` resolves to the 2.0 SDK, which is exactly what #118 fixes. Rebasing on top of #118 once it's merged should turn the matrix green.
- The project convention is that feature PRs touching the MCP tools also sync `examples/test_agent.py` in the same PR; this one updates `mcp_server.py` but not the example agent. Worth adding while you're rebasing.

dshvedchenko:
already rebased on MCP2.0 migration #118

cdelmonte-zg:
Before the in-depth review, one structural request. This PR bundles two independent features: first-class `groups`, and the SAML roles/groups attribute export. Per the project's VISION.md principle 5 ("features ship whole"), please split it into two PRs, each complete with its MCP exposure, `examples/test_agent.py` e2e coverage, and docs:

- **PR A - first-class groups**: User model, users.yaml, `groups` claim on access token and /userinfo, OIDC `claims` request support, `GROUP_` authority prefix, UI, `/api/users`, MCP create_user/update_user, test_agent.py, docs.
- **PR B - SAML roles/groups export** (on top of A): the `saml.export_roles`/`export_groups` toggles and attr names, SSO assertion + AttributeQuery, settings UI, MCP update_settings, test_agent.py where applicable, docs.

Both features look in line with the project's scope (helping people test group-based clients and real SPs is exactly what nanoidp is for), so this is about reviewability, not direction: two focused PRs get reviewed and merged much faster than one 3300-line one.

Practical notes: #118 is now merged, so rebase on main rather than on the fix_mcp2 branch, and note the `groups` schema entries in `_TOOLS` currently carry a leftover indentation fix from the rebase - worth folding into the split. Review of PR A will follow within a few days of it opening.

PR #118, opened 2026-08-16
fix: Migrate MCP server to the mcp 2.0 SDK

dshvedchenko:
mcp 2.0 replaced the lowlevel Server decorators (@server.list_tools() / @server.call_tool()) with on_* constructor parameters, so `mcp>=1.0.0` floating onto the new major left nanoidp.mcp_server unimportable: AttributeError: 'Server' object has no attribute 'list_tools'. That broke collection of every test module touching the MCP server.

Port to the v2 handler contract and pin mcp>=2,<3:

- Handlers take (ctx, params) and return ListToolsResult / CallToolResult. v2 removed the automatic wrapping of bare lists and dicts, so the results are constructed explicitly.
- The Server is constructed at the bottom of the module, since it now takes its handlers as constructor arguments and they must exist first.
- params.arguments is None when a call carries no arguments (the v1 decorator defaulted it to {}); it is copied into a plain dict because _check_admin_secret pops admin_secret off it and params is a validated protocol model.
- The server now reports its own version instead of the SDK's, which v2 otherwise leaves empty.

The tool set, tool schemas, readonly mode, the admin-secret gate, the stdio transport, and the nanoidp-mcp entry point are all unchanged.

Rejected and failed tool calls now set is_error: true. v2 no longer turns a handler exception into an error-flagged result, so nanoidp builds it - and readonly-mode and admin-secret rejections, which previously came back as successful results whose JSON body happened to carry an `error` key, are flagged too. Response bodies are unchanged.

Tests drive the handlers through two conftest fixtures that wrap the (ctx, params) convention, plus new coverage for the things that broke: handlers are actually registered, every tool schema declares "type": "object" (v2 validates results against the protocol schema on the wire), and a call with no arguments behaves as {}. pytest now fails on MCPDeprecationWarning so a deprecated SDK surface cannot creep in quietly.

The is_error assertion also caught a latent bug in an existing test fixture: mock_config never set default_user, so list_users returned an unserializable MagicMock and the tool call had been failing silently under an assertion that could not see it.


cdelmonte-zg (dismissed):
Thanks for the PR. CI is red and a review pass turned up four issues, two of them reproduced empirically under a real mcp 2.0.0 install. In rough priority order:

**1. CI: mcp 2.0 renamed `Tool(inputSchema=...)` to `input_schema` - mypy fails with 24 errors**

The `test` jobs fail at the mypy step before pytest even runs: all ~24 `Tool(...)` definitions in `src/nanoidp/mcp_server.py` still use the v1 `inputSchema` keyword. It works at runtime because pydantic accepts the alias, but the type-check does not. A mechanical rename fixes the CI; the wire serialization still produces `inputSchema`, so the change is zero-cost for clients.

**2. The migration silently drops the SDK's automatic argument validation**

mcp 1.x's `@server.call_tool()` had `validate_input=True` and ran `jsonschema.validate` against each tool's input schema before dispatch. The 2.0 `on_call_tool` path performs no such validation, and the PR adds no replacement, so every required-field and type constraint declared in the tool schemas is now unenforced.

Reproduced on this branch: calling `create_user` with no `username` used to be rejected with a clear schema-validation error; now it raises `KeyError` inside `_execute_tool` and returns the cryptic body `{"error": "'username'", "tool": "create_user"}`. Validate `arguments` against the tool's schema at the top of `call_tool` and return a proper `is_error` result on mismatch.

**3. The try/except no longer covers the whole handler**

v1 wrapped the entire tool handler and converted any exception into an `isError` result. The new `try` block covers only `_execute_tool`, so exceptions from `_ensure_config()`, `_check_readonly_mode`, `_check_admin_secret`, or the rejection-branch `_log_mcp_tool` calls escape the handler and surface as raw JSON-RPC errors instead of the graceful `{"error": ..., "tool": ...}` payload. `_ensure_config()` is the most likely case in practice: a config init or read failure currently contradicts the PR's own guarantee that failed calls return `CallToolResult(is_error=True)`. Widen the `try` to cover the whole body to restore the v1 invariant that no exception escapes the tool handler, and add a test where `_ensure_config` raises to pin it.

**4. Test hygiene: leaked module global and cwd-dependent test**

- `tests/conftest.py`: the autouse `reset_singletons` fixture never resets `nanoidp.mcp_server._config`, and the new `test_missing_arguments_are_treated_as_empty` is the first test to drive the real `_ensure_config()` path, caching a ConfigManager for the repo's `./config` into that global for the rest of the pytest session. Later tests then silently operate on the leaked config, and an unpatched mutating call could write to the repo's real `config/` directory. One-line fix: reset `mcp_server._config` in `reset_singletons`.
- `tests/test_mcp.py::test_missing_arguments_are_treated_as_empty` wires up the real config stack via the cwd-relative `NANOIDP_CONFIG_DIR=./config` instead of patching `_ensure_config` (the pattern used by the neighboring tests) or using `tmp_path`. Run from a foreign cwd it silently creates a `keys/` directory there with a freshly generated RSA key pair, and what it asserts varies with the invocation directory.

**Also**: `pyproject.toml` now pins `mcp>=2,<3` but `uv.lock` was not regenerated, so uv-based installs still resolve the old tree. Run `uv lock` and include the updated lock file.

This PR goes in first, then #117/#119 get rebased on top. Will re-review once the above is addressed.

dshvedchenko:
Hello, ok , lets address this PR first. question 
 there is no `uv.lock` in this repo at all — not tracked in git, not gitignored, nothing on disk. 
 so should i then add it ?

cdelmonte-zg:
You're right, my mistake: there is no `uv.lock` on main, so there is nothing to regenerate here. I misread the brand-new `uv.lock` that #117 introduces as an existing file. The "Also" point is withdrawn; the four numbered findings stand.

For context on where this is heading: the project currently uses pip everywhere (CI included) and has no lock file, so I'll be asking on #117 to drop its `uv.lock` rather than adopt one ad hoc. Nothing needed on this PR.

dshvedchenko:
removed uv.lock in all 3 prs

dshvedchenko:
PR is ready for review

cdelmonte-zg:
Second review pass on 6af1194: the migration itself checks out (constructor kwargs, snake_case fields, the gates, the no-arguments path, and the validation error path all verified live under mcp 2.0.0; full suite green). The fixes for the first round all hold. Three remaining items, the first two of the same kind the round-one findings were:

**1. Unknown-tool calls come back as successes**

`_TOOL_SCHEMAS.get(name)` returns `None` for an unknown name, so validation is skipped; `_execute_tool` falls through to `return {"error": f"Unknown tool: {name}"}` as a plain dict, and `call_tool` takes the success path: `_log_mcp_tool(success=True)` and `is_error=False` (and this error body is the only one without the `"tool"` key). A client gating on `isError` treats a misspelled tool name as success, and the audit log records a successful `mcp_tool` event for a tool that does not exist. Suggest rejecting unknown names in `call_tool` right where the schema lookup already happens: `if schema is None: return _text_result({... "code": "MCP_UNKNOWN_TOOL", "tool": name}, is_error=True)` plus the failure audit entry.

**2. Domain-level failures still return `is_error=False`, contradicting the CHANGELOG**

`save_config`'s caught write failure (`{"success": False, "error": ...}`), "user not found", and "client already exists" all return through the success branch: `success=True` in the audit log, `is_error=False` on the wire. The CHANGELOG headline says "Rejected and failed MCP tool calls now set `is_error: true`", which promises more than that. Either flag `success: False` payloads (e.g. in `call_tool`: `is_error = result.get("success") is False or "error" in result`) or narrow the CHANGELOG wording to name exactly what is flagged (the two gates, validation errors, handler exceptions).

**3. One-liner: the `call_tool` docstring names an impossible raiser**

It justifies the function-wide try/except with "the rejection branches' own `_log_mcp_tool` calls", but `_log_mcp_tool` swallows every exception internally (`try/except -> logger.warning`). Only `_ensure_config()` and `_execute_tool()` can actually raise; the docstring should say that.

Optional, take or leave (not blocking): `_TOOL_SCHEMAS` could hold precompiled `jsonschema` validators instead of raw dicts (validation currently recompiles the static schema on every call; the SDK itself caches compiled validators for output schemas); and the three copy-pasted rejection blocks (readonly / admin-secret / validation) have already drifted in their audit `details` shape, so a small `_reject(name, code, message, audit_error)` helper would pin payload and audit shape in one place.

With 1-3 addressed this is good to go from my side.

dshvedchenko:
submitted 2 review fixes

cdelmonte-zg (approved):
All three review rounds addressed and verified (line-by-line diff review plus live checks under mcp 2.0.0), CI green. The remaining non-blocking findings (isError semantics for verify_token and not-found results, audit detail shapes, precompiled validators, in-memory-client test harness) will be tracked as follow-up issues on our side. Thanks for the thorough turnaround.

PR #119, opened 2026-08-16
Issuer from request

dshvedchenko:
depends on #117 

## Summary

Adds an opt-in `oauth.issuer_from_request` setting (default `false`). When
enabled, the issuer used in the discovery document, every minted token's
`iss`, and the device flow's `verification_uri` is derived from the
incoming request's own `Host` header instead of the fixed `oauth.issuer`.

## Motivation

Docker Compose / multi-hostname dev setups often make the same NanoIDP
reachable at more than one hostname — e.g. a service name (`nanoidp:9900`)
from other containers, `localhost:9900` from the host browser. OIDC
Discovery requires the document's `issuer` to exactly equal the URL it was
fetched from, and a token's `iss` must match what discovery advertised. A
single fixed `issuer` value can only be correct for one of those hostnames
at a time, which breaks `auth/oidc`-style flows (e.g. Vault's "Sign in with
OIDC Provider" button) for whichever hostname it wasn't configured for.

## Design

- **Opt-in, default off.** Every existing deployment and test keeps its
  fixed, configured issuer unchanged.
- **MCP tools are a deliberate, documented exception.** `get_oidc_discovery`
  and `get_settings` have no HTTP request to derive a host from, so they
  always report the fixed `issuer` regardless of this setting. This is a
  narrow carve-out to the discovery-parity guarantee in
  `test_discovery_parity.py` (which only exercises the default-off case).
- Threaded through everywhere `settings.issuer` was read for a live
  response: `discovery.py`, `TokenService.create_token`, and the device
  authorization `verification_uri` in `routes/oauth.py`.

## Known limitation

The port is preserved verbatim from the `Host` header, but the **scheme**
(`http`/`https`) comes from `request.scheme` — what the WSGI server itself
saw the connection as, not the `Host` header. Behind a reverse proxy that
terminates TLS and forwards plain HTTP internally, this would report `http`
even for an external `https` request unless something like Werkzeug's
`ProxyFix` (trusting `X-Forwarded-Proto`) is added. Not an issue for a
same-network Docker Compose setup with no proxy in front; flagging it so
it isn't mistaken for full reverse-proxy support.

## Files changed

- `src/nanoidp/models.py` — new `issuer_from_request` setting
- `src/nanoidp/config.py`, `src/nanoidp/serialization.py` — load/persist
- `src/nanoidp/services/discovery.py` — optional `issuer` override param
- `src/nanoidp/services/token.py` — optional `issuer` override param on
  `create_token`, applied to access/ID/refresh token `iss`
- `src/nanoidp/routes/oauth.py` — `_effective_issuer()` helper, wired into
  discovery, token issuance, and device authorization
- `src/nanoidp/routes/ui.py`, `src/nanoidp/routes/api.py`,
  `src/nanoidp/services/yaml_writer.py`, `src/nanoidp/templates/settings.html`,
  `src/nanoidp/mcp_server.py` — Settings UI toggle, `/api/config`, MCP
  `get_settings`/`update_settings`
- `book/src/reference/configuration.md`, `CHANGELOG.md` — docs
- `tests/test_issuer_from_request.py` — new coverage (default-off
  unaffected, discovery/token/device-flow all reflect the Host header when
  on, MCP tools stay fixed)

## Test plan

- [x] `pytest` — 852 passed, no regressions
- [x] `ruff check` — clean on all touched files
- [x] Manually verified port handling: `Host: nanoidp:9900` →
      `issuer: http://nanoidp:9900`; `Host: nanoidp` (no port) →
      `issuer: http://nanoidp` (none added)


dshvedchenko:
also i think we should combine it with #117 , what do you think ? @cdelmonte-zg

dshvedchenko:
already rebased on #117

PR #124, opened 2026-08-17
feat: Add first-class group support

dshvedchenko:
Users gain a groups list alongside roles, modelled exactly the same way: it is loaded from and persisted to users.yaml (omitted when empty), emitted as a groups claim on the access token and from /userinfo, requestable in the ID Token via the OIDC claims parameter, advertised in claims_supported, and flattened into authorities with the new GROUP_ authority prefix (default). Groups are editable from the user form, shown on the users list and detail pages, exposed by /api/users, and settable through the MCP create_user / update_user tools. Users without groups behave exactly as before.

examples/test_agent.py gains coverage: Token Decode now checks for the groups claim, and a new UserInfo Groups & Authorities test verifies /userinfo returns groups and the access token's authorities carry the GROUP_ prefix.


cdelmonte-zg:
Thanks, this is a clean and thorough mirror of the `roles` field across every surface (models, config, serialization, token/authorities, discovery, `/userinfo`, REST API, MCP, UI, and tests). Coverage in `test_groups.py` is genuinely complete, including the "groups must not leak into attributes" case and the voluntary-claim omission for empty groups.

One small thing worth fixing:

- `src/nanoidp/templates/test.html`: the token tester's authority preview builds `GROUP_${group}` without upper-casing, while the backend emits `GROUP_${group.upper()}` in `build_authorities`. For a lowercase group value the preview shows e.g. `GROUP_engineering`, while the real authority baked into the token is `GROUP_ENGINEERING`. A `.toUpperCase()` fixes it. The same quirk already exists for `ROLE_${role}` just above, so it would be nice to align both.

Optional: the `test_groups.py` module docstring points at SAML export coverage via `roles_attr_name` / `groups_attr_name` in `test_saml.py`. Those settings and that coverage live in the SAML export change (#125, the next PR in the stack), so the reference only resolves once that lands. If these merge together it is fine as is; otherwise it may be cleaner to move that docstring line to #125.

Otherwise this looks good to me.

dshvedchenko:
@cdelmonte-zg updated according to feedback

cdelmonte-zg (approved):
Approving. The upper-case fix for GROUP_ / ROLE_ in the token tester preview landed in 0d40646, coverage is complete, and CI is green. Thanks for the quick turnaround.

PR #125, opened 2026-08-17
feat: Add optional SAML export of roles and groups 

dshvedchenko:
feat: Add first-class group support

Users gain a groups list alongside roles, modelled exactly the same way: it
is loaded from and persisted to users.yaml (omitted when empty), emitted as
a groups claim on the access token and from /userinfo, requestable in the
ID Token via the OIDC claims parameter, advertised in claims_supported, and
flattened into authorities with the new GROUP_ authority prefix (default).
Groups are editable from the user form, shown on the users list and detail
pages, exposed by /api/users, and settable through the MCP create_user /
update_user tools. Users without groups behave exactly as before.
examples/test_agent.py gains coverage: Token Decode now checks for the
groups claim, and a new UserInfo Groups & Authorities test verifies /userinfo
returns groups and the access token's authorities carry the GROUP_ prefix


cdelmonte-zg:
Thanks, this is well-structured and heavily tested, with `groups` mirrored across the token, `/userinfo`, discovery, UI, REST, and MCP, and the SAML export wired to both the SSO assertion and the AttributeQuery endpoint. The off-by-default toggles keep existing behavior intact.

One thing worth fixing:

- `src/nanoidp/mcp_server.py` (`update_settings`): the `saml_roles_attr_name` / `saml_groups_attr_name` values are assigned onto `Settings` directly. Because the `Settings` model does not set `validate_assignment=True`, this bypasses the `normalize_saml_attr_name` validator that turns a blank name back into the default. A blank name passed through MCP (`saml_roles_attr_name=""`) is stored verbatim, and with export on the SAML layer then emits `<Attribute Name="">`. The UI and the YAML-load path both normalize blank to `roles` / `groups`, so MCP is the odd one out. Normalizing in the MCP path (or enabling `validate_assignment`) would align the three surfaces.

Two smaller notes, both non-blocking:

- The AttributeQuery path comma-joins roles/groups (`",".join(...)`) while the SSO assertion passes them as a list. The response builder splits comma strings back into multiple `AttributeValue` elements, so normal values render identically and this matches the existing `entitlements` convention. The only edge case is a role/group value that itself contains a comma, which the two endpoints would then represent differently. Worth a mention only if comma-bearing names are a concern.
- `resolve_user_claim` omits an empty `groups` claim, but `roles` always returns its list (so empty roles emit `"roles": []`). This looks deliberate and is pinned by a test; flagging only so the asymmetry is a conscious choice.

Otherwise this looks good, assuming it lands on top of #124 as the stack intends.

dshvedchenko:
@cdelmonte-zg updated according to feedback

PR #126, opened 2026-08-17
feat: add option to infer issuer from request for discovery/OIDC

dshvedchenko:
- **`oauth.issuer_from_request`** (off by default): when enabled, the
  discovery document's `issuer`, every minted token's `iss`, and the device
  flow's `verification_uri` are derived from the incoming request's own Host
  header instead of the fixed `oauth.issuer`. Lets the same NanoIDP be
  reachable under more than one hostname (e.g. a Docker Compose service name
  from other containers and `localhost` from the host browser) without a
  discovery/token issuer mismatch - each hostname advertises and issues
  tokens against itself. The MCP `get_oidc_discovery`/`get_settings` tools
  have no request of their own and always report the fixed `issuer`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,40 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Settings page and the MCP `update_settings` tool, and apply to both the SSO
   assertion and the AttributeQuery endpoint, with one `AttributeValue` per
   entry.
+- **`oauth.issuer_from_request`** (off by default): when enabled, the
+  discovery document's `issuer`, every minted token's `iss`, and the device
+  flow's `verification_uri` are derived from the incoming request's own Host
+  header instead of the fixed `oauth.issuer`. Lets the same NanoIDP be
+  reachable under more than one hostname (e.g. a Docker Compose service name
+  from other containers and `localhost` from the host browser) without a
+  discovery/token issuer mismatch - each hostname advertises and issues
+  tokens against itself. The MCP `get_oidc_discovery`/`get_settings` tools
+  have no request of their own and always report the fixed `issuer`.
+- **`oauth.issuer_allowlist`**: restricts `issuer_from_request` to a list of
+  allowed origins (e.g. `["http://localhost:8000", "http://nanoidp:9900"]`).
+  Empty (default) allows any Host header, unchanged from before; when set, a
+  request whose Host doesn't match falls back to the fixed `oauth.issuer`
+  instead of trusting an arbitrary Host header. Settable from the Settings
+  page and the MCP `update_settings` tool.
+- **`oauth.device_verification_base_url`**: pins the device flow's
+  `verification_uri` to a fixed, human-reachable URL (e.g.
+  `https://idp.example.com`), overriding `issuer_from_request`'s derivation
+  for that field only - discovery's `issuer` and a token's `iss` are
+  unaffected. Fixes a backend/container caller of `/device_authorization`
+  (e.g. `Host: nanoidp:9900`) otherwise leaking its own Host into a URL the
+  end user's browser can't open. Unset by default. Settable from the
+  Settings page and the MCP `update_settings` tool.
+- **`oauth.issuer_from_proxy_headers`** (off by default): trusts
+  `X-Forwarded-Proto`/`X-Forwarded-Host`/`X-Forwarded-For` from a single
+  reverse-proxy hop (via werkzeug's `ProxyFix`), so `issuer_from_request` and
+  rate-limit client IPs see the original scheme/host/client instead of the
+  proxy's own connection when TLS is terminated upstream. Only changes the
+  derived issuer/`iss`/`verification_uri` when `issuer_from_request` is also
+  on; the rate-limit effect applies regardless. Only enable this when
+  NanoIDP is deployed directly behind exactly one trusted proxy - these
+  headers are otherwise spoofable by any client. Readable/settable via the
+  MCP `get_settings`/`update_settings` tools; since `ProxyFix` is wired at
+  app startup, a value changed at runtime only takes effect after a restart.
 
 ### Changed
 - **Migrated the MCP server to the mcp 2.0 SDK** and pinned `mcp>=2,<3`. mcp 2.0
--- a/book/src/reference/configuration.md
+++ b/book/src/reference/configuration.md
@@ -51,6 +51,33 @@ server:
 
 oauth:
   issuer: "http://localhost:8000"
+  issuer_from_request: false    # true: derive issuer/iss/verification_uri from
+                                 # each request's own Host header instead of the
+                                 # fixed issuer above - lets the same NanoIDP be
+                                 # reachable under more than one hostname (e.g. a
+                                 # Docker Compose service name vs. localhost) without
+                                 # a discovery/token issuer mismatch. MCP tools have
+                                 # no request of their own and always report the
+                                 # fixed issuer.
+  issuer_allowlist: []          # origins allowed to be reflected by issuer_from_request,
+                                 # e.g. ["http://localhost:8000", "http://nanoidp:9900"].
+                                 # Empty (default) allows any Host header; a non-matching
+                                 # Host falls back to the fixed issuer above.
+  device_verification_base_url: null  # fixed, human-reachable URL (e.g.
+                                 # "https://idp.example.com") for the device flow's
+                                 # verification_uri, overriding issuer_from_request's
+                                 # derivation there - use this when a backend/container
+                                 # calls /device_authorization so the returned URL is
+                                 # still one a human's browser can open. Discovery's
+                                 # issuer and a token's iss are unaffected.
+  issuer_from_proxy_headers: false  # true: trust X-Forwarded-Proto/Host/For from a
+                                 # single reverse-proxy hop in front of NanoIDP
+                                 # (applies werkzeug's ProxyFix). Only affects the
+                                 # issuer_from_request derivation above - has no
+                                 # visible effect unless that's also on - but always
+                                 # affects rate-limit client IP attribution. Only
+                                 # enable behind exactly one trusted proxy - these
+                                 # headers are otherwise spoofable.
   audience: "my-app"            # access token "aud" (resource audience, RFC 9068)
   token_expiry_minutes: 60
   refresh_token_rotation: false # true: each refresh invalidates the used refresh token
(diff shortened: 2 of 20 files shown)


cdelmonte-zg:
Thanks. The `issuer_from_request` feature itself is clean and correctly off by default, the MCP tools are rightly excluded from request-based derivation, and it is well tested. Before merge, one change is needed and a few things are worth deciding.

Needs fixing:

- `config/settings.yaml` was committed as a settings-save round-trip rather than a hand edit, and it regressed the tracked default config: `issuer`, `saml.entity_id` and `saml.sso_url` lost their `${PORT:8000}` substitution (now hardcoded to `8000`, while `server.port` still uses `${PORT:8000}`), and the `registered-client` description was truncated from `(exact matching, issue #67)` to `(exact matching, issue`. Running `PORT=9900 python -m nanoidp` now listens on 9900 but advertises issuer `http://localhost:8000`, which is exactly the discovery/token mismatch this PR is meant to prevent. Please restore `config/settings.yaml` to a hand edit that only adds `issuer_from_request: false` (and any intended new keys), preserving the `${PORT}` placeholders, the `#67` reference, and the comments.

This one is broader than the committed file and probably deserves its own issue: the truncation at `#` and the loss of `${PORT}` / comments come from the settings-save path (yaml_writer / serialization) writing unquoted values, so any user who saves settings through the UI or MCP risks the same data loss in their own config. Quoting values on write would fix it at the source.

Worth deciding (secondary, and only in effect when the flag is enabled, which it is not by default):

- No `X-Forwarded-Proto` / ProxyFix handling: `request.host_url` yields `http://host` even when the external URL is `https://host`, so behind a TLS-terminating proxy the discovery `issuer` and token `iss` would advertise `http`, re-introducing a mismatch in exactly the multi-hostname setup this targets.
- Device flow: `verification_uri` is derived from the calling request's Host, so a backend container calling `/device_authorization` with `Host: nanoidp:9900` gets `http://nanoidp:9900/device`, which a human's browser on the host cannot resolve. The user-facing step breaks in the multi-hostname scenario.
- The reflected issuer is taken from the Host header with no allowlist, and `iss` is not validated server-side. Acceptable for an off-by-default dev-tool convenience, but a one-line doc note that the toggle trusts the incoming Host header and should only be used on trusted networks would be worth adding.

Happy to help with any of these.

dshvedchenko:
for Device flow: verification_uri is derived from the calling request's Host, so a backend container calling /device_authorization with Host: nanoidp:9900 gets http://nanoidp:9900/device, which a human's browser on the host cannot resolve. The user-facing step breaks in the multi-hostname scenario.

jwt verification should support match agaisnt several allowed issuer , this is exact what i added at backend in my services and jwt libraries support https://github.com/mpdavis/python-jose/blob/master/jose/jwt.py#L84

dshvedchenko:
for No X-Forwarded-Proto / ProxyFix handling: request.host_url yields http://host even when the external URL is https://host, so behind a TLS-terminating proxy the discovery issuer and token iss would advertise http, re-introducing a mismatch in exactly the multi-hostname setup this targets.

i think we should respect those options from header, so i can add it in this PR

dshvedchenko:
more on this : The reflected issuer is taken from the Host header with no allowlist, and iss is not validated server-side. Acceptable for an off-by-default dev-tool convenience, but a one-line doc note that the toggle trusts the incoming Host header and should only be used on trusted networks would be worth adding.

maybe it is good to have allowlist of issuer - to be close to industry standards

cdelmonte-zg:
Thanks, both good instincts. Here is where I would land on each, keeping NanoIDP's dev-tool scope in mind.

Yes to `X-Forwarded-Proto` / `X-Forwarded-Host`, please add it in this PR. It genuinely completes the feature: without it, `issuer_from_request` behind a TLS-terminating proxy advertises `http` while clients reached it over `https`, which is exactly the mismatch the option is meant to remove. Two asks to keep it contained and safe:

- Read `X-Forwarded-Proto` / `X-Forwarded-Host` inside `_effective_issuer`, and only when `issuer_from_request` is on (the operator has already opted into request-based derivation there). Prefer the forwarded values and fall back to `request.host_url` when they are absent. That keeps the added trust localized to this one path, rather than enabling a global `ProxyFix` that would affect every `request.*` in the app.
- Please note, in the config comment and the docs, that with the flag on NanoIDP now trusts `X-Forwarded-Proto` / `X-Forwarded-Host` from the immediate caller, so it should sit behind a trusted proxy or on a trusted network.

On the issuer allowlist: I would rather not add one. An allowlist is production-IdP hardening, and NanoIDP is deliberately a development and testing tool (see VISION.md: production use is a non-goal, hardening stays opt-in, defaults stay permissive). The whole point of `issuer_from_request` is to reflect whatever hostname reached it without the operator maintaining a list. For the "trusts the Host header" concern, the proportionate response for a dev tool is the one-line docs note that the toggle should only be enabled on trusted networks, not a validation layer. If a concrete need for an allowlist comes up later, it can be its own opt-in setting.

One item neither comment covered: with the flag on, the device flow's `verification_uri` reflects the calling request's Host, so a container calling `/device_authorization` with `Host: nanoidp:9900` produces a URL that a human's browser on the host cannot open. This does not need to block this PR; a short "known limitation" note in the docs (or a follow-up issue) is enough.

Thanks again for iterating on this so quickly.

cdelmonte-zg:
On the allowlist specifically: rather than lose the idea, would you mind opening a separate issue for it? That way it is tracked and we can pick it up if a concrete need shows up, without holding up this PR or growing its config surface.

The line I am steering by is that NanoIDP should stay usable with almost no configuration. That is a lot of its value as a dev tool: if getting a first token starts requiring an allowlist and a list of hostnames to maintain, it stops being the thing you reach for instead of standing up Keycloak. So I would want an allowlist to arrive only if and when someone actually hits a need for it, as its own opt-in, rather than by default.

Same offer for the device-flow `verification_uri` limitation: happy for that to be a tracked issue too if you would rather that than a short docs note inline.

dshvedchenko:
On the issuer allowlist : i would implement it and empty list will work as permission, only if there is items in allowlist - we will respond with issuer settings exactly - im almost completed this feature - it will be sad to rollback

dshvedchenko:
as those 3 points are all bound to issuer from request - i would like to have them implemented in one PR

cdelmonte-zg:
Sounds good, let's keep all three in this one PR, and go ahead with the allowlist as you designed it. Your empty-list-is-permissive default is what changes my mind: it keeps the zero-config path exactly as it is today (empty = reflect any Host, unchanged) and makes the allowlist a purely opt-in hardening knob, which is how the rest of NanoIDP's strictness works. So no need to split it out or open a separate issue, please carry on with what you have.

A few things to nail down so it lands cleanly:

- Contract: empty allowlist = current behavior (reflect the request host). Non-empty = reflect the host only if it is in the list, otherwise fall back to the configured `oauth.issuer`. Please cover all three in tests: empty reflects, a listed host reflects, an unlisted host falls back to the fixed issuer.
- It should compose with the `X-Forwarded-Host` handling: the host you match against the allowlist and the host you reflect should be the same resolved value (the forwarded header when present and the flag is on, otherwise `request.host_url`).
- Docs: describe it as optional hardening (default empty = permissive), useful when the instance is reachable by hosts you do not fully trust, alongside the note that the toggle otherwise trusts the incoming Host / forwarding headers.

One thing to keep in mind: the allowlist does not really fix the device-flow `verification_uri` point. If a container's hostname is in the list, a device flow started by that container still yields a URL that a human's browser cannot open, so a short "known limitation" note in the docs is still the right call there.

Thanks for pushing on this. The opt-in framing is the right shape.

dshvedchenko:
as of One thing to keep in mind: the allowlist does not really fix the device-flow verification_uri point. If a container's hostname is in the list, a device flow started by that container still yields a URL that a human's browser cannot open, so a short "known limitation" note in the docs is still the right call there.

i will add small ( off by default , override for device verififcation url )
Add a new optional setting (None by default = current behavior unchanged). When set, /device_authorization always builds verification_uri from this fixed, human-reachable URL instead of _effective_issuer(settings) — regardless of which internal Host called the endpoint. Discovery/iss keep following issuer_from_request as today (they must match the calling host per OIDC); only the device-flow's human-facing URL is decoupled. Minimal, additive, off by default, no breaking change.

cdelmonte-zg:
Yes, that works well, please add it in this same PR. Keeping it `None` by default so current behavior is unchanged, and decoupling only the device flow's human-facing URL while `iss` and discovery keep following `issuer_from_request`, is exactly the right split. Good call on preserving the issuer/discovery match; that is the part that has to keep reflecting the calling host.

So #126 ends up with four opt-in pieces (`issuer_from_request`, `X-Forwarded-Proto` / `X-Forwarded-Host`, the issuer allowlist, and this device verification URL override), all off or empty by default. Ping me once it is pushed and rebased on main and I will review the whole thing.

dshvedchenko:
implemented  four opt-in pieces (issuer_from_request, X-Forwarded-Proto / X-Forwarded-Host, the issuer allowlist, and this device verification URL override),

rebased on main and squashed

@cdelmonte-zg

cdelmonte-zg:
This looks great. Verified locally: CI is green, all four pieces are opt-in and off/empty by default, `config/settings.yaml` is clean (the `${PORT}` placeholders and comments are intact), and the settings are wired through config load, serialization, the MCP `update_settings` / `get_settings` tools, and the UI. The test coverage is thorough and hits exactly the contract points, including that the allowlist matches the forwarded origin and that the device override leaves `iss` / discovery alone.

On ProxyFix: you were right, the global `ProxyFix` behind its own off-by-default `issuer_from_proxy_headers` flag is the better call. It is the idiomatic mechanism and it makes the corrected `request.host_url` flow into the allowlist and device-url logic for free, which your `test_allowlist_matches_the_forwarded_origin` proves. Thanks for documenting the global effect (including the rate-limit client IP) honestly.

One thing before I merge: please add coverage in `examples/test_agent.py`. In this project a feature lands with its e2e agent test alongside the MCP and unit tests (same as the groups and SAML-export PRs did), so I would like a small check there too. For example, with `issuer_from_request` on, assert that the discovery `issuer` reflects the request Host and that a minted token's `iss` matches it. Once that is in I will approve and merge.

Thanks again, this came together well.

dshvedchenko:
@cdelmonte-zg ,  coverage in examples/test_agent.py added

cdelmonte-zg (approved):
Approving. The e2e coverage in examples/test_agent.py is in and is config-agnostic (reads /api/config, asserts the contract whether the toggle is on or off, and checks token iss matches the discovery issuer). All four pieces are opt-in and off/empty by default, config/settings.yaml is clean, parity across config/serialization/MCP/UI is complete, and CI is green. Thanks for the thorough iteration.

PR #128, opened 2026-08-19
fix: return lost UI for issuer_from_proxy_headers

dshvedchenko:
oauth.issuer_from_proxy_headers was added to the model/config/serialization/MCP layers but never exposed on the Settings page, so it was only configurable by hand-editing settings.yaml. This restores UI parity with the other issuer_from_request-related toggles.

Changes
Settings page: new "Trust Proxy Headers" switch under OAuth2/OIDC Settings, wired into the JS preview panel.
routes/ui.py / services/yaml_writer.py: update_oauth_settings() now accepts and persists issuer_from_proxy_headers.
models.py: field description updated to mention the Settings page (previously only referenced the MCP update_settings tool).
settings.yaml: restored ${PORT:8000} placeholders and the truncated #67 client description that had been lost in an earlier settings-save round-trip; reset issuer_from_request back to its default false.
Tests: TestIssuerFromProxyHeadersRoundTrip in test_ui_oauth_settings.py — toggle persists on, persists off when unchecked, and renders on the page.
Testing
Full suite: 888 passed.

### DIFF
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -26,6 +26,8 @@ oauth:
     - http://localhost:3000/callback
   issuer_from_request: false
   issuer_from_proxy_headers: false
+  require_pkce: false
+  refresh_token_rotation: false
 saml:
   entity_id: http://localhost:${PORT:8000}/saml
   sso_url: http://localhost:${PORT:8000}/saml/sso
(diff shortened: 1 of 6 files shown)


dshvedchenko:
@cdelmonte-zg overlooked missed UI control

cdelmonte-zg (approved):
Approving. Good self-catch on the missing Settings-page control for issuer_from_proxy_headers. Verified: the switch is wired through update_oauth_settings and yaml_writer, the form-text is honest (rate-limit IP effect, single-trusted-proxy caveat, takes effect on restart), tests cover the round-trip and rendering, config/settings.yaml is untouched, and CI is green. Thanks.

Issue #134, opened 2026-08-19
SAML export: attr-name collision overwrites roles; AttributeQuery should pass lists


cdelmonte-zg:
Two edge cases in the #125 SAML roles/groups export, both P3/minor correctness (not security), found in a post-merge review:

1. **Attribute-name collision silently drops roles.** With both exports enabled and `saml_roles_attr_name` == `saml_groups_attr_name` (e.g. both `memberOf`), the code writes roles first and groups immediately after into the same dict, so groups fully replace roles. Validation prevents a blank name but not two equal names. Options: reject equal names when both exports are enabled, or (likely better) merge the two lists into the single shared SAML attribute - `memberOf` carrying both is a common real-world shape.

2. **AttributeQuery comma-join splits comma-bearing values.** The AttributeQuery path does `",".join(user.roles)` while the SSO assertion passes the raw list, so a legitimate value like `"Finance, EMEA"` becomes two AttributeValues in the AttributeQuery response but one in the SSO assertion. The response builder already handles `list`/`tuple` natively (one AttributeValue per entry), so passing the lists directly removes the divergence. Note the same pattern pre-exists for `entitlements`; aligning it at the same time would be consistent.

dshvedchenko:
i can look

cdelmonte-zg:
Thanks for the offer! I am actually picking this one up right now, so no need - the direction is: merge the two lists into the single shared attribute when the configured names collide (deduplicated, roles first), and pass the lists directly to the AttributeQuery builder instead of comma-joining (it already emits one AttributeValue per entry). A review on the PR once it is up would be very welcome.

dshvedchenko:
ive done this pr - so you can pick it or discard

PR #136, opened 2026-08-19
fix: SAML export: attr-name collision overwrites roles; AttributeQuer…

dshvedchenko:
…y should pass lists

#134

### DIFF
--- a/src/nanoidp/routes/saml.py
+++ b/src/nanoidp/routes/saml.py
@@ -8,7 +8,7 @@ import uuid
 import zlib
 from base64 import b64decode, b64encode
 from datetime import datetime, timedelta, timezone
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from flask import Blueprint, Response, abort, render_template, request, session
 from flask.typing import ResponseReturnValue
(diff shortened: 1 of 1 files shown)


cdelmonte-zg:
Sorry, we crossed paths here - I had just claimed #134 (see my comment there) and the fix is now up as PR #137: shared merge helper for colliding attribute names (roles first, deduplicated) plus lists passed straight to the AttributeQuery builder, docs and tests included. Closing this one as a duplicate of #134; a review on #137 would be very welcome!

cdelmonte-zg:
Correcting my earlier comment: this is of course your PR, not a duplicate issue - I closed it too hastily and mislabeled it, sorry about that. I have read it properly now: your `_build_roles_groups_attrs` shape is clean and functionally equivalent to what landed in #137 (the two implementations crossed within minutes of each other). #137 additionally carries the tests (collision merge asserted in both SSO and AttributeQuery, plus a comma-bearing value staying a single AttributeValue), the docs note, the CHANGELOG entry, and deduplication on the merged list, so per your own "pick it or discard" I am keeping that one. Thanks for the quick turnaround, and sorry again for the crossed wires.

PR #138, opened 2026-08-19
Saving settings preserves env placeholders and empty optional values

dshvedchenko:
This fixes the remaining regression in the settings save path.

Preserve raw ${VAR:default} placeholders when an unrelated field changes
Avoid rewriting the whole oauth.clients list when only a sibling client changes
Merge client entries by client_id and update fields individually so unchanged env-backed secrets stay in the YAML instead of being materialized
Treat empty optional values such as ${DEVICE_URL:} as unchanged when they still resolve to an empty string, instead of dropping the key from the file
Add regression tests covering both the env-backed client secret case and the empty-default placeholder case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   real client.
 
 ### Fixed
+- **Saving settings no longer discards comments, inline `#` text or `${VAR:default}`
+  placeholders in `settings.yaml`** (#127): the settings writer now round-trips
+  the file with `ruamel.yaml` (comments and quote style survive) and only
+  rewrites a field when its expanded on-disk value actually differs from the
+  new one, so an untouched `${PORT:8000}`-style placeholder is no longer
+  replaced by its resolved value on the next save. Free-form text
+  (`description`, `client_secret`, `password`, attribute values) is now quoted
+  on write so an embedded `#` can't be mistaken for a comment. Applies to both
+  the web UI settings form and the MCP `save_config` tool.
+- **Env-backed client secrets and empty optional placeholders are preserved when
+  unrelated settings change.** The OAuth client merge now updates entries by
+  `client_id` field-by-field instead of rewriting the whole list, so an
+  unchanged `${APP1_SECRET:dev}` secret stays in the raw file even when a
+  sibling client is edited. Empty optional values such as
+  `${DEVICE_URL:}` are also treated as unchanged when they still expand to an
+  empty string, instead of being popped out of the YAML on a save that changed
+  some other field.
 - **`/api/config` now exposes `issuer_allowlist`, `device_verification_base_url`
   and `issuer_from_proxy_headers`** alongside `issuer_from_request`. The
   config-agnostic e2e agent reads the allowlist from `/api/config` to predict
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,7 +86,7 @@ The test agent covers:
 - **Key Management**: Key info, rotation, post-rotation token validation
 - **REST API**: Users, config, audit log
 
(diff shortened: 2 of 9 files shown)


dshvedchenko:
@cdelmonte-zg

cdelmonte-zg (changes requested):
The overall approach is right and CI is green on `f1cf8b9` (full suite also passes locally). However, there is one blocker that contradicts the PR's own goal, plus a smaller related issue. Requesting changes for those.

## 1. Blocker: env-resolved secrets get materialized into `settings.yaml` via the client list

The unchanged-field guard works per key for scalars, but `oauth.clients` is compared and rewritten as a whole list:

```python
new_clients = [client_to_yaml(client) for client in settings.clients]
if not is_unchanged(oauth.get("clients"), new_clients):
    oauth["clients"] = new_clients
```

As soon as any element of the list differs, the entire list is rebuilt from the in-memory values, which are already env-expanded. Reproduced on `f1cf8b9` with:

```yaml
clients:
- client_id: app1
  client_secret: ${APP1_SECRET:dev}
  description: first
- client_id: app2
  client_secret: foo
  description: second
```

With `APP1_SECRET=SUPERSECRET` set, editing only `app2.description` and calling `ConfigManager.save()` produces:

```yaml
- client_id: app1
  client_secret: 'SUPERSECRET'
```

So the placeholder is not just lost: a secret coming from the environment is written to disk in cleartext.

The same happens through the UI path: `save_client()` replaces the whole entry with `client_to_yaml(client)` unconditionally, and the edit route keeps the already-expanded `client.client_secret` when the secret field is left blank (`routes/ui.py:400-403`), so editing a client's description in the web UI also materializes its `${VAR}` secret. Reproduced as well.

A side effect of the full-list rewrite: untouched sibling entries get cosmetic churn too (`client_secret: foo` becomes `client_secret: 'foo'`).

### Suggested fix

Merge clients field by field, keyed on `client_id`, instead of replacing the list:

- for each client in `settings.clients`, find the raw entry with the same `client_id` and update each field only when `is_unchanged(raw_entry.get(field), new_value)` is false; in particular the raw `client_secret` must stay untouched when `is_unchanged(raw_entry["client_secret"], client.client_secret)` holds;
- still handle additions (append entries for new `client_id`s) and removals (drop raw entries whose `client_id` is no longer in `settings.clients`);
- apply the same per-field merge inside `save_client()` when the client already exists, instead of `clients[existing_idx] = client_data`.

Please also add a regression test with an environment variable actually set (not only `monkeypatch.delenv`), e.g. `monkeypatch.setenv("APP1_SECRET", "SUPERSECRET")`, edit a different client, save, and assert both that `${APP1_SECRET:dev}` is still in the raw file and that `SUPERSECRET` is not.

## 2. Placeholders resolving to an empty string in optional fields are still deleted

```yaml
device_verification_base_url: ${DEVICE_URL:}
```

With `DEVICE_URL` unset this expands to `""`, and the model keeps `""` in memory. `apply_settings_document()` compares against `settings.device_verification_base_url or None`, so the comparison is `"" != None`, the guard fails, and the falsy branch does `pop()`: the key (and the placeholder) disappears from the file on a save that changed something else entirely. Reproduced on `f1cf8b9`.

Suggested fix: normalize the comparison to `""` instead of `None` (or, more simply, when `is_unchanged` is true leave the key completely alone, never `pop` it). Same consideration applies to the equivalent empty-means-clear branches in `YamlWriter.update_oauth_settings()`.

## 3. Cosmetic: PR title and body are truncated

The title is `- **Saving settings no longer discards...` and the body starts mid-sentence with `${VAR:default}` - it looks like the commit message got split by `gh pr create`. Please fix the title/body so they read as a normal PR description.

---

After the field-by-field client merge, the empty-default fix and the env-set regression test, this should be ready to go.

cdelmonte-zg (approved):
LGTM, thank you! Verified the full round empirically: env-backed secrets stay raw, empty-default placeholders survive, comments intact, suite green.

Issue #143, opened 2026-08-20
Refactor: extract shared merge_client_entry() helper for the per-field OAuth client merge


cdelmonte-zg:
Follow-up to #138: the per-field client-merge logic (update a raw YAML client entry from an OAuthClient, guarded by is_unchanged(), quoting free-form text, dropping emptied optional list fields) now exists in two copies:

- serialization.merge_oauth_clients() (full-list merge used by apply_settings_document)
- the inline block in YamlWriter.save_client() for the existing-client case

The two copies must stay behaviorally identical, so any future change to the merge rules has to be made twice. Extract a shared merge_client_entry(raw_entry, client) helper in serialization.py and use it from both call sites. Pure refactor, no behavior change; existing #127 round-trip tests cover it.

Natural carrier: the #142 hooks work touches the same save path, so this can ride along there.

cdelmonte-zg:
Done in #144: extracted serialization.merge_client_entry(), shared by merge_oauth_clients() and YamlWriter.save_client(). Ships in 2.6.0.

PR #144, opened 2026-08-20
Consolidated the OAuth client YAML merge logic into a single

dshvedchenko:
`serialization.merge_client_entry()` helper, shared by the settings save
  path (`merge_oauth_clients()`) and `YamlWriter.save_client()`'s web UI
  edit path, which previously duplicated the same field-by-field merge
  rules. Internal cleanup, no behavior change.
#143


cdelmonte-zg (approved):
LGTM, thank you! Verified behavior is unchanged (round-trip repros + full suite green); clean consolidation into merge_client_entry().

Issue #139, opened 2026-08-20
Import crashes with PackageNotFoundError when package metadata is absent


cdelmonte-zg:
`src/nanoidp/__init__.py` resolves the version unguarded:

```python
from importlib.metadata import version

__version__ = version("nanoidp")
```

When the source tree is used without an installed distribution (e.g. copied into a container image without `pip install`, or vendored), `importlib.metadata.version()` raises `PackageNotFoundError` at import time and the whole package becomes unusable. A downstream deployment ran into exactly this and had to patch the file.

Proposed fix: wrap in `try/except PackageNotFoundError` and fall back to a static version string (kept in sync by the release process), so importing the package never fails just because distribution metadata is missing.

dshvedchenko:
you can read version from myproject.toml as fallback

cdelmonte-zg:
Reading `pyproject.toml` would indeed cover exactly the case where the fallback matters (running from a source tree, where the file is present). The catch is the parser: `tomllib` is stdlib only from Python 3.11 and this project supports 3.10, so it would mean either a conditional `tomli` dependency or hand-rolled parsing - more machinery than this papercut warrants.

So the plan is the boring version: try `importlib.metadata.version()`, fall back to a static `"X.Y.Z"` string kept in sync by the release process. Zero dependencies, and the import can never fail. The version being potentially stale in an unreleased source tree is acceptable for a fallback whose only job is to keep the package importable.

If you want to pick this one up, go ahead - it is a one-file change plus a test that imports the package with the distribution metadata absent.

dshvedchenko:
myproject.toml version can be read with plain open file.- no toml lib - as tool already loads it as python code inside we can try to do minimal work to be consistent

cdelmonte-zg:
Done in #146: version() guarded with a best-effort pyproject.toml fallback, so importing from an uninstalled source tree no longer raises. Ships in 2.6.0.

PR #146, opened 2026-08-20
fix: #139 - crash whe metadata packegae absent

dshvedchenko:


cdelmonte-zg (approved):
LGTM, thank you! Verified both paths: installed metadata and the pyproject fallback when metadata is absent. Good call keeping the source-tree version accurate.

Issue #150, opened 2026-08-20
enhancement: ui branding per client


dshvedchenko:
allow to set 

- [x] background color selection
- [x] header color selection
- [x] footer color selection
- [x] show client id and client description
- [x] show client inline logo - as file from folder on nanoidp filesystem ( filename == client_id )

cdelmonte-zg:
Thanks for this - the underlying use case is worth taking seriously. Beyond automated testing, nanoidp is genuinely useful for local demos and prototyping a login experience, and making the login page show *which* client you're authenticating to fits that well. I've recorded that as a supported secondary use in VISION, together with a principle for how to do it safely (#153).

There's one firm constraint on *how*, and it's a security one. Anything a user can configure that then gets rendered into the login page has to be structured data or a local asset - never free-form CSS or markup, and never a remote URL fetched into the page:

- **Per-client CSS** (background/header/footer) is user-controlled markup rendered into the auth UI: that's a stored-XSS surface on the login page, which for a security-adjacent tool is exactly the wrong thing to add.
- **A remote logo URL** is milder but still two problems: attribute-injection if the templating slips, and a third-party beacon - every visitor's browser calls a URL the client operator chose, leaking IP/UA/timing from the auth page.

Keycloak, for comparison, does support per-client login theming, but as operator-deployed theme *files*, not CSS or URLs pasted into config - precisely to avoid the above.

So here's what I'd like to land instead, and I'd happily review a PR for it:

- **In scope now**: show the client's `client_id` and `description` on the login page (data already in `settings.yaml`, zero new surface).
- **In scope, done safely**: an optional per-client logo as a locally-served asset (a file the operator provides), referenced by name - not an arbitrary remote URL.
- **Out of scope**: arbitrary per-client CSS / header / footer markup.

This is post-2.6.0 work (that release is a security + fixes release). If you want to take the id/description part first, it's small and self-contained. Does this direction work for you?

dshvedchenko:
Great !

cdelmonte-zg:
Shipped in #157 (merge 73160c7): per-client hex colors, client_id/description toggles on the login page, and local-asset logos under static/logos/ (overridable via oauth.logos_dir). Per-client free-form CSS stays out of scope per VISION principle 7.

PR #157, opened 2026-08-20
feat: Per-client login page branding

dshvedchenko:
optional per-client colors (background,
  header, footer, all as validated hex values), show/hide client_id and
  description on the `/authorize` login page, and per-client logo images
  stored locally in `static/logos/` keyed by client ID (no YAML config needed
  for logos; place the image file and it's served automatically; the
  directory is overridable via `oauth.logos_dir`). Colours and toggles are
  editable from the OAuth client form in the UI and from the MCP
  `create_client`/`update_client`/`get_client` tools, descriptions are
  already supported, and logos are deployed by the operator to the server
  filesystem. Designed for demos and prototyping; colours are structured
  (not free-form CSS) to prevent stored-XSS on the auth UI, and logos are
  local files only (no remote URLs) to avoid beacons.
implementation of #150

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the security caveats inline.
 
 ### Added
+- **Per-client login page branding**: optional per-client colors (background,
+  header, footer, all as validated hex values), show/hide client_id and
+  description on the `/authorize` login page, and per-client logo images
+  stored locally in `static/logos/` keyed by client ID (no YAML config needed
+  for logos; place the image file and it's served automatically; the
+  directory is overridable via `oauth.logos_dir`). Colours and toggles are
+  editable from the OAuth client form in the UI and from the MCP
+  `create_client`/`update_client`/`get_client` tools, descriptions are
+  already supported, and logos are deployed by the operator to the server
+  filesystem. Designed for demos and prototyping; colours are structured
+  (not free-form CSS) to prevent stored-XSS on the auth UI, and logos are
+  local files only (no remote URLs) to avoid beacons.
 - **First-class group support**: users gain a `groups` list alongside `roles`,
   modelled exactly the same way. It is loaded from and persisted to
   `users.yaml` (omitted when empty), emitted as a `groups` claim on the access
--- a/book/src/reference/configuration.md
+++ b/book/src/reference/configuration.md
@@ -96,6 +96,15 @@ oauth:
       description: "Client whose redirect_uri is pinned"
       redirect_uris:            # optional; when set, /authorize enforces
         - "http://localhost:3000/callback"  # exact string matching
+    - client_id: "branded-client"
+      client_secret: "secret"
+      description: "Demo client with custom login page branding"
+      background_color: "#2c3e50"  # optional; hex only, behind the login card
+      header_color: "#3498db"      # optional; hex only, the card's header band
+      footer_color: "#e8f4f8"      # optional; hex only, the card's footer band
+      show_client_id: true         # optional; default true
+      show_description: true       # optional; default false
+  # logos_dir: "./static/logos"    # optional; defaults to src/nanoidp/static/logos
 
 saml:
   entity_id: "http://localhost:8000/saml"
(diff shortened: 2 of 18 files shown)


cdelmonte-zg (changes requested):
Reviewed with a full local pass (945 tests, ruff, mypy all green) plus a live run: branded client rendered correctly on /authorize (all three colors, client_id and description toggles), CSS-injection attempts via the UI form rejected on every path, and the validation chokepoint holds - every write path (YAML load, UI forms, MCP create/update) goes through the validating `OAuthClient` constructor or `validate_assignment=True`, with the MCP pre-check failing before any field is mutated. Path traversal on the logo lookup is blocked by the client-id charset whitelist. Good, disciplined implementation of VISION principle 7.

On scope: the hex colors go beyond what #150 originally offered (client_id/description + local logo), but validated hex is exactly the "structured data" principle 7 calls for, so they are accepted as part of this PR.

Findings - the first two need changes, the rest are cleanups.

**Blocking**

1. **The `oauth.logos_dir` override is broken end to end.** `resolve_client_logo()` honours it for the filesystem lookup, but the URL is always built with `url_for("static", ...)`, which serves only the app's own `static/` folder. Reproduced live: with `logos_dir` pointing at a custom directory containing `branded-client.png`, the page renders `` and that request 404s - a broken image on the login page. The knob silently works only when set to the app's own default. Fix: a dedicated route (e.g. `send_from_directory` over the resolved `logos_dir`, with the same client-id charset validation) used for both the default and the override - or drop the knob entirely. The `clients_form.html` help text should then also render the effective directory instead of the hardcoded `src/nanoidp/static/logos/`.

2. **`examples/test_agent.py` is not updated.** Project rule: features ship with MCP + e2e agent coverage in the same PR (the MCP side is done, with tests). A small scenario is enough: create a client with colors/toggles via the API, GET /authorize, assert the hex values and the id/description visibility in the HTML, then clean up.

**Cleanups**

3. **The hex pattern now exists in four copies**: three `Field(pattern=...)` literals in `models.py` plus `_HEX_COLOR_RE` in `mcp_server.py`. Define it once (e.g. `HEX_COLOR_PATTERN` in `models.py`) and reference it from all four places, so the MCP pre-check and the model can never diverge on what a valid color is.

4. **Please rebase onto current main** - the branch is based on pre-2.6.0 main. While at it, keep the `config/settings.yaml` diff minimal: the two `issuer_from_request`/`issuer_from_proxy_headers` lines were moved above the clients list (churn with no behavior change). The new `branded-client` demo entry itself is welcome - it makes the feature visible out of the box.

5. Optional: `resolve_client_logo` uses `re.match` with `^...$`; `re.fullmatch` (or `\Z`) avoids the trailing-newline edge of `$`. Harmless today (a newline can't produce an existing file), but `fullmatch` says what is meant.

dshvedchenko:
@cdelmonte-zg addressed, thank you for feedback

cdelmonte-zg (approved):
Re-verified at 97f49fd, empirically: worktree venv, 946 unit tests, ruff and mypy clean, live server, full e2e 48/48.

Both blocking findings are resolved:

- **logos_dir override works end-to-end.** With `oauth.logos_dir` pointing at an out-of-tree directory, `/authorize` renders the logo and the new `/client-logos/<client_id>` route serves the file from the override (HTTP 200). Traversal probes against the route (plain and URL-encoded `..`, backslash variant, null byte, extensionless name matching a file outside the dir) all return 404.
- **The e2e branding test is in** and passes against a live server as part of the 48/48 run (create via UI form, colors + id + description asserted on /authorize, cleanup).

Also verified on this head: invalid colors are still rejected on every write path with nothing persisted (UI create/edit probes with `;}body{display:none` and `url(...)` payloads left the file byte-identical); the logos_dir settings round-trip sets and clears the key without touching any other field; `HEX_COLOR_PATTERN` is shared between the model and the MCP pre-check; the branch is on current main with no settings.yaml churn.

One nit, fine to fix at merge: the logo help text in `clients_form.html` contains an em dash ("on the server — no configuration needed"); house style uses " - ".

cdelmonte-zg:
Merged. Thanks for turning the branding request into something that fits the project's principles.

Issue #156, opened 2026-08-20
Feature Request: Passwordless login


gprossliner:
First I want to say thank you for this very nice project. Read you redit, and was in the same situation: Configuring keycloak for local dev seems to be overkill and specially not self-contained. With nanoidp I have everything needed for devs in git. The UI is a bonus feature so I don't have to write yaml manually.

Before I found your project, I was considering writing something myself, reduced to what I actually need. I would have planed it in a way that it don't require passwords to be given, but with predefined "personas" that defines claims I need to test various different functions in the app.

So my suggestion would be:
* Make password optional, maybe in combination with a "passwordless" boolean flag
* If there are just passwordless users defined, have a dropdown list of the configured users instead of a text / password input on login screen

This would enabled developers so see directly which users are created for test, without checking the configuration files / project docs.

What do you think about this feature?
I am willing to contribute PR also if requested.

Thank you!

cdelmonte-zg:
Thanks, and glad it fits your workflow - this is exactly the kind of testing convenience nanoidp is for, so yes to the feature and yes to a PR (targeting 2.7.0; 2.6.0 is about to be tagged).

To save you from guessing at design decisions, here's the contract I'd like it to follow:

1. Opt-in, server-level mode, off by default - e.g. `login.mode: persona` (default `password`), rather than a per-user flag. A per-user `passwordless` boolean alongside a password creates contradictory states and an awkward mixed-user UI; a server mode keeps the behavior explicit.
2. `User.password` becomes optional. A user without a password works in persona mode and cannot authenticate via password mode or the OAuth password grant.
3. Persona mode authenticates by selecting the user in the interactive login UI; the configured users are listed instead of prompting for a password.
4. The OAuth password grant is unchanged. A missing password is not an empty credential - `authenticate()` simply does not authenticate a password-less user. Persona login is about interactive authentication, not the `/token` grant, so it does not change OAuth grant semantics.
5. `oauth21` is orthogonal and stays independent: it governs OAuth protocol strictness, not how the authorization server authenticates the resource owner.
6. The behavior should be consistent across the interactive user flows: OIDC `/authorize`, SAML `/saml/sso`, nanoidp `/login`, and device authorization. One important detail: SAML currently always emits `AuthnContextClassRef = PasswordProtectedTransport`. After a persona login that would be untrue, so persona authentication should use an appropriate context such as `urn:oasis:names:tc:SAML:2.0:ac:classes:unspecified`.
[...]

Issue #149, opened 2026-08-20
Enforce architectural invariants with import-linter (currently comment-only)


cdelmonte-zg:
Several architectural invariants are currently enforced only by comments, so nothing stops a future change (ours or an external contributor's) from silently violating them. The clearest example is the module docstring of `serialization.py`:

> This module deliberately has no runtime imports from the package (models are type-checking-only), so it can be imported from `config.py` without cycles.

If someone adds `from .services import ...` at runtime there, the import cycle comes back and no test catches it - the guarantee lives in prose.

Proposal: adopt **import-linter** (dependency-contract enforcement, declarative TOML, runs in CI next to ruff/mypy) and pin the invariants we already rely on. One tool only - deliberately not PyTestArch on top, to avoid two overlapping ways to express the same import rules.

### Contracts to encode

Current intra-package import structure (verified):

- `routes` -> `services`, `config`
- `services` -> `config`, `serialization`
- `config` -> `models`, `serialization`
- `serialization` -> `config` **only under `TYPE_CHECKING`** (no runtime package import)

So, two contracts:

1. **serialization stays runtime-standalone** (the docstring invariant):

   ```ini
   [importlinter:contract:serialization-standalone]
   name = serialization has no runtime package imports
   type = forbidden
   source_modules = nanoidp.serialization
   forbidden_modules =
       nanoidp.services
       nanoidp.routes
       nanoidp.config
   ```

2. **layering** routes -> services -> config -> serialization/models:

   ```ini
   [importlinter:contract:layers]
   type = layers
   layers =
       nanoidp.routes
       nanoidp.services
       nanoidp.config
   ```

### Known nuance to handle

`serialization.py` imports `config` under `TYPE_CHECKING`, and `config` imports `serialization` at runtime - a type-checking-only cycle that is fine at runtime. import-linter/grimp include `TYPE_CHECKING` imports in the graph by default, so contract 1 would false-positive on that import unless we tell it to ignore type-checking imports (grimp's `include_external_packages`/typing handling, or an `ignore_imports` line for exactly that edge). The task must confirm the config uses the "ignore TYPE_CHECKING" behavior so the contract reflects *runtime* dependencies, which is the whole point of the invariant.

### Scope

- Add `import-linter` to the dev dependencies and a `.importlinter` config.
- A CI step (`lint-imports`) in the Tests workflow, same job style as ruff/mypy.
[...]

PR #158, opened 2026-08-20
feat: Personas login mode

gprossliner:
Issue see #156

With help of my friend, Claude Sonnet 5, I was able to provide the PR with the all the points considered that you wrote in the Design contract. I left the planing markdown `./docs/plans/persona-login-mode.md` file in the branch, so you can check how we implemented this. It also contains a proposed CHANGELOG entry.

I tried to introduce no breaking changes, behavior or UI changes if persona mode is not enabled.

There maybe an existing bug that occured during e2e testing. I have not confirmed / analysed this further, see plan doc. I can analyse this an create a separate issue once confirmed.


cdelmonte-zg (changes requested):
Reviewed with a full local pass (971 unit tests, ruff, mypy) plus a live run: e2e agent 48/48 including the new Persona category, and manual negative probes on all four surfaces. The design contract holds up empirically: persona is gated server-side everywhere (no username-only bypass in password mode), `authenticate()` rejects password-less users on the password grant including the bcrypt path, and SAML emits `unspecified` for inline persona login and for a reused persona dashboard session while password logins keep `PasswordProtectedTransport`. Settings round-trip follows the omit-at-default and `is_unchanged` conventions.

Findings below - the first six are blocking.

**Blocking**

1. **CI: mypy fails.** `src/nanoidp/routes/ui.py:262`: `password` is inferred `str` from `request.form.get("password", "")` and is then assigned `user.password`, which is now `Optional[str]`. The behavior is right (blank on edit keeps the existing value, including "no password"); only the annotation is missing.

2. **Device flow, persona branch: Enter authorizes the first listed user.** The persona form in `templates/device.html` puts the `user_code` text input and one submit button per user (`name="username"`) in the same form. HTML implicit submission (pressing Enter in the text field) activates the first submit button, and `routes/oauth.py:1429` defaults `action` to `"authorize"`, so typing a code and hitting Enter - password-mode muscle memory - silently authorizes the pending device grant as whichever user happens to be listed first. Make authorization require an explicit selection (e.g. per-user buttons carry an explicit action value and a missing/unknown action re-renders the form).

3. **`POST /settings` persists an invalid `login_mode` and bricks the server.** `update_login_settings()` in `services/yaml_writer.py` writes any non-default string to disk first and only then hits validation via `get_config().reload()`. Reproduced live: posting `login_mode=banana` returns 302 and writes `login: mode: banana`; the next `python -m nanoidp` dies with the `Settings.validate_login_mode` ValidationError until the file is hand-edited. The MCP path is protected by its input_schema enum; the UI path validates nothing. Validate against `{password, persona}` before writing, and treat a blank value as "unchanged" rather than persisting it.

4. **A bare `login:` line in settings.yaml crashes the loader and both writers.** `yaml.safe_load("login:\n")` yields `{"login": None}`, so `config.py`'s `data.get("login", {})` returns `None` (the key exists) and `login.get("mode", ...)` raises AttributeError at startup; the `setdefault("login", {})` writes in `yaml_writer.py` and `serialization.py` likewise TypeError on a null section. The commented `# login:` block in the new configuration docs invites exactly this partial uncomment. Normalize with `data.get("login") or {}` on every read - and consider extracting the omit-at-default merge logic (currently duplicated between `serialization.apply_settings_document` and `yaml_writer.update_login_settings`) into one shared helper.

5. **`examples/persona-login/settings.yaml` ships `host: "0.0.0.0"`.** This is the one preset where the binding matters most: one-click login as any listed user, admin included, exposed to the whole network. As of today's 2.6.0 release (GHSA-2473-px8h-rvg6) the project default is `127.0.0.1`; use that here. Related: the branch is based on pre-release main - please rebase onto current main (that also explains the `NanoIDP v2.5.0` banner when running the branch).

6. **`user_create` now strips passwords in both modes.** `request.form.get("password", "").strip() or None` changes default-mode behavior: a password of `' secret '` is stored as `'secret'` and the user then fails the exact compare in `config.authenticate()`; a whitespace-only password flips from stored-verbatim to "Password is required"; and `user_edit` still stores the raw value, so create and edit disagree on the same input. Keep the persona semantics without touching password mode: treat the field as persona-blank only when `raw.strip()` is empty, otherwise store `raw` unchanged.

**Cleanups (non-blocking, but please address in this PR)**

7. **The persona-vs-password decision is hand-copied into four surfaces** (`oauth.py` `authorize()`, `saml.py` `sso()`, `ui.py` `login()`, `services/device_code.py` `verify()`), and the copies have already drifted: only two of them record `session["auth_method"]`, and `DeviceCodeStore.verify` grew a `persona_mode`/`get_user` parameter pair whose `persona_mode=True, get_user=None` combination silently reports invalid credentials. The plan's own "single choke point" argument for `authenticate()` applies here too: a config-level `interactive_authenticate(username, password)` consulting `persona_mode_enabled` would reduce the four call sites to one-liners and keep future persona rules single-site.

8. **The picker markup is triplicated** across `login.html`, `authorize.html`, `device.html` (and `device.html` repeats the entire form per mode branch, `user_code` input included). A shared Jinja macro for the picker, with only the credential block branched per template, means a fix like item 2 lands once instead of three times.

9. **Editor reformatting noise.** The five templates were auto-rewrapped (attribute wrapping, JS re-indentation, CSS spacing) and all five lost their trailing newline; the ~60 functional lines are buried in several hundred cosmetic ones and blame is broken for every touched line. Please restore the original formatting (and final newlines) outside the persona blocks.

10. **`create_persona_user`'s input_schema hand-duplicates `create_user`'s property definitions.** The handler side is deduped via `_build_user_from_arguments`, but the two Tool schemas remain independent copies that can drift - and neither declares `attributes`, which the shared builder reads. A shared properties dict spread into both schemas closes the gap.

11. **Em dashes.** The three new UI strings (`login.html:78`, `authorize.html:110`, `device.html:110`) use an em dash; the project uses plain " - " everywhere. Same throughout `docs/plans/persona-login-mode.md` - which per its own header should be dropped from the branch before merge anyway.

12. Optional test nit: one test asserting persona mode still works under `security_profile: oauth21` would pin the orthogonality the contract calls out.

On the reload observation in the plan: confirmed, and it is pre-existing on main. `_load_users()` never clears `self.users`, so `reload()` merges instead of replacing and a user deleted on disk survives in memory (`/api/users` still lists it; settings and clients are rebuilt fresh, so only users are affected). Please do file it as a separate issue - your analysis was correct.

CHANGELOG: leaving it to the maintainer was the right call; the proposed entry reads well and will be added at merge.

gprossliner:
Thank you for the tests, I'll fix this. 

#### Some comments

> Related: the branch is based on pre-release main - please rebase onto current main 

I created the folk yesterday, and based the feature branch on `main`, so it was current yesterday? I assume I got the correct branch, but should rebase so it includes the 3 commits made since then?

> 7. The persona-vs-password decision is hand-copied into four surfaces

I understand this, I just didn't know where to place common code. I think I will introduce a new personas helper file for this.

> 8. The picker markup is triplicated

I notices this on my own review. I just have not found existing marcros used in the tree, so I decided not to introduce new "concepts" into the tree. I will refactor that into a shared macro.

#### GH Advanced Security Checks

I think this are considered false positives / unreleated to the PR? The verbose logs only include `username` and `client_id` which are not considered secret values, and the redirect to `callback_url` may be unvalidated here, but is part of the protocol (maybe validated before), or at least unrelated to the PR.

Should I ignore them?

cdelmonte-zg:
Answers in order:

**Rebase**: your branch was indeed current when you cut it - but 2.6.0 was released this morning, so main gained three commits since: the security fix changing the default bind address to 127.0.0.1 (GHSA-2473-px8h-rvg6), the release housekeeping (version 2.6.0), and a docs follow-up. Rebasing picks up the new loopback default (which is what finding 5 about the example preset relates to) and fixes the "NanoIDP v2.5.0" banner when running the branch. It should apply cleanly or nearly so.

**Where to put the common auth code**: no new file needed - add it to `ConfigManager` in `config.py`, right next to `authenticate()`. Something like `interactive_authenticate(username, password)` that consults `settings.persona_mode_enabled` and returns the user (or `None`), so the four surfaces become one-liners and the persona rules live beside the password rules they mirror. `DeviceCodeStore.verify` can then go back to taking a single callable. A separate personas module would be more surface than the feature needs.

**Shared macro**: introducing a Jinja macro file is fine even though the tree has none yet - macros are standard Jinja, not a new concept. A small `templates/_macros.html` with the persona picker, imported where needed, is exactly right.

**Advanced Security alerts**: your read is correct - both are pre-existing patterns that your refactor moved, so CodeQL attributes them to the PR. The username/client_id log line is the documented `verbose_logging` dev convenience (the non-verbose branch exists precisely for deployments that don't want it; CodeQL's "password" classification is a mis-taint from `authenticate(username, password)`), and `redirect(callback_url)` is the OAuth redirect, with exact matching enforced for clients that register `redirect_uris` (#67) and deliberate permissiveness otherwise. I've dismissed both alerts as intended behavior - nothing for you to change in this PR.

gprossliner:
PR is rebase, and all blocking and none blocking changes are implemented. I made isolated commits for each task, I assume you will squash merge.

I noticed that while running the e2e tests, two failed, which should be based on the configuration:

  - **"Redirect URI Exact Match"** - this test requires a
    `registered-client` (with a registered `redirect_uris` entry, #67)
    to exist; the persona-login preset's `settings.yaml` only defines
    `demo-client`, so the check can't pass against this preset
    regardless of the code under test.
  - **"UserInfo Groups & Authorities"** - checks that
    `saml.export_groups`-driven claims match the test user's configured
    groups; this preset's minimal `users.yaml` doesn't set up groups
    for that assertion.

I also deleted the `personal-login-mode.md` planing document as the last commit.

cdelmonte-zg (commented):
Re-verified at fcfbacd, empirically: worktree venv, 981 unit tests, ruff and mypy all clean; CI approved and green; live-server probes; full e2e 48/48 including the new persona test.

(The e2e run was against the default config, where the suite's baseline fixtures live - the two failures you saw against the persona-login preset are expected there and not related to your changes: the preset has no `registered-client` and no group fixtures. The persona e2e test toggling the mode itself on the default config is exactly the right shape.)

Confirmed fixed, all six:

1. mypy clean (`ui.py` Optional password annotation).
2. Device picker: persona buttons are `type=button` with an explicit click handler, so Enter in the code field can no longer implicitly authorize the first listed user. One non-blocking note: the only submit button left in that form is Deny, so Enter in the code field now denies and consumes the code. Fail-closed, which is the right direction; consider suppressing implicit submission in that form entirely so Enter is a no-op.
3. Invalid `login_mode` is rejected before anything is written: a live probe with `banana` left settings.yaml byte-identical, flashed the error, server healthy. Valid values persist, and `password` removes the section (omit-at-default).
4. Bare `login:` line: cold start, reload and both writer paths handle the null section; setting and clearing the mode with the null section present works and cleans it up.
5. `user_create` stores the password verbatim, only a stripped-blank means persona-less, and password mode still rejects the blank.
6. Password-grant contract holds live: a password-less user gets 400 on an empty password and 401 on an arbitrary one; a password user authenticates through the grant unchanged while persona mode is on.

The cleanups as well: `interactive_authenticate` on ConfigManager as agreed, the `_macros.html` picker, the shared user schema properties including the previously missing `attributes`, the oauth21 orthogonality test, em dashes gone, plan doc dropped. The SAML `AuthnContextClassRef: unspecified` contract is asserted non-vacuously by the e2e test.

Two side notes from the verification, neither caused by this PR: the settings drift you may notice after an e2e run (`default_acs_url` cleared) reproduces on main and is now #165; the users.yaml re-indentation on the first user write is the pre-existing writer normalization, also reproduced on main.

One item stays open: `examples/persona-login/settings.yaml` ships `host: "0.0.0.0"`. Your observation is factually right - all four existing presets set it explicitly. But they predate GHSA-2473-px8h-rvg6 and are a leftover the advisory work didn't sweep (it changed the default, not the presets); #164 is now open to align them all with the loopback default. That makes the existing pattern one being phased out rather than a convention to extend, and this preset is the worst pairing for it: passwordless login bound to all interfaces puts a session as any user one LAN click away. Please ship the preset as `host: "127.0.0.1"`, optionally with a commented `# host: "0.0.0.0"` opt-in line as in #164. After that one line this is ready to approve.

gprossliner:
I did fix the two things you mentioned. Will you merge the conflict in `mcp_server.py` or should I rebase again?

cdelmonte-zg:
Both fixes verified. I resolved the mcp_server.py conflict myself with a merge commit on your branch (it was purely additive on both sides, the shared user-properties block next to main's new hex-color helpers): 1018 unit tests, ruff and mypy clean, and the full e2e is 49/49 live with the branding and persona features composed. Merging once CI is green.

cdelmonte-zg:
Merged. Thanks for the feature and for the clean, isolated fix rounds - welcome aboard as a contributor.

gprossliner:
It was a rather large PR in the end. Thank you for merging and the very fast and solid responses. I had the experience that some maintainers doesn't like large PRs at all. I'm happy to contribute also in the future. There is nothing that comes to my mind now, but you'll never know. You can also tag me if you think I can help you with something in the future. 

Now as we are in main, when are you planing the 2.7.0 release?

cdelmonte-zg:
Thanks - and the PR size worked out fine in this case: what made it reviewable wasn't the line count but the isolated commits per finding and the fact that every fix round held up under verification. That's the pattern to keep.

On 2.7.0: the features (persona login, per-client branding, the UI login gate from #166) are all on main already; what's left is small housekeeping (#149, #164, #165). Rather than making you wait for the final, there is a pre-release out now:

```bash
pip install --pre nanoidp==2.7.0rc2
# or
docker pull ghcr.io/cdelmonte-zg/nanoidp:v2.7.0-rc2
```

It would be great if you could run the persona flow in your real setup and report anything odd. Plain `pip install nanoidp` and the Docker `latest` tag keep serving 2.6.0. Final 2.7.0 realistically in the first half of September, with any rc feedback folded in.

And noted - I'll tag you if something comes up that fits. Thanks for the contributions.

cdelmonte-zg:
Small update: the current release candidate is now **2.7.0rc3** - rc2 stays valid, but main gained a few fixes right after its tag (`/api/config` exposing `saml.default_acs_url` so an e2e run no longer clears it, the example presets moving to loopback, a path-containment barrier in the logo resolver, and a wheel smoke test in the publish pipeline). If you haven't started testing yet, or when you next upgrade, grab rc3:

```bash
pip install --pre nanoidp==2.7.0rc3
# or
docker pull ghcr.io/cdelmonte-zg/nanoidp:v2.7.0-rc3
```

Nothing changes in what's worth testing - the persona flow, branding and the login gate are identical across the two rcs.

Issue #162, opened 2026-08-21
Web UI login is not enforced on any route


cdelmonte-zg:
Now that GHSA-2473-px8h-rvg6 is public, this loose end can be tracked openly (it was held back to avoid pre-empting the coordinated disclosure).

The web UI ships `/login` + `/logout` and sets `session["user"]`, but no route enforces authentication (there is no `login_required` / `before_request` gate). `session.get("user")` is only used for navbar display and for SAML SSO end-user auth, so the whole config UI is fully usable without ever logging in.

That makes the login page misleading: it looks like a security boundary and is not one - a violation of VISION principle 2 ("metadata never lies"). The unauthenticated management surface itself is a documented dev-tool property (see the advisory and docs/SECURITY.md); the defect here is the UI *pretending* otherwise.

Two ways to resolve it, pick one:

1. **Wire the gate behind an opt-in setting** (off by default, per principle 3), so logging in actually protects the config UI when enabled; or
2. **Remove/rework the misleading login page**, keeping session auth only where it is real (SAML SSO end-user flow), so the UI stops implying a boundary that does not exist.

Option 1 overlaps with the broader centralized management-auth idea (see the companion issue); if that lands, the UI gate should be one consumer of the shared guard rather than a bespoke check.

dshvedchenko:
proposed

# Wire an opt-in login gate for the config UI

## Context

The web UI ships `/login` + `/logout` ([ui.py:54-117](src/nanoidp/routes/ui.py#L54-L117)) and sets `session["user"]`, but no route enforces it. Every other `ui_bp` route reads `session.get("user")` only to feed `current_user=` into templates for navbar display — never as a gate. This makes the login page misleading: it looks like a security boundary and isn't one.

The unauthenticated management surface is an intentional, documented dev-tool property (`docs/SECURITY.md`, GHSA-2473-px8h-rvg6 for the related `0.0.0.0`→`127.0.0.1` bind fix) — the defect being fixed here is narrower: the UI implies protection that doesn't exist.

[...]

Issue #165, opened 2026-08-22
e2e run silently clears saml.default_acs_url: /api/config omits it, the c14n test round-trips it as blank


cdelmonte-zg:
Running `examples/test_agent.py` against a fresh default config leaves `settings.yaml` modified: `saml.default_acs_url` goes from the configured value to `''` (and the key is then dropped as cleared).

Root cause chain, verified on current main (effec52):

1. The "SAML Exclusive C14N" e2e test rebuilds the settings form from `/api/config` and posts `default_acs_url: saml_config.get("default_acs_url", "")` (examples/test_agent.py, both the change and the restore POST).
2. `/api/config`'s `saml` section does not expose `default_acs_url` (routes/api.py), so the `""` fallback always wins.
3. `POST /settings` follows the "present-but-blank = clear" contract (#131), so the field is wiped on the first round-trip and the restore POST wipes it again.

Repro: fresh config dir, run `examples/test_agent.py`, diff `settings.yaml`.

Fix: expose `default_acs_url` in `/api/config`'s `saml` section. Same seam as the #130 additions (`issuer_allowlist`, `device_verification_base_url`, ...): the e2e agent can only round-trip what `/api/config` exposes, and CI never catches it because the default CI config path does not fail on the wipe.

Found while re-verifying #158; the drift reproduces on main and is unrelated to that PR.

Issue #164, opened 2026-08-22
Example presets still bind to 0.0.0.0 explicitly; align them with the 2.6.0 loopback default


cdelmonte-zg:
All four example presets ship an explicit `host: "0.0.0.0"`:

- `examples/cli-device-flow/settings.yaml`
- `examples/microservices-client-credentials/settings.yaml`
- `examples/react-spa-pkce/settings.yaml`
- `examples/spring-boot-saml/settings.yaml`

The 2.6.0 security fix (GHSA-2473-px8h-rvg6, #159) changed the *default* bind address to 127.0.0.1, but presets that set `host` explicitly override it back to all interfaces. Anyone copying a preset gets the pre-advisory exposure back without making a deliberate choice.

Proposal, matching the reverse-proxy guide's framing (#161): presets ship loopback with a commented opt-in line, e.g.

```yaml
server:
  host: "127.0.0.1"
  # host: "0.0.0.0"  # opt-in: reachable from the whole network; use only on a trusted LAN
```

The LAN / multi-device use case some presets target (device flow on a phone, SPA on another machine) stays one uncomment away, but becomes the deliberate opt-in the advisory intends.

Context: surfaced during the review of #158, whose new persona-login preset followed the existing presets' pattern.

PR #166, opened 2026-08-22
Opt-in login gate for the config web UI

dshvedchenko:
Opt-in removal of the invalid-bcrypt-hash plaintext fallback

### Security
- **Opt-in login gate for the config web UI**: new `session.require_ui_login`
  setting (off by default) makes `/login` actually enforce a logged-in
  session on the dashboard, users, clients, settings, keys, claims, audit log
  and token tester pages - previously `/login`/`/logout` existed but nothing
  gated on them, so the login page implied protection it didn't provide.
  Does not affect the separate `/api/*` management API, which remains
  unauthenticated by design regardless. YAML-only for now, following the
  `secret_key`/`security_profile` precedent. Related to the network-binding
  hardening in GHSA-2473-px8h-rvg6.
- **Opt-in removal of the invalid-bcrypt-hash plaintext fallback**: new
  `session.enforce_password_check` setting (off by default). When
  `password_hashing` is on, a `users.yaml` password that isn't a valid
  bcrypt hash previously fell back to plaintext comparison with only a
  warning logged - this setting removes that fallback, rejecting the login
  outright instead. Default behavior (the fallback) is unchanged; opt-in
  only. YAML-only, same treatment as `require_ui_login`.

#162

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Security
+- **Opt-in login gate for the config web UI**: new `session.require_ui_login`
+  setting (off by default) makes `/login` actually enforce a logged-in
+  session on the dashboard, users, clients, settings, keys, claims, audit log
+  and token tester pages - previously `/login`/`/logout` existed but nothing
+  gated on them, so the login page implied protection it didn't provide.
+  Does not affect the separate `/api/*` management API, which remains
+  unauthenticated by design regardless. YAML-only for now, following the
+  `secret_key`/`security_profile` precedent. Related to the network-binding
+  hardening in GHSA-2473-px8h-rvg6.
+- **Opt-in removal of the invalid-bcrypt-hash plaintext fallback**: new
+  `session.enforce_password_check` setting (off by default). When
+  `password_hashing` is on, a `users.yaml` password that isn't a valid
+  bcrypt hash previously fell back to plaintext comparison with only a
+  warning logged - this setting removes that fallback, rejecting the login
+  outright instead. Default behavior (the fallback) is unchanged; opt-in
+  only. YAML-only, same treatment as `require_ui_login`.
+
 ## [2.6.0] - 2026-08-21
 
 ### Documentation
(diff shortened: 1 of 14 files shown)


cdelmonte-zg (approved):
Verified empirically at 7a9d286 (worktree venv, 1028 unit tests, ruff and mypy clean, live server probes):

- **Gate contract holds live**: with `require_ui_login: true`, every ui_bp page redirects to `/login` (GET and POST both - an unauthenticated `POST /users/create` is turned away with nothing written to disk), `/login` itself stays reachable with no redirect loop, and a real login grants access. All non-UI surfaces confirmed untouched while gated: `/api/config`, discovery, `/authorize`, `/token` (client_credentials), `/saml/metadata`.
- **Persona combination behaves as documented**: with `login.mode: persona` on top, a passwordless identity selection satisfies the gate - and the field description plus SECURITY.md say exactly that, including what the gate does not prove in that combination.
- **enforce_password_check**: all four quadrants verified live (flag off keeps the plaintext fallback, flag on rejects an invalid stored hash, a valid bcrypt hash keeps working, wrong password against a valid hash still fails). Defaults unchanged.
- Round-trip regression tests for both YAML-only keys are in, `config/settings.yaml` is net-untouched, the Settings-page addition is read-only status display (no write path), CHANGELOG carries the two features as separate entries.

Both features are accepted in this one PR - two small session-section security opt-ins with the same YAML-only treatment didn't warrant two review rounds.

I pushed one commit on the branch (9edd995) adding the two tests that pin the remaining contract points from the #162 thread: the persona-mode combination satisfying the gate, and the blueprint-separation property (`/api/*` and discovery reachable while gated) - the property the whole design leans on deserved a regression test rather than only a structural argument. 1030 tests green with them. Merging once CI confirms.

cdelmonte-zg:
Merged. Thanks - the honest /api/* caveat in SECURITY.md and the read-only status card are exactly the right way to keep YAML-only settings visible.

cdelmonte-zg:
Heads-up: a 2.7.0 pre-release is out with your recent work in it - the per-client branding (#157) and this login gate with `enforce_password_check` (#166), alongside the persona login mode (#158):

```bash
pip install --pre nanoidp==2.7.0rc2
# or
docker pull ghcr.io/cdelmonte-zg/nanoidp:v2.7.0-rc2
```

If you can exercise the gate and the branding in your real setups, any feedback before the final would be valuable - especially the `require_ui_login` + `login.mode: persona` combination and the YAML round-trip of the new `session:` keys. Plain `pip install nanoidp` and the Docker `latest` tag keep serving 2.6.0. Final 2.7.0 is planned for the first half of September, after #149, #164 and #165.

cdelmonte-zg:
Small update: the current release candidate is now **2.7.0rc3** - rc2 stays valid, but main gained a few fixes right after its tag (`/api/config` exposing `saml.default_acs_url` so an e2e run no longer clears it, the example presets moving to loopback, a path-containment barrier in the logo resolver, and a wheel smoke test in the publish pipeline). If you haven't started testing yet, or when you next upgrade, grab rc3:

```bash
pip install --pre nanoidp==2.7.0rc3
# or
docker pull ghcr.io/cdelmonte-zg/nanoidp:v2.7.0-rc3
```

Nothing changes in what's worth testing - the persona flow, branding and the login gate are identical across the two rcs.

Issue #163, opened 2026-08-21
Centralized opt-in management auth across UI, /api/* and MCP


cdelmonte-zg:
Companion to #162, deferred until GHSA-2473-px8h-rvg6 went public.

nanoidp's management surfaces are unauthenticated by design (dev tool, loopback default since 2.6.0). For the "shared team instance" scenario - one nanoidp deliberately exposed on a trusted network for a whole team - an **opt-in** management auth layer could still be worth having:

- **One shared guard, three consumers**: web UI, `/api/*` routes, and the MCP server. MCP already has its own opt-in (`NANOIDP_MCP_ADMIN_SECRET` + `--readonly`); the guard should absorb/align with that rather than adding a fourth mechanism. Per VISION principle 4 (parity, no drift): if built, centralize - don't bolt on per-surface checks.
- **Off by default** (principle 3): zero-config behavior is unchanged; enabling it is an explicit operator choice.
- Ideally anchored on the existing `ADMIN_ACCESS` entitlement rather than a new concept.
- #162's opt-in UI login gate, if chosen, should be this guard's UI consumer.

Worth building only if the shared-instance scenario proves real; otherwise the loopback default + documented threat model already cover the risk.

dshvedchenko:
Analysis: opt-in management auth layer (companion to #162)
Context
#162 added require_ui_login, an opt-in session gate scoped to ui_bp (the config web UI), in response to a "metadata never lies" concern raised after GHSA-2473-px8h-rvg6: /login existed but enforced nothing. That fix was deliberately scoped to the UI blueprint only, deferring the broader question — a self-registered-OIDC-client idea was explicitly declined and pushed to "a possible future centralized-management-auth issue (#163)."

This request picks that thread back up: nanoidp's management surfaces (ui_bp, api_bp, and the MCP server) are unauthenticated by design, and that's fine for the documented single-developer/loopback default. But for the "shared team instance" case — one nanoidp deliberately reachable by a whole team on a trusted network — a single opt-in guard spanning all three surfaces could be worth adding, rather than leaving /api/* permanently unauthenticated while ui_bp and MCP each grow their own opt-in mechanism independently (VISION principle 4: parity, no drift).

Current state, confirmed by reading the code:

ui_bp — has require_ui_login (src/nanoidp/routes/_auth.py, wired via ui_bp.before_request), session-cookie based, off by default.
[...]

PR #176, opened 2026-08-22
Settings.management_secret (models.py) — YAML-only field, loadable fr…

dshvedchenko:
…om settings.yaml's session: block, NANOIDP_MANAGEMENT_SECRET, or the legacy NANOIDP_MCP_ADMIN_SECRET (kept as alias) — precedence in that order (config.py).

Shared choke point in routes/_auth.py: get_management_secret(), verify_management_secret() (constant-time compare, not bcrypt — a deliberate choice discussed above), management_secret_required_for_api() (X-Management-Secret header, api_bp), management_secret_required_for_ui() (session-based, ui_bp). ui.py gained POST /management/unlock (one-time secret entry, remembered for the session) and a matching form on login.html. mcp_server.py's _check_admin_secret now sources from the same setting instead of its own os.getenv call. app.py's network-bind warning and docs/SECURITY.md updated to reflect the new gate and its relationship to require_ui_login.

One incidental bug found and fixed: get_yaml_writer()'s module-level singleton was never reset between tests (unlike the other four singletons conftest.py resets), so once any test touched the real config dir, a later test's isolated tmp_path app could silently write through the stale cached writer into the repo's real config/*.yaml. Fixed in tests/conftest.py's reset_singletons fixture — this was a pre-existing gap, not something my feature introduced, but my new tests were the first to trigger it.

Verification: full suite is 1048 passed (was 1034), run three times consecutively with git status clean each time. Added a CHANGELOG.md entry under [Unreleased] → Security.

#163

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -211,6 +211,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   contract (#131) silently wiped `default_acs_url` from `settings.yaml`.
 
 ### Security
+- **Opt-in `management_secret` mutation gate** (#163): one shared secret that
+  gates state-changing calls across all three management surfaces - the MCP
+  server (via the existing `admin_secret` tool argument, which now reads from
+  this setting instead of a standalone env var), `/api/*` (via a new
+  `X-Management-Secret` request header on mutating calls), and the config web
+  UI (a one-time "unlock" form at `/login` that then trusts the session for
+  further mutating requests). Off by default - unset, nothing changes.
+  Configurable via `settings.yaml`'s `session.management_secret` or the
+  `NANOIDP_MANAGEMENT_SECRET` env var; the previous MCP-only
+  `NANOIDP_MCP_ADMIN_SECRET` still works as an alias, though an explicit
+  `management_secret: null`/`""` in `settings.yaml` now wins over either env
+  var rather than falling through to it. Independent of `require_ui_login`:
+  that gate is the UI's session front door (who can view the dashboard),
+  this is the write guard (who can change anything) - either, both, or
+  neither can be enabled; the unlock form stays reachable even when
+  `require_ui_login` is also on. YAML-only, same treatment as
+  `require_ui_login`/`secret_key`. Hardened since first landing: the UI
+  unlock flag is now an HMAC of the secret itself (not a bare session
+  boolean), so it can't be forged just by knowing `secret_key`'s public
+  default; a non-ASCII or non-string secret compares safely instead of
+  500ing; an unlocked UI session now also satisfies the `/api/*` gate, so
+  the dashboard's own buttons (generate token, clear audit log) keep working
+  after one unlock; and the MCP check now always reads the `ConfigManager`
+  actually serving the request.
 - **Opt-in login gate for the config web UI**: new `session.require_ui_login`
   setting (off by default) makes `/login` actually enforce a logged-in
   session on the dashboard, users, clients, settings, keys, claims, audit log
--- a/book/src/reference/configuration.md
+++ b/book/src/reference/configuration.md
@@ -380,11 +380,16 @@ own `/login` and the SAML SSO login page are unbranded.
 The SAML options (`strict_binding`, `sign_responses`, `c14n_algorithm`)
 are covered in detail in [SAML options](saml.md). Security-related
 settings are covered in the [Security guide](../guides/SECURITY.md):
-profiles, `require_pkce`, key management, `jwt.external_keys`, and the
+profiles, `require_pkce`, key management, `jwt.external_keys`, the
 config UI's opt-in [login gate](../guides/SECURITY.md#config-ui-login-gate)
-(`session.require_ui_login`), and the invalid-bcrypt-hash
+(`session.require_ui_login`), the invalid-bcrypt-hash
 [fallback removal](../guides/SECURITY.md#invalid-bcrypt-hash-fallback)
-(`session.enforce_password_check`).
+(`session.enforce_password_check`), and the
+[management secret](../guides/SECURITY.md#management-secret) write guard
+shared by MCP, `/api/*`, and the web UI (`session.management_secret`) - all
+YAML-only, like `session.secret_key` (see
+[Session Cookie Trust](../guides/SECURITY.md#session-cookie-trust-secret_key)
+for why that one matters here too).
 
 ## Logging
 
@@ -490,6 +495,7 @@ directory.
 
 ## Environment variables
 
-The environment variables (`NANOIDP_CONFIG_DIR`, `NANOIDP_MCP_ADMIN_SECRET`,
-`NANOIDP_MCP_READONLY`, `PORT`) are listed in the
+The environment variables (`NANOIDP_CONFIG_DIR`, `NANOIDP_MANAGEMENT_SECRET`,
+the legacy `NANOIDP_MCP_ADMIN_SECRET` alias, `NANOIDP_MCP_READONLY`, `PORT`)
+are listed in the
 [Security guide](../guides/SECURITY.md#environment-variables).
--- a/docs/MCP_WORKFLOW.md
+++ b/docs/MCP_WORKFLOW.md
@@ -210,7 +210,13 @@ Prompt: "I need to test role-based access. Create a user 'role-test' with
 
 ### Admin Secret Protection
 
-When `NANOIDP_MCP_ADMIN_SECRET` is set, mutating tools require the secret:
+When [`management_secret`](SECURITY.md#management-secret) is configured -
+via `settings.yaml`'s `session.management_secret`, the
+`NANOIDP_MANAGEMENT_SECRET` env var, or the legacy `NANOIDP_MCP_ADMIN_SECRET`
+env var (this setting's MCP-only predecessor, still supported as an alias) -
+mutating tools require the secret as the `admin_secret` tool argument. The
+same setting also gates `/api/*` and the web UI's mutating form actions -
+see [Management Secret](SECURITY.md#management-secret) for the full picture.
 
 ```json
 {
@@ -219,7 +225,7 @@ When `NANOIDP_MCP_ADMIN_SECRET` is set, mutating tools require the secret:
       "command": "nanoidp-mcp",
       "env": {
         "NANOIDP_CONFIG_DIR": "./config",
-        "NANOIDP_MCP_ADMIN_SECRET": "your-secret-here"
+        "NANOIDP_MANAGEMENT_SECRET": "your-secret-here"
       }
     }
   }
@@ -267,6 +273,6 @@ NANOIDP_CONFIG_DIR=./config nanoidp-mcp
 ### Permission Denied for Mutating Tools
 
 Either:
-1. Provide `admin_secret` in tool arguments (if `NANOIDP_MCP_ADMIN_SECRET` is set)
-2. Or remove `NANOIDP_MCP_ADMIN_SECRET` env var for development
+1. Provide `admin_secret` in tool arguments (if `management_secret` is configured)
+2. Or unset `management_secret` (settings.yaml, `NANOIDP_MANAGEMENT_SECRET`, or the legacy `NANOIDP_MCP_ADMIN_SECRET`) for development
 3. Or check you're not running in `--readonly` mode
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -14,7 +14,7 @@ By design, NanoIDP prioritizes developer convenience over security hardening. It
 
 NanoIDP binds to `127.0.0.1` (loopback only) by default, so out of the box it is reachable only from the local machine.
 
-This matters because the management API (`/api/*`) is **unauthenticated by design** (see [MCP Server Security](#mcp-server-security) for the equivalent concern on the MCP side). Those endpoints can mint a validly signed access token for any user, including `admin`, rotate the signing keys, and clear the audit log. Loopback binding keeps that surface off the network.
+This matters because the management API (`/api/*`) is **unauthenticated by design** (see [MCP Server Security](#mcp-server-security) for the equivalent concern on the MCP side). Those endpoints can mint a validly signed access token for any user, including `admin`, rotate the signing keys, and clear the audit log. Loopback binding keeps that surface off the network. If you need it reachable by more than one host, also consider [`management_secret`](#management-secret), which requires a shared secret for these mutating calls specifically (reads stay open either way).
 
 ### Exposing on a network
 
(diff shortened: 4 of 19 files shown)


cdelmonte-zg (changes requested):
Thanks, this is the agreed design and it is close: one secret, one choke point in `routes/_auth.py`, three consumers, off by default, YAML-only, docs and e2e touched in the same PR. CI green, 1048 passing locally, working tree clean after three runs (the `get_yaml_writer()` singleton fix in conftest is real and welcome).

Everything below was verified empirically on `cbcd101` (manual pass plus an automated review, every finding reproduced before being listed). Requesting changes for the blocking items; the rest is at your discretion.

## Blocking

**B1. The UI leg is bypassable with the default `secret_key`.** `session["management_verified"]` lives in the Flask cookie session, signed with `secret_key`, which defaults to the public `dev-secret-key-change-in-production`. Repro: mint a cookie with `SecureCookieSessionInterface().get_signing_serializer(app).dumps({"management_verified": True})` using the default key, then `POST /users/create` on a stock config with `management_secret` set: 302, user created, the secret was never presented. `/api/*` and MCP require the real secret per call, the UI does not. Since this gate exists precisely for the shared-instance scenario, it has to hold there. Options, pick one or combine: (a) `create_app()` refuses or loudly warns when `management_secret` is set while `secret_key` is still the default; (b) store something bound to the secret (e.g. an HMAC of `management_secret` keyed by `secret_key`) and re-verify it on each mutating request, so a forged flag is not enough; plus a callout in the new SECURITY.md section that the UI trust model is only as strong as `secret_key` (the same caveat applies to `require_ui_login`, worth stating once for both).

**B2. Non-ASCII secret or candidate is a 500, not a 401.** `secrets.compare_digest` on `str` raises `TypeError` when either side has a non-ASCII character. Repro: `management_secret: "segretò"` plus the correct `X-Management-Secret` header: `POST /api/users/admin/token` is a 500; `POST /management/unlock` 500s too, so the instance cannot be unlocked at all; with an ASCII secret, any client sending a non-ASCII header value triggers an unauthenticated 500 on every mutating `/api/*` route; on MCP, `admin_secret: 123` (int) raises `TypeError` and surfaces as a generic internal error instead of "Invalid admin_secret". Fix: `isinstance(candidate, str)` guard and compare `.encode("utf-8")` bytes on both sides.

**B3. With `require_ui_login` and `management_secret` both on, the unlock form cannot work.** `ui_login_required` is registered first and exempts only `ui.login`, so an unauthenticated `POST /management/unlock` is redirected to `/login` before the view runs: session stays empty, no error shown, the form is rendered again as if the secret were wrong. Repro: both settings on, anonymous `GET /login` shows the unlock form, `POST /management/unlock` with the correct secret: 302 `/login`, `session == {}`. Either exempt `ui.management_unlock` in `ui_login_required` too, or render the unlock form only when `is_ui_authenticated()` or the login gate is off. No test covers the both-on combination; please add one.

**B4. The dashboard's own JS buttons are now broken when the secret is configured.** `users.html`, `test.html` ("Generate token") and `audit.html` ("Clear audit log") `fetch()` `POST /api/*` with no header, and `management_secret_required_for_api` ignores `session["management_verified"]`. Repro: unlock the UI session, then `POST /api/audit/clear` with the same cookie jar: 401 (`audit.html` ignores the status and reloads, so it fails silently). Either accept an unlocked session in the API hook for same-origin browser requests, or have the templates send the header (the former is simpler and keeps the header as the contract for non-browser clients).

**B5. MCP reads the gate from the wrong singleton.** `_check_admin_secret` now goes through `routes._auth.get_management_secret()` which uses `nanoidp.config.get_config()` (the config-module global), not `mcp_server._config` (the MCP server's own singleton). Repro: `mcp._config = ConfigManager(dir_with_secret)`, `config._config = None`, cwd at the repo root: `_check_admin_secret("create_user", {})` returns `(True, "")` because `get_config()` lazily built a secret-less `ConfigManager` from `./config`. The production path (`_ensure_config()` -> `init_config()`) is fine, but `tests/test_mcp.py:208` sets `mcp._config` directly, so those tests evaluate the gate against whatever `./config` holds. Read `_ensure_config().settings.management_secret` in the MCP path (or pass the `ConfigManager` into the `_auth` helpers).

**B6. `examples/test_agent.py` does not exercise the gate and its UI flows now silently fail.** The header is attached to three `/api` POSTs only; the agent's eight `ui_bp` form mutations (`/settings` x4, `/users/create`, `/clients/create`, `/clients/<id>/delete`, `/users/<u>/delete`) are now redirected to `/login`, `requests` follows the redirect to a 200, the mutation never happens and later assertions fail (persona-mode cleanup at 2790-2791 leaves the server in persona mode). There is also no negative check. Please: set the header once on `self.session.headers`, `POST /management/unlock` on the session when a secret is given, and add one negative assertion per surface (mutating `/api` call without header -> 401, UI mutation before unlock -> redirect to `/login`). House rule for anything touching the management surfaces.

## Non-blocking

- **Env var is now read at config load, not per call.** The old MCP code read `NANOIDP_MCP_ADMIN_SECRET` on every `_check_admin_secret`; now it is snapshotted into `Settings` until `reload()`. Fine as a behaviour, but say so in SECURITY.md, and the `or` chain means an explicit YAML `management_secret: ""`/`null` does not override an inherited env var despite the "YAML explicit value wins" comment; decide which you mean (key present in YAML wins, or non-empty value wins) and test it. `tests/test_mcp_security.py` and `tests/test_mcp.py` patch `os.environ` around the call and pass only because conftest nulls the config singleton; pointing them at a config fixture would make that robust.
- **Docs-site pages are stale**: `docs/MCP_WORKFLOW.md` (lines ~208, 217, 265) still documents `NANOIDP_MCP_ADMIN_SECRET` as the setting; `book/src/reference/configuration.md` lists the YAML-only `session.*` knobs but omits `management_secret`. The site is canonical for reference pages, please update both here.
- `logout()` does `session.clear()`, which also drops `management_verified`, while the docstrings present the unlock as independent of login. Either document that logout re-locks, or preserve the flag.
- `_UI_MANAGEMENT_EXEMPT_ENDPOINTS` contains `ui.logout`, but logout is GET-only and safe methods return before the endpoint check, so the entry and its comment are dead. The two hooks also duplicate the same prologue; a single `_management_gate_applies()` predicate would do.
- A numeric YAML value (`management_secret: 123456`) fails startup with a clear Pydantic error. Acceptable (fail loud), but worth a "quote the value" note in the YAML example.

Happy to re-verify the next round as before.

dshvedchenko:
All six blocking items and the non-blocking items are addressed, with 1067 tests passing (up from 1048) and a live smoke test against a running server confirming the end-to-end flows work. Summary:

Blocking

B1 — session['management_verified'] is now an HMAC of management_secret itself (keyed by secret_key), not a bare boolean, so a forged cookie built from the known default secret_key no longer unlocks anything. Added a startup warning when secret_key is left default while either gate is on, plus a new "Session Cookie Trust" doc section.
B2 — verify_secret() now type-checks the candidate and compares UTF-8 bytes, so non-ASCII or non-string secrets (including MCP's admin_secret: 123) return a clean rejection instead of a 500/TypeError.
B3 — ui_login_required now exempts ui.management_unlock alongside ui.login, so the unlock form works even with require_ui_login also on.
B4 — management_secret_required_for_api now also accepts an already-unlocked ui_bp session, so the dashboard's own fetch() calls (users/test/audit pages) work post-unlock without template changes.
B5 — _check_admin_secret now takes the ConfigManager actually serving the MCP request instead of going through the separate global singleton.
B6 — examples/test_agent.py sets the header once on self.session, unlocks the UI session up front, routes all UI mutations through that session, and adds negative checks for both surfaces.
Non-blocking — explicit YAML management_secret: null now beats an inherited env var; stale NANOIDP_MCP_ADMIN_SECRET-only docs updated in MCP_WORKFLOW.md and book/src/reference/configuration.md; dead ui.logout exemption entry removed. Left logout()'s re-locking behavior and the two-hooks duplication as-is (discretionary, low value to touch).

Also amended the CHANGELOG's still-unreleased management_secret entry to mention the hardening, since the feature hasn't shipped yet.

cdelmonte-zg (changes requested):
Re-verified round 2 (`0eaf84a`) live against a running instance with the default `secret_key`: B1 forgery attempts (bare `True`, HMAC with a guessed secret, forged `user`) are all rejected, B2 no longer 500s on any surface, B3 anonymous unlock works with both gates on, B4 an unlocked session satisfies `/api/*` mutations, B5 the MCP gate evaluates against its own ConfigManager with no config in cwd, B6 test_agent is all green with the secret and fails explicitly without it. 1067 unit tests pass. Thanks, this is a solid round.

Two small things before merge:

1. **Non-ASCII secrets don't work through the header.** Werkzeug decodes request headers as latin-1, so a UTF-8 `X-Management-Secret: topsecret-ü` never matches (it only matches when the client sends latin-1 bytes, which no normal client does). `test_non_ascii_configured_secret_can_still_unlock` only exercises the form path. Simplest contract: a validator on `Settings.management_secret` restricting it to printable ASCII (fail at startup with a clear message), plus a test. That keeps header, form and MCP JSON on the same footing.

2. **`SameSite` on the session cookie (non-blocking but cheap).** B4 makes the unlocked session cookie authorize `/api/*` mutations (`keys/rotate`, `audit/clear`), so the cross-site form-POST surface that already existed for ui_bp forms now covers the API too; browsers that don't default to Lax (Firefox) send the cookie. `app.config["SESSION_COOKIE_SAMESITE"] = "Lax"` in `create_app()` closes it with one line; a sentence in the new "Session Cookie Trust" section would round it off.

Everything else looks good to go.

dshvedchenko:
Will address tomorrow, viel spass

cdelmonte-zg:
> Will address tomorrow, viel spass

Danke!

cdelmonte-zg:
Heads-up for your next round, unrelated to the review items above: #197 (`feat/175-document-models`, under review) moves the settings/users loaders to document models (`src/nanoidp/config_documents.py`, one Pydantic model per YAML section, `extra="forbid"`, unknown keys warned with their path). Once it merges, this PR will need two small adjustments at rebase time:

1. `session.management_secret` must be declared on `SessionSection`, otherwise every file that sets it will log `unknown key session.management_secret (ignored)` and the value will be dropped.
2. Your absent-vs-null rule (absent -> env fallback, explicit `null` -> off) must survive the document model: a plain `management_secret: Optional[str] = None` on the section would make both cases identical. Either check `"management_secret" in section.model_fields_set` in `SettingsDocument.to_settings()`, or keep the precedence decision at the raw-document edge in `config.py` before the document is built (the same place `check_config_version` and `${VAR}` expansion run). Your existing tests for the null/absent distinction should catch a wrong rebase.

No action needed until #197 lands; I will ping here when it does.

cdelmonte-zg:
#197 is merged (document models in `src/nanoidp/config_documents.py`). When you rebase, please add `management_secret` to `SessionSection` keeping your absent-vs-null rule: absent -> env fallback, explicit `null` -> off. `"management_secret" in self.session.model_fields_set` in `SettingsDocument.to_settings()` is the simplest way; the field type should be `Optional[str]` so `null` is accepted by the document and the decision stays yours. Your existing null/absent tests should confirm it.

cdelmonte-zg:
Update on scope and timing, before your next round: 2.7.0 is now feature-frozen (rc4 is out: hooks/plugins v1, the config document models, the profile/redirect/SAML fixes). This PR introduces a new security boundary across three surfaces and its base predates nine merged PRs, so rather than rushing it into the freeze we have retargeted #163 to **2.8.0**: the review continues exactly as before, and once 2.7.0 final is tagged you can rebase once onto a stable base instead of chasing a moving one.

Two things that will matter for that rebase, both already noted above: `session.management_secret` must be declared on `SessionSection` in `config_documents.py`, and your absent-vs-null semantics need `model_fields_set` (or the decision at the raw-document edge) because the document model would otherwise flatten them.

Meanwhile, if you want to help 2.7.0 ship: **v2.7.0-rc4 is on PyPI (`pip install --pre nanoidp==2.7.0rc4`) and GHCR (`ghcr.io/cdelmonte-zg/nanoidp:v2.7.0-rc4`)**. Your stack (compose, proxy, MCP, shared instance) is exactly the environment our CI cannot replicate: issue reports from running the rc there, especially config migration from 2.5/2.6 files and the new `hooks:`/`plugins:`/`bootstrap.yaml` surface, would be the most valuable soak feedback we could get.

cdelmonte-zg:
Soak update: **v2.7.0-rc5 supersedes rc4** and is the intended-final candidate: `pip install --pre nanoidp==2.7.0rc5` / `ghcr.io/cdelmonte-zg/nanoidp:v2.7.0-rc5`. New since rc4, all relevant to your setup: the published config schema (`docs/schema/config.v1.json`), `nanoidp validate-config` (lint a config directory in CI/pre-commit without starting anything, hooks never execute), the MCP `validate_config` tool, `config_validation: warn|strict` + `--strict-config`, and fully transactional loads (a failed reload commits nothing). If you run one rc in your stack, run this one.

dshvedchenko:
I'm working on fixing final bugs

cdelmonte-zg (changes requested):
Round 3 verified live at head 03fe4c3 (full unit suite: 1421 passed locally; every item from rounds 1-2 re-checked at this head, plus the two round-2 items):

**Round-2 items - both confirmed fixed:**
- **ASCII validator**: startup with a non-ASCII `NANOIDP_MANAGEMENT_SECRET` fails loudly with the ValidationError, `validate-config` reports it, and an ASCII secret now authenticates via the `X-Management-Secret` header end to end (the round-2 gap).
- **SameSite**: the session `Set-Cookie` now carries `SameSite=Lax` (verified on the unlock response).

**Rebase integration - verified:**
- `session.management_secret` is declared on `SessionSection` and the precedence lives in `SettingsDocument.to_settings()` with `model_fields_set`, exactly as requested. All five cases verified live: absent key falls back to `NANOIDP_MANAGEMENT_SECRET`, explicit `null` beats the env var, explicit value beats the env var, legacy `NANOIDP_MCP_ADMIN_SECRET` still works as last resort, and the new var beats the legacy one.
- `docs/schema/config.v1.json` gained the field; `validate-config --strict` is clean on a config with the secret set.
- Nothing was dropped by the rebase: round 2's `config.py` logic moved into the document models, everything else carried over (checked the full-branch diffstat against round 2's, per file).
- Re-checked rounds 1-2 live at this head: forged bare-`True` / wrong-guess-HMAC / forged-user cookies all get 401 with the default `secret_key`; non-ASCII header candidate is 401 not 500; anonymous unlock works with both gates on; an unlocked session satisfies `/api/*` mutations; GET is never gated; `/api/config` does not leak the secret; the default-`secret_key` startup warning fires.

**One new blocking item (same family as B6, introduced by the rebase, not by you):**

`examples/test_agent.py::test_native_app_redirect_uris` creates and deletes its test client with bare `requests.post` (the `/clients/create` call around line 861 and the `finally` delete around line 967) instead of `self.session`. With a management secret configured, the create POST is redirected to `/login` - a 302, which the `in (302, 303)` success check happily accepts - so the client is never created and the whole accept/reject matrix fails; the cleanup delete is a silent no-op too.

Repro: fresh config, server with `NANOIDP_MANAGEMENT_SECRET` set, `test_agent.py --management-secret ...` -> "Native App Redirect URIs" FAILS; without a secret it passes. CI's http-e2e job doesn't catch it because that job runs without a secret.

This test landed on main after your round 2 (the #178-#184 batch), so it just needs the same `self.session` treatment you already gave `test_client_branding`. Please also grep for any other bare `requests.post` against `/clients/*` or `/api/*` mutations while you're there - the remaining ones I checked (authorize/token/device/saml/login) are protocol endpoints and correctly stay bare.

(Not yours, for the record: "Redirect URI Exact Match" fails on a fresh `init` config even on main - it needs the pre-provisioned `registered-client`. Unrelated to this PR.)

**Housekeeping note, non-blocking:** the squash reused the commit message and authorship of my #206 docs commit (the branch commit shows me as author with the #206 body on top of your summaries). The eventual squash-merge will attribute the commit to you as PR author regardless, but if you push again anyway, an amended author/message would keep the branch history readable.

Merge plan unchanged: this stays on 2.8.0 (issue #163's milestone), with one rebase onto the stable contract after 2.7.0 final.

cdelmonte-zg (approved):
Round 4 verified live at head 05fe9a3 - the fix is exactly right and everything holds.

- The two `self.session` switches in `test_native_app_redirect_uris` (create + finally-delete) fix the failure: fresh config, server with `NANOIDP_MANAGEMENT_SECRET` set, `test_agent.py --management-secret ...` -> "Native App Redirect URIs" now PASSES, 55/56 total with only the pre-existing environmental "Redirect URI Exact Match" (needs the pre-provisioned `registered-client`, fails on main with a fresh `init` config too - not this PR's concern).
- Negative run (agent without the secret against the same server): the seven mutation-dependent tests fail explicitly, none silently pass - the gate is enforced end to end.
- No stale `native-test-*` client left behind: the cleanup delete now rides the unlocked session.
- Full unit suite: 1421 passed locally; CI green on all jobs.

Approving. Merge timing per the plan already discussed: this lands with 2.8.0, after 2.7.0 final is tagged - expect one last trivial rebase then (the 2.7.0 CHANGELOG consolidation will touch your `[Unreleased]` entry's surroundings; nothing in the code contract should move). Thanks for the quick turnaround, and for carrying the feature cleanly through the document-models migration.

PR #178, opened 2026-08-23
fix: --profile overrides YAML for every value and survives reloads (#172)

### LINKED ISSUE #172: CLI --profile cannot override a stricter YAML security_profile, and the override is lost on the first config save
`create_app()`'s comment says the CLI `--profile` wins over `settings.yaml`'s `security_profile`, but the override only fires for `stricter-dev` and `oauth21` (app.py:37). An explicit `--profile dev` therefore cannot bring a file configured with `oauth21`/`stricter-dev` back to `dev`, even though the CLI advertises `dev` as a valid choice and always passes it to `run_app()`.

Verified empirically on current main: YAML `security_profile: oauth21` + `create_app(profile="dev")` -> effective profile stays `oauth21`.

Root cause: `--profile` declares `default="dev"` (__main__.py), so the code cannot distinguish "user explicitly asked for dev" from "flag omitted" - overriding unconditionally would make the default clobber the YAML value on every start, which is why the current code special-cases the two non-default values.

Fix direction:
- `--profile` gets `default=None`; help text says "overrides settings.yaml when given".
- `create_app()` applies `if profile is not None: settings.security_profile = profile` (validation already exists on the model).
- Comment updated to match reality; regression tests for all three explicit values over a conflicting YAML, plus omitted-flag leaving YAML untouched.

Not a security issue (the current behavior fails toward the stricter profile), but the flag contradicts its own documentation. Surfaced during the v2.7.0-rc2 review; previously noted in the 2.6.0 preflight as "real but non-security, defer 2.6.1/2.7".

cdelmonte-zg:
Closes #172.

## What was wrong

- `--profile` defaulted to `dev`, so the code could not tell "asked for dev" from "omitted" and only honoured `stricter-dev`/`oauth21`; an explicit `--profile dev` could not bring a YAML `oauth21` back to `dev`.
- The override only mutated the in-memory `Settings` once in `create_app()`. Every `YamlWriter` mutation ends with `get_config().reload()`, which rebuilds `Settings` from YAML, so the first UI/MCP save silently dropped the CLI profile (dshvedchenko's repro in the issue).
- Found while fixing: the `stricter-dev` runtime hardening (`require_pkce`, `password_hashing`, `rate_limit_enabled`, debug off) was applied in the same one-shot block and lost on the same first reload, even when `security_profile: stricter-dev` came from `settings.yaml` itself. Repro on main: YAML stricter-dev, `create_app()`, `reload()` -> `require_pkce=False`, `password_hashing=False`, `rate_limit_enabled=False` while `security_profile` still reads `stricter-dev`.

## What changed

- `ConfigManager(config_dir, profile_override=None)` / `init_config(..., profile_override=)` own the override; `_load_settings()` is now read + `_apply_profile()`, so the effective profile and its hardening are re-derived after every load. `create_app()` just passes the flag through.
- `--profile` defaults to `None`; choices come from a single `SECURITY_PROFILES` tuple in `models.py` shared with the validator.
- `_save_settings()` preserves the on-disk `security_profile` when an override is active: a transient flag never rewrites the operator's file.
- `GET /api/config` exposes `security_profile`, `profile_override` and the derived `effective` values (the document did not report the profile at all, which is why this could not be observed from outside).
- `examples/test_agent.py`: new `API Profile Survives Reload` check (reads the document, reloads, asserts it is identical and that stricter-dev implies `require_pkce`).
- Docs: `configuration.md` comment, SECURITY.md paragraph, CHANGELOG.

## Verification

- 27 new unit tests (`tests/test_profile_override.py`): full CLI x YAML matrix, omitted flag, invalid override rejected, reload and `YamlWriter` save keep the override, override never persisted and an explicit YAML profile preserved on save, stricter-dev hardening after reload from YAML and from CLI, `/api/config` exposure. 1061 passed, ruff and mypy clean.
- Live: server with YAML `security_profile: oauth21` and `--profile stricter-dev`, `/api/config` reports `stricter-dev` with `require_pkce: true` before and after `POST /api/config/reload`, YAML still says `oauth21`. `test_agent.py` 50/50 including the new check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   local files only (no remote URLs) to avoid beacons.
 
 ### Fixed
+- **`--profile` overrides settings.yaml for every value and survives reloads**
+  (#172). An explicit `--profile dev` could not bring a file configured with
+  `oauth21`/`stricter-dev` back to `dev` (the flag defaulted to `dev`, so the
+  code could not tell "asked for dev" from "omitted"), and any CLI profile was
+  dropped by the first configuration reload, i.e. by the first web UI or MCP
+  save. Worse, the `stricter-dev` runtime hardening (`require_pkce`,
+  `password_hashing`, `rate_limit_enabled`, debug off) was applied once in
+  `create_app()` and silently lost on that same first reload, even when the
+  profile came from `settings.yaml` itself. The override now lives on
+  `ConfigManager` (`--profile` defaults to none, `init_config(...,
+  profile_override=)`), the effective profile and its hardening are re-derived
+  after every settings load, and a save never writes the override into the
+  operator's file. `GET /api/config` exposes `security_profile`,
+  `profile_override` and the derived `effective` values; the e2e agent checks
+  they are stable across a reload.
 - **Example presets now bind to `127.0.0.1`** (#164). All four pre-2.6.0
   presets (`cli-device-flow`, `microservices-client-credentials`,
   `react-spa-pkce`, `spring-boot-saml`) still shipped an explicit
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,8 +44,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   profile came from `settings.yaml` itself. The override now lives on
   `ConfigManager` (`--profile` defaults to none, `init_config(...,
   profile_override=)`), the effective profile and its hardening are re-derived
-  after every settings load, and a save never writes the override into the
-  operator's file. `GET /api/config` exposes `security_profile`,
+  after every settings load, and a save serializes the DECLARED state
+  (`ConfigManager.persistable_settings()`), so neither the override nor
+  the hardening it implies is ever written into the operator's file. `GET /api/config` exposes `security_profile`,
   `profile_override` and the derived `effective` values; the e2e agent checks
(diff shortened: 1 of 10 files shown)


cdelmonte-zg:
Confirmed and fixed in 4160a44. You were right on both counts: the bug was real (repro on the previous commit: `--profile stricter-dev`, `save()`, restart without the flag -> `require_pkce=True`), and chasing each derived field with a manual restore was the wrong shape.

What changed:
- `ConfigManager.settings` is the effective state. `_apply_profile()` records, for every field it forces, the value the file declared (`self._declared`), with the forced fields listed once in `_STRICTER_DEV_HARDENING` so applying and excluding cannot drift.
- `persistable_settings()` returns `settings.model_copy(update=self._declared)`: the declared state. `_save_settings()` serializes that instead of `self.settings`; the `security_profile` special case is gone.

Tests added (`TestDerivedHardeningIsNeverPersisted`): CLI stricter-dev + save writes neither the profile nor a forced `true` for any hardening field; an explicit `require_pkce: false` in the file survives a save under the flag; a YAML-declared stricter-dev does not spell its effects out on save either; and the end-to-end case: run with the flag, save, restart without it -> `dev`, `require_pkce=False`, `password_hashing=False`, `rate_limit_enabled=False`. Same sequence on the previous commit gave `require_pkce=True`. 1066 passed, ruff/mypy clean.

One note on what the file looks like after a save: the writer already materializes managed keys at their declared value (`require_pkce: false` appears, as it did before this PR); what never appears now is the forced `true`.

PR #179, opened 2026-08-23
chore: enforce import contracts with import-linter (#149)

### LINKED ISSUE #149: Enforce architectural invariants with import-linter (currently comment-only)

cdelmonte-zg:
Closes #149.

Adds import-linter with the two contracts from the issue in `[tool.importlinter]` (pyproject.toml), and a `Check import contracts` step in the Tests workflow after mypy.

- `exclude_type_checking_imports = true` is supported natively by import-linter 2.13 / grimp 3.15, so the `serialization -> config` TYPE_CHECKING edge is excluded without an `ignore_imports` hack. The exclusion is load-bearing: with it off, contract 1 false-positives on that edge.
- Import graph verified against the package before writing the contracts; it matches the issue (`app.py` and `mcp_server.py` sit above routes/services and are outside both contracts).
- Negative tests done locally: a runtime `from .services import ...` in `serialization.py` and a `from .routes import ...` in `config.py` each break their contract (`Contracts: 0 kept, 2 broken`, with the offending lines reported), reverted before commit.
- Clean branch: `Analyzed 28 files, 56 dependencies. Contracts: 2 kept, 0 broken.` Full suite 1034 passed, ruff and mypy clean.
- CONTRIBUTING.md gets a short paragraph on the contracts; CHANGELOG entry under Unreleased.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,10 @@ jobs:
         run: |
           mypy src
 
+      - name: Check import contracts
+        run: |
+          lint-imports
+
       - name: Run tests
         run: |
           pytest tests/ -v --tb=short
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **Import contracts enforced in CI** (#149): `import-linter` now pins the
+  package layering (`routes -> services -> config`) and the invariant that
+  `serialization.py` has no runtime imports from the package (it is what
+  lets `config.py` import it without a cycle). Both used to live only in
+  comments; `lint-imports` runs next to ruff and mypy in the Tests workflow
+  and fails when a change adds a forbidden import.
 - **Persona login mode** (#156): opt-in `login.mode: persona` lists the
   configured users on every interactive login surface (`/login`,
   `/authorize`, `/saml/sso` and the device flow's verification page) and
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,7 +99,17 @@ ruff check src tests
 
 # Fix linting issues automatically
 ruff check --fix src tests
+
+# Check architectural import contracts (#149)
+lint-imports
 ```
+
+`lint-imports` enforces the package layering `routes -> services -> config`
+and keeps `serialization.py` free of runtime imports from the package (that is
+what lets `config.py` import it without a cycle). The contracts live in
+`[tool.importlinter]` in `pyproject.toml`; a type-checking-only import does not
+count. If a change needs a new edge between layers, adjust the contract in the
+same PR and say why.
 
 ## Code Style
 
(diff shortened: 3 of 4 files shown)


cdelmonte-zg:
Fixed in 79d2e1d: `forbidden_modules = ["nanoidp.*"]` (import-linter bumped to `>=2.1` for wildcards), so the contract now matches the docstring's invariant instead of a hand-picked list.

Negative tests re-run locally against the new contract:
- `from .models import Settings` appended to `serialization.py`: `Contracts: 1 kept, 1 broken`, reported as `nanoidp.serialization -> nanoidp.models (l.485)`.
- `from .services import get_audit_log`: `Contracts: 0 kept, 2 broken` (forbidden + layers).
- Clean branch: `Contracts: 2 kept, 0 broken`.

Thanks for the note on `exclude_type_checking_imports` (1.12.1), left as is.

PR #180, opened 2026-08-23
feat: accept native-app redirect URIs per RFC 8252 (#81)

### LINKED ISSUE #81: Native-app redirect URIs: custom schemes and loopback variability (RFC 8252)

cdelmonte-zg:
Closes #81.

Native-app redirect URIs per RFC 8252, both changes additive and cited in code:

- **Syntax** (`is_absolute_redirect_uri`): scheme required, authority optional when a path is present, so `com.example.app:/oauth2redirect` (RFC 8252 section 7.1) is accepted; relative references and fragments rejected (RFC 6749 section 3.1.2). Fragments were previously let through by the `scheme and netloc` check; now rejected, noted in the CHANGELOG.
- **Matching** (`redirect_uri_matches`): exact string comparison everywhere, with the single RFC-mandated exception: when the REGISTERED URI is `http://127.0.0.1[:port]/...` or `http://[::1][:port]/...`, the requested port is ignored (RFC 8252 section 7.3); scheme, host, path and query stay exact. Keyed on the registered side so a web or `localhost` registration never gains port flexibility. `localhost` is deliberately not loopback here (sections 7.3 and 8.3). The OAuth 2.1 draft keeps the carve-out (section 4.1.1), so `oauth21` does not tighten it; a test pins that.
- One shared helper, `services/redirect_uri.py`, used by both legs of `/authorize`. The token endpoint's check compares against the URI used at `/authorize` (RFC 6749 section 4.1.3 "identical"), so exact compare stays correct there and is unchanged.
- Parity: `models.py` field description and the MCP `create_client`/`update_client` schema descriptions updated; `examples/test_agent.py` gets `test_native_app_redirect_uris` (registers a client via the UI form, 6-case matrix incl. negatives, completes the loopback flow, cleans up). Docs: "Native apps (RFC 8252)" paragraph in `configuration.md`.

## Verification

- `tests/test_native_app_redirect_uris.py`: 66 cases (helper matrices, `/authorize` accept/reject matrix with no-redirect assertions on rejects, oauth21 profile, UI/API/MCP registration storing native URIs verbatim). Full suite 1100 passed, ruff and mypy clean.
- Live (client registered with `com.example.app:/oauth2redirect`, `http://127.0.0.1:0/callback`, `http://localhost:3000/callback`, `https://app.example.com/cb`): custom scheme 200, `127.0.0.1:51234` 200, `[::1]:7` 400 (only v4 registered), `localhost:3001` 400, `app.example.com:444` 400, `127.0.0.1:51234/other` 400, `https://127.0.0.1:5` 400, `/relative` and `#frag` 400 "Invalid redirect_uri"; login POST leg 302s to `http://127.0.0.1:51234/callback?code=...` and to `com.example.app:/oauth2redirect?code=...`. `test_agent.py` 50/50 including the new test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   /api/config` and the MCP `get_settings` tool expose the effective value;
   the e2e agent asserts it. Bumps only on renames, removals or semantic
   changes (with a loader migration), never on optional additions.
+- **Native-app redirect URIs** (#81, RFC 8252): `/authorize` accepts
+  private-use scheme redirect URIs such as `com.example.app:/oauth2redirect`
+  (§7.1: a scheme and a path, no authority) as absolute URIs, and a
+  registered loopback URI (`http://127.0.0.1:{port}/...`,
+  `http://[::1]:{port}/...`) matches any port (§7.3), since native apps
+  bind an ephemeral port. Everything else keeps exact string matching (RFC
+  6749 §3.1.2.3): scheme, host, path and query of a loopback URI, every
+  port of a non-loopback or `localhost` registration. Fragments are now
+  rejected explicitly (§3.1.2). One shared matcher,
+  `services/redirect_uri.py`, serves both legs of `/authorize`; MCP tool
+  descriptions and `examples/test_agent.py` updated.
 - **Persona login mode** (#156): opt-in `login.mode: persona` lists the
   configured users on every interactive login surface (`/login`,
   `/authorize`, `/saml/sso` and the device flow's verification page) and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   changes (with a loader migration), never on optional additions.
 - **Native-app redirect URIs** (#81, RFC 8252): `/authorize` accepts
   private-use scheme redirect URIs such as `com.example.app:/oauth2redirect`
-  (§7.1: a scheme and a path, no authority) as absolute URIs, and a
+  (§7.1: a scheme and a path, no authority) as absolute URIs and applies
+  §7.1's minimum rule to them (a non-`http(s)` scheme without a period,
+  such as `myapp://`, is rejected with a message naming the rule; domain
+  ownership is not verified), and a
   registered loopback URI (`http://127.0.0.1:{port}/...`,
   `http://[::1]:{port}/...`) matches any port (§7.3), since native apps
   bind an ephemeral port. Everything else keeps exact string matching (RFC
--- a/book/src/reference/configuration.md
+++ b/book/src/reference/configuration.md
@@ -215,7 +215,12 @@ any absolute URI, the permissive dev default.
 **Native apps (RFC 8252)**: two things a native client needs are built
 in. A private-use scheme URI such as `com.example.app:/oauth2redirect`
 (§7.1: a scheme and a path, no host) is a valid `redirect_uri` and can be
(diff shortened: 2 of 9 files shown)


cdelmonte-zg:
Decision taken: nanoidp is a spec-honest test IdP, so it enforces the section 7.1 minimum rule rather than staying permissive. Fixed in 8046730.

- New single syntactic gate `redirect_uri_rejection_reason()` in `services/redirect_uri.py`, used by `/authorize`: non-absolute or fragment -> generic "Invalid redirect_uri" (RFC 6749 3.1.2); a scheme other than http/https without a period -> `Invalid redirect_uri: private-use scheme must be reverse-domain based, e.g. com.example.app:/callback (RFC 8252 section 7.1)`, so a client developer learns why `myapp://` is not enough. The comment states explicitly that this is the "at a minimum" rule and that domain ownership is not verified. Authorization time only; registration surfaces still store whatever they are given.
- Tests flipped/added (`myapp://callback`, `myapp:/x`, `app:/cb` rejected with the reason; `com.example.app:/..`, `com.example.app://callback`, `a.b:/x`, http/https accepted); e2e matrix gains the negative; `configuration.md` and the CHANGELOG now say "applies section 7.1's minimum rule" instead of claiming full RFC 8252 validation.
- 1111 passed, ruff/mypy clean. Live: `com.example.app:/cb` 200, `myapp://cb` 400 with the message above.

cdelmonte-zg:
Good catch, fixed in 79cf30a. `_without_port` re-serialized the parsed URI; it is replaced by `_strip_loopback_port`, which uses `urlparse` only to locate the authority and to validate the port (`.port` raises on `:evil` and on out-of-range values) and then slices only the literal `:port` out of the ORIGINAL string, so scheme case, an empty query, trailing slashes and everything else must be byte-identical. A zero-padded `:0080` is refused as well (parses as 80 but is not the literal suffix).

Negative tests added (`TestLoopbackPortExceptionTouchesOnlyThePort`): `:evil`, `:99999`, `/cb?`, `HTTP://`, trailing slash, path case, `:0080`, userinfo, the same for `[::1]`, and a registered URI with an invalid port never matching; plus the positive cases (`:123`, `:65535`, no port). 1124 passed, ruff/mypy clean.

cdelmonte-zg:
You are right, RFC 3986 3.2.3 `port = *DIGIT` and RFC 8252 8.4 "except the port component". Fixed in 2d8fc5e (on top of a rebase onto main after #182): `_strip_loopback_port` now removes the port text as written (`parsed.netloc.rsplit(":", 1)[1]`, validated by `parsed.port` just above) instead of requiring it to equal `str(port)`. `:0080` and `:00` moved to the accepted cases, plus `[::1]:0042` for the IPv6 path; all other negatives unchanged. 1187 passed, ruff/mypy clean, `lint-imports` 2 kept. CI running on the new head.

Issue #177, opened 2026-08-22
idea: to have option for connecting hashicorp vault for storing configuration in structured way


dshvedchenko:
it requires vault host, apprrole, and path under which all information is stored.

On bootstrap init process connects to predefined path and deploys all secrets and credentials along with users records in vault,
then works with vault only . 
Optionally can export data back to yaml or dump into both locations

cdelmonte-zg:
Thanks for writing this down. Before deciding anything I want to understand the workflow behind it, because the issue as stated describes two different things and only one of them fits nanoidp.

**What I think I'm reading.** From #119 (Compose stack, "Sign in with OIDC Provider" in Vault) and the sequence #126 -> #163 -> #175 -> #177, your setup looks like a shared dev/test stack (Compose or k8s) for a team, where nanoidp is the IdP for Vault and other services, and Vault is already there. If that's right, the need I'd guess is "I don't want client secrets and user passwords committed in YAML next to the stack". That is a legitimate need for that scenario and it is mostly covered today:

- both `settings.yaml` and `users.yaml` expand `${VAR}` / `${VAR:default}` placeholders on load (see the [reverse proxy guide](https://cdelmonte-zg.github.io/nanoidp/guides/reverse-proxy.html) for the syntax), so a client secret or a password can be `${NANOIDP_CLIENT_SECRET}`;
- Vault Agent / External Secrets / a one-shot init container can render those env vars, or write the whole `users.yaml` into a volume that `NANOIDP_CONFIG_DIR` points at, before nanoidp starts.

In that shape nanoidp never knows Vault exists, the YAML stays the readable, diff-able contract, and the secret-handling policy lives in your deploy where it belongs.

**What does not fit.** "Then works with Vault only" makes Vault the persistence layer. That crosses a non-goal in [VISION.md](https://github.com/cdelmonte-zg/nanoidp/blob/main/VISION.md#non-goals) ("YAML files plus in-memory runtime state are a feature"), adds a hard infrastructure dependency and a client to maintain, and pulls nanoidp toward a hosted product, which it is not. I will not take that part in any form, including as an opt-in backend.

**So, three questions** to find out whether anything is actually missing:

1. What exactly would live in Vault: client secrets, user passwords, the full users/clients list, the whole settings file?
2. Who writes it, and how: a person, a pipeline, or an agent through the UI/MCP? (If the answer is "the UI/MCP should write back to Vault", that is the part that is out of scope.)
[...]

Issue #175, opened 2026-08-22
Config contract: config_version, model-driven loader, generated schema, startup validation


cdelmonte-zg:
Umbrella for the config contract work triggered by #174. Goal: one source of truth for the YAML config format, a declared schema version that external tools can rely on, and a startup check that stops silently ignored keys.

## Why

The config contract is currently restated independently in six places, with nothing verifying they agree:

| Layer | File | Restates |
|---|---|---|
| Model | `models.py` | 46 `Settings` fields + 20 `User`/`OAuthClient` fields, defaults and validators |
| Loader | `config.py` | 56 `.get(key, default)` for settings + 12 for users/clients: every YAML key and default, by hand |
| Writer | `serialization.py` | 37 default literals deciding what to omit at the default |
| MCP | `mcp_server.py` | 14 `update_settings` properties |
| UI | `routes/ui.py` | 20 form fields |
| Docs | `book/src/reference/configuration.md` | every key, by hand |

Consequences: a typo in `settings.yaml` (`oauth.isuer`) is silently replaced by the default because the loader only picks the keys it knows; adding a field touches 5-6 files; no file declares which format version it follows, so an external tool has no contract to target.

## Pieces (in this order)

- [x] **1. `config_version` field** (#182) (small, ships first): top-level `config_version: 1` accepted in `settings.yaml`, `users.yaml`, `clients.yaml`; absent = 1 (no change for existing files); a higher unknown major is rejected at startup with a clear message; written by `nanoidp init`/wizard and preserved by the writer. Single integer, not semver: optional additions do not bump it, renames/removals/semantic changes do, with a migration step in the loader. Documented in `configuration.md` and CHANGELOG gets a "Config schema" section whenever it changes.
- [x] **2. Loader from the models** (#197): replace the hand-written `.get()` mapping in `config.py` with a nested Pydantic model mirroring the YAML sections (`oauth`, `saml`, `session`, `login`, `jwt`, `logging`) with `extra="forbid"`, so unknown keys and wrong types surface as `ValidationError` with the key path; `serialization.py` takes its defaults from the model instead of its own literals. `${VAR}` placeholders and legacy keys must keep loading (see #127/#131 write conventions).
[...]

PR #182, opened 2026-08-23
feat: declare config_version in settings.yaml and users.yaml (#175 piece 1)

cdelmonte-zg:
Piece 1 of #175 (`config_version`), plus the configuration-reference docs agreed in #177's discussion.

## config_version

- Top-level integer `config_version: 1` accepted in `settings.yaml` and `users.yaml`; absent = 1, so existing files load unchanged. (No `clients.yaml` exists in this codebase: clients live under `oauth.clients`; nothing was invented for it.)
- `CONFIG_VERSION = 1` and `check_config_version(data, file_path)` live in `serialization.py` (no package imports, import contract respected); `config.py` runs the check before reading any key of either file and stores the effective value on `ConfigManager.config_version`.
- A value that is not a positive integer (`bool` explicitly excluded, since YAML `true` is an `int` in Python), or one newer than this release, stops startup with a message naming the file, the value and the supported version:
  - `settings.yaml: config_version 2 is newer than this release supports (config_version 1); upgrade nanoidp or downgrade the file`
  - `settings.yaml: config_version must be a positive integer, found 'one' (this release supports config_version 1)`
- `nanoidp init` templates and both wizard templates write `config_version: 1` with a one-line comment. Writers needed no change: `apply_settings_document`/`apply_users_document` are read-modify-write and leave unknown top-level keys alone; tests pin that an existing key survives `save()` and `YamlWriter` and that no key is added to files without one (absent = 1 is the contract, operators' files are not churned).
- Exposure: `GET /api/config` top-level `config_version`; MCP `get_settings` result carries the same key; `examples/test_agent.py` gains `API Config Version`.
- Docs: "Config schema version" section at the top of `configuration.md` (single integer, not semver; optional additions never bump it; renames/removals/semantic changes do, with a loader migration); CHANGELOG gets a "Config schema" subsection under Unreleased.

## Docs: placeholders and the config directory as the interface

New section in `configuration.md`: `${VAR}` / `${VAR:default}` expansion (previously shown only in the reverse-proxy guide), and the contract that lets any external configuration source work without a backend in nanoidp: populate `NANOIDP_CONFIG_DIR` before start, `POST /api/config/reload` or the MCP `reload_config` tool afterwards, nanoidp never reads from or writes to anything but the files.

## Verification

- 20 new tests (`tests/test_config_version.py`); full suite 1054 passed, ruff and mypy clean.
- Live: absent and explicit `1` start and report `config_version: 1` on `/api/config`; `2` in settings.yaml, `"one"`, and `3` in users.yaml each exit 1 with the messages above. `test_agent.py` 51/51 including the new check.


cdelmonte-zg:
All three addressed in 40b40a5.

1. Code, not docs: `_load_users()` now runs `check_config_version(data, users_file)` then `data = _expand_env_vars(data)` (the expander already recurses into roles lists and the attributes dict). Tests: `password: ${ALICE_PASSWORD}` / `email: ${ALICE_EMAIL:alice@example.org}` load expanded and the password grant works; live on a scratch config: with the variable set, alice/s3cret -> 200 and `/api/users` shows the defaulted email; with it unset the server refuses to start (`User.password` min length 1, so an unset secret never becomes an empty password). One thing worth knowing that the docs now state: `YamlWriter.save_user()` rewrites only that user's entry and leaves `${ALICE_PASSWORD}` literal for the others (tested), but `ConfigManager.save()` (MCP `save_config`) rebuilds the whole users map and materializes expanded placeholders. A per-field merge for the users writer, like settings has, would be the follow-up if you want parity; not done here.
2. `config_version` is a literal integer: documented in both sections ("any scalar value except config_version"), `check_config_version` docstring says so, and `config_version: ${CONFIG_VERSION:1}` is tested as rejected in both files with the clear message.
3. Contract rule fixed now, in the docs and the docstring: the version belongs to the configuration directory's contract as a whole; both files declare the same number; each file is checked independently against the supported version; a future bump applies to both files together with one loader migration.

1062 passed, ruff/mypy clean. I will rebase this on main after #178 merges (shared lines in config.py, api.py, test_agent.py and the CHANGELOG).

cdelmonte-zg:
Rebased on main after #178 and #179 (now b4930cd); the only conflicts were in the CHANGELOG (both entries kept). Full suite 1094 passed, ruff/mypy clean, `lint-imports` 2 kept, 0 broken. Ready for the final review.

cdelmonte-zg:
Both halves fixed in 34bed26; agreed that they are one finding.

- `IMPLICIT_CONFIG_VERSION = 1` (frozen) is what an absent key declares; `CONFIG_VERSION` is the newest contract this release understands and moves with bumps. `check_config_version` returns the former for an absent key. The comment on the constants says why there are two. `ConfigManager.config_version` starts from the implicit one as well.
- `_load_users()` now keeps the users.yaml version and refuses a mismatch with settings.yaml (`users.yaml: config_version X does not match settings.yaml's config_version Y; the configuration directory follows one contract version`).
- Tests: the helper test asserts `== 1`, not `== CONFIG_VERSION`; `test_absent_stays_v1_after_a_future_bump` monkeypatches `CONFIG_VERSION = 2` and checks absent -> 1, explicit 1 -> 1, 2 -> 2, 3 refused; `test_files_must_agree_on_the_contract_version` (same simulated v2) checks both mismatch directions refuse and both-equal / both-absent load. Docs updated ("Absent means 1, permanently"; mismatch refuses to start).
- 1096 passed, ruff/mypy clean, `lint-imports` 2 kept.

Issue #131, opened 2026-08-19
POST /settings resets every setting absent from the form (silent config wipe)


cdelmonte-zg:
## Summary

The `POST /settings` handler derives every value from the submitted form: a checkbox absent from the form is stored as `false`, and a missing or blank text field clears the stored value. Any client that submits a partial form therefore silently resets settings it never intended to touch.

## Observed live

During a toggles-on release-readiness run (`issuer_from_request` + `issuer_allowlist` + `issuer_from_proxy_headers` + SAML export with custom attribute names + `device_verification_base_url`, all enabled), the e2e agent's **"SAML Exclusive C14N" test** does a `/settings` round-trip sending only the fields it knows about (issuer, audience, expiry, SAML entity/SSO/ACS, sign/strict/c14n, identity classes). Mid-suite, the running server's `settings.yaml` was rewritten:

- `issuer_from_request`, `issuer_from_proxy_headers`, `saml.export_roles`, `saml.export_groups` flipped back to `false`;
- `issuer_allowlist`, `device_verification_base_url`, `roles_attr_name`, `groups_attr_name` deleted entirely.

The test then "restores" afterwards, but only the fields it knows, so the wipe is permanent.

## Impact

- Any operator who saves the Settings page from a stale tab, or any script that posts a partial form, loses unrelated configuration with no warning.
- The e2e agent itself currently mutates and damages the config of the server under test, despite its config-agnostic design intent.
- Every new setting widens the blast radius: six were added since 2.5.0, and the C14N test predates all of them.
- This likely also explains the `settings.yaml` clobbering repeatedly encountered while developing #126/#128.

## Relation to #127

Sibling, but distinct. #127 is the serialization layer losing comments, inline `#` text and `${VAR}` placeholders on write. This issue is the form layer actively resetting values that were never on the form. Fixing #127 alone would not fix this one.

## Suggested direction

1. Make the handler's contract **"absent = unchanged"**: start from the current settings and apply only the fields actually present in the submitted form. Checkboxes are absent when unchecked, so each needs a companion marker to distinguish "unchecked" from "not on this form" (the usual hidden-input pairing).
2. Independently, the e2e "SAML Exclusive C14N" test should stop relying on a full-form round-trip for a single field, but fixing only the test would leave the operator-facing footgun in place, and every future setting would reintroduce it.

Target: before the next release, since the e2e suite currently damages the configuration of the server it tests.

Found during the toggles-on pass that produced #130.

PR #183, opened 2026-08-23
fix: derive SAML entity_id/sso_url from the effective issuer when unset (#181)

### LINKED ISSUE #181: fix: SAML entity_id/sso_url do not follow the effective issuer (issuer_from_request)
Use case: nanoidp is reached under more than one hostname at once, e.g. a Compose stack where the browser sees `http://localhost:9900` and other containers see `http://nanoidp:9900`, or an instance behind a reverse proxy. `oauth.issuer_from_request` (#126) makes the OIDC discovery document and token `iss` follow the request host, but `saml.entity_id` and `saml.sso_url` stay the fixed strings from `settings.yaml`: `routes/saml.py` reads `settings.saml_entity_id` directly and `routes/_issuer.py` is not involved. So a SAML SP fetching `/saml/metadata` through the proxy gets an `entityID` and an SSO location pointing at `http://localhost:8000/...` while OIDC on the same instance advertises the real host, and the `<Issuer>` in responses and assertions carries the same stale value. Today the only way out is a `${VAR}` placeholder per SAML field, set by hand to match the issuer.

Proposed fix, additive:
- when `saml.entity_id` / `saml.sso_url` are absent from the YAML, derive them from the effective issuer (`<issuer>/saml`, `<issuer>/saml/sso`), which already honours `issuer_from_request`, the proxy headers and the allowlist; an explicit value keeps winning, as today;
- metadata `entityID` and `SingleSignOnService` location, and `<Issuer>` in responses and assertions, all go through the same helper;
- `/api/config` and the MCP settings tool expose the effective values; `examples/test_agent.py` asserts SAML metadata `entityID` equals the OIDC issuer when `issuer_from_request` is on;
- the reverse-proxy guide gets a SAML paragraph.

RFC-citable: the metadata `entityID` must be the value the IdP uses as `<Issuer>` (SAML 2.0 Core 2.2.5, Metadata 2.3.2); VISION principle 2, metadata never lies.

cdelmonte-zg:
Closes #181.

- `Settings.saml_entity_id` / `saml_sso_url` become `Optional[str] = None`; `None` means "derived from the effective issuer" as `<issuer>/saml` and `<issuer>/saml/sso`. An explicit value wins exactly as before. With the default issuer the derived values equal the old hard-coded defaults, so untouched configs behave identically (test pins it); `${VAR}` placeholders still expand.
- One helper pair in `routes/_issuer.py` (`effective_saml_entity_id` / `effective_saml_sso_url`) built on `effective_issuer()`, so request reflection, proxy headers and the allowlist come for free. Used by `/saml/metadata` (`entityID` and both `SingleSignOnService` locations), the `<Issuer>` of SSO responses and assertions, the AttributeQuery response, `/api/config`, the settings page and the dashboard. SAML 2.0 Metadata 2.3.2 / Core 2.2.5 cited in code.
- Writer: a derived value is never materialized (`apply_settings_document` pops the key when `None`); `YamlWriter.update_saml_settings("")` clears, `None` leaves untouched (matches the #131 present-but-blank = clear contract). The UI form shows the effective value as a placeholder and posts `""` to keep it derived.
- Parity: MCP `get_settings` reports effective values plus `entity_id_derived` / `sso_url_derived` (resolved against the fixed `oauth.issuer`, same documented exception as the MCP discovery tools); `update_settings` gains both fields (empty string clears). `examples/test_agent.py`: new `SAML Metadata Follows Issuer` check (metadata == discovery issuer + `/saml[/sso]` when derived, == explicit values when not, `/api/config` agrees), and the two settings-form round-trips now post `""` for derived values instead of freezing them into explicit ones on every run.
- Shipped `config/settings.yaml`, the `nanoidp init` template, the wizard template and `examples/spring-boot-saml/settings.yaml` now carry both keys commented out with the "absent = derived" rule, so new setups get the derivation.
- Docs: `configuration.md` saml section, new "SAML metadata follows the issuer" section in the reverse-proxy guide, CHANGELOG.

## Verification

- 32 new tests (`tests/test_saml_effective_issuer.py`: derivation static/reflected/explicit, response and assertion `<Issuer>` == metadata `entityID` under a reflected Host with the XML parsed, writer never persists derived / preserves explicit / clears, UI placeholder and blank-POST semantics, MCP get/set/clear). Full suite 1066 passed, ruff and mypy clean.
- Live (`issuer_from_request: true`, no SAML keys): `Host: idp.test:9900` gives `entityID="http://idp.test:9900/saml"`, SSO location `http://idp.test:9900/saml/sso`, discovery issuer `http://idp.test:9900`, `/api/config` `entity_id_derived: true`; default Host gives `http://localhost:8795/saml`; an explicit `entity_id: "urn:fixed:idp"` wins regardless of Host with `entity_id_derived: false`. `test_agent.py` 51/51 and the YAML still has no SAML URL keys afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   file" use case only worked for settings. A UI/MCP save of one user still
   rewrites only that user's entry; the MCP `save_config` tool rewrites the
   whole map and materializes expanded placeholders, as documented.
+- **SAML `entity_id`/`sso_url` follow the effective issuer** (#181). With
+  `oauth.issuer_from_request` on (or behind a proxy), OIDC discovery reflected
+  the request host while `/saml/metadata`, the `<Issuer>` in responses and
+  assertions and the SSO location kept the fixed `http://localhost:8000/...`
+  strings. Both settings are now optional: absent (or blank in the UI/MCP)
+  means derived as `<effective issuer>/saml` and `<effective issuer>/saml/sso`
+  through one helper shared by every SAML surface, an explicit value still
+  wins, and a derived value is never written back to `settings.yaml`. SAML 2.0
+  Metadata 2.3.2 requires `entityID` to be the value used as `<Issuer>` (Core
+  2.2.5). `/api/config` and the MCP `get_settings` tool report the effective
+  values plus `entity_id_derived`/`sso_url_derived`; `update_settings` gains
+  `saml_entity_id`/`saml_sso_url` (empty string clears); the e2e agent checks
+  metadata against discovery and no longer posts derived values back as
+  explicit ones.
 - **Example presets now bind to `127.0.0.1`** (#164). All four pre-2.6.0
   presets (`cli-device-flow`, `microservices-client-credentials`,
   `react-spa-pkce`, `spring-boot-saml`) still shipped an explicit
--- a/book/src/guides/reverse-proxy.md
+++ b/book/src/guides/reverse-proxy.md
@@ -214,3 +214,17 @@ environment:
   the trust model behind `issuer_from_request`
 - [Endpoints reference](../reference/endpoints.md) - `GET /api/config`,
   `POST /api/config/reload`
+
+## SAML metadata follows the issuer
+
+`saml.entity_id` and `saml.sso_url` are optional. When they are absent from
+`settings.yaml`, nanoidp derives them from the effective issuer as
+`<issuer>/saml` and `<issuer>/saml/sso`, so `/saml/metadata`, the `<Issuer>`
+in responses and assertions, and OIDC discovery all name the same origin under
+every option above (fixed issuer, request-derived issuer, proxy headers,
+allowlist). Nothing SAML-specific needs to be configured for a proxy.
+
+Set them explicitly only when the SP expects a fixed, different value; an
+explicit value always wins and does not follow the request. `GET /api/config`
+reports the effective values together with `entity_id_derived` /
+`sso_url_derived` so you can tell which case applies.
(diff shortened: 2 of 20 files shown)


cdelmonte-zg:
Accidental indeed, fixed in 7affbfc. The two explicit lines came from the unit suite rewriting `config/settings.yaml` during my verification run (the conftest yaml_writer singleton gap noted on #176), and `git add -A` committed the rewrite on top of the commented block.

- Removed the two lines; the shipped preset now loads with `saml_entity_id is None` / `saml_sso_url is None`, i.e. derivation on.
- Guard tests added (`TestShippedConfigDerivesSamlUrls`): the shipped `config/settings.yaml` and `examples/spring-boot-saml/settings.yaml` carry no `saml.entity_id`/`sso_url` key, and the `nanoidp init` and wizard templates keep them commented. The shipped files are read at collection time, before any test runs, precisely because the suite can rewrite them mid-run; reading inside a test would check whatever an earlier test wrote, which is how this slipped through green CI.
- 1069 passed, ruff clean.

cdelmonte-zg:
Rebased onto main after #180 (head 49752da, CHANGELOG conflicts only, both sides kept). 1222 passed, ruff/mypy clean, `lint-imports` 2 kept; the shipped `config/settings.yaml` still has no explicit SAML keys after the full run. CI running on the new head; merging once green per your code-level approval.

cdelmonte-zg:
The 3.12 failure was the conftest gap itself: that job runs the suite twice (plain, then with coverage); the first run rewrote `config/settings.yaml` and the second run's collection-time snapshot read the rewritten file. Extracted the isolation fix as #184 (merged): every test now works on a per-test copy of `config/` via `NANOIDP_CONFIG_DIR`, and the `yaml_writer` singleton is reset. Rebased this branch on it (head 94928f1): suite run twice in a row locally, 1222 passed both times, `config/` clean both times. CI running; merging once green.

Issue #188, opened 2026-08-23
Public clients: token_endpoint_auth_method none with mandatory PKCE S256


cdelmonte-zg:
**Why.** Desktop agents, CLIs, MCP clients and native apps cannot keep a secret. nanoidp requires `client_secret` on every client (`OAuthClient.client_secret` is mandatory) and discovery advertises only `client_secret_basic` / `client_secret_post`, so a public-client flow cannot even be configured, let alone tested. It is the natural completion of the `oauth21` profile and of #81.

**Scope.**
- `OAuthClient.token_endpoint_auth_method: client_secret_basic | client_secret_post | none` (default: today's behaviour); `client_secret` optional and ignored when `none`.
- For `none` clients: PKCE required with `S256` only, regardless of profile (OAuth 2.1 draft 7.5.1 / RFC 7636); the `client_credentials` grant is refused (`unauthorized_client`); refresh tokens allowed but rotation forced (OAuth 2.1 draft 4.3.1 / 6.1, sender-constraining not available).
- `/token`: a `none` client is identified by `client_id` alone (plus PKCE); a secret presented for a `none` client is ignored; a missing secret for a confidential client stays `invalid_client`.
- `/revoke`: a public client may revoke with `client_id` only, and the token must belong to that client (RFC 7009 section 2.1 ownership check; otherwise `invalid_client` / silent 200 per the RFC's guidance, decided in the PR).
- `/introspect` stays authenticated: RFC 7662 requires the endpoint to be protected against token scanning and the caller to be an authorized protected resource; a public `client_id` is not authentication. `none` is NOT added to `introspection_endpoint_auth_methods_supported`. Public clients that need token state use `/revoke` and the token response, resource servers use their own credentials.
- Discovery: `token_endpoint_auth_methods_supported` and `revocation_endpoint_auth_methods_supported` include `none`; the introspection list does not.
- UI form and MCP schemas (there is no client-management REST surface today - `/api/config` exposes only `clients_count`); e2e: a public client completes the PKCE flow, is refused without PKCE, is refused `client_credentials`.

PR #197, opened 2026-08-23
refactor: load settings.yaml and users.yaml through document models (#175 piece 2)

cdelmonte-zg:
Piece 2 of #175: `YAML -> document models -> domain models`, with `${VAR}` and legacy keys kept at the edge. No format change, no semantic change for files that load today; the one visible difference is that an unknown key is now reported with its path instead of being silently ignored.

## What changed

- New `config_documents.py` (imports only `models`): one Pydantic model per YAML section (`ServerSection`, `OAuthSection` with `ClientEntry`, `SamlSection`, `JwtSection`, `SessionSection`, `LoggingSection`, `LoginSection`), `SettingsDocument.to_settings()`, `UserEntry.to_user()`, `UsersDocument.to_users()`, `load_settings_document()` / `load_users_document()`, `document_defaults()`. Field names are the YAML keys; defaults are the literals the loader used to hard-code. All `extra="forbid"` except `UserEntry` (`extra="allow"`: unknown user fields keep folding into `attributes`, the legacy behaviour). A before-validator turns a bare `login:` line (`None`) into `{}`, the one section the old loader special-cased (`data.get("login") or {}`); every other bare section, and every null collection the old loader did not `or []`, is a type error with its path, exactly as before. Null and missing are different: missing -> default, explicit null -> the domain validator's verdict (so `email: null` fails while an absent email defaults to `<username>@example.org`, and a blank `saml.roles_attr_name` still normalizes to `roles`). The #35 scalar coercions, the duplicate `client_id` check and every domain-model validator run unchanged, so nothing that failed before loads now (`port: 70000` is still refused by `Settings`).
- `config.py`: `_read_settings()` = read -> `check_config_version` -> `_expand_env_vars` -> `load_settings_document(...).to_settings()`; same shape for users (the version-agreement check from #182 kept). Unknown keys: each `extra_forbidden` error is logged as `WARNING <file>: unknown key <dotted.path> (ignored)`, removed, and validation retried; any other error raises `ValueError("<file>: invalid value at <path>: <msg>")`. No strict mode here (piece 4). `_set_default_settings()` builds from `SettingsDocument()` defaults plus the demo client.
- Writer defaults: `apply_settings_document(document, settings, defaults=None)` takes a flat `section.key -> default` mapping; `config.py` passes `document_defaults()`, `YamlWriter.update_login_settings` reads `document_defaults()["login.mode"]`. `serialization.py` keeps a two-entry fallback because it must not import package modules at runtime (#149 contract). Honest finding: the "~37 writer literals" counted in #175 are mostly "absent when empty" patterns tied to Optional semantics, not defaults; only `security_profile` and `login.mode` were true default literals, and those are now model-driven. The omit-at-empty logic was left as is: rewriting it as omit-at-default would change files.
- Keys that shipped files carry but the loader never consumed are declared as `Any` so they keep loading without a warning and without any new validation, and are still not consumed (`cors_allowed_origins`, `device_flow` as a whole, `logging.format`, `oauth.refresh_token_expiry_minutes`, `session.permanent`); they get a real type the day they get behaviour. `session.management_secret` (#176) is not in the models yet: #176 must add the field when it lands, or files declaring it will warn (noted there).

## Findings (pre-existing, not changed here)

- Loader defaults `roles` to `["USER"]` while `user_to_yaml` omits empty `roles`, so a user saved with `roles: []` reloads as `["USER"]`.
- The writer materializes managed keys a file omitted on first save, so a load/save of the shipped preset is not a byte no-op (placeholders survive and the second pass is a no-op; the test pins that real contract).
- `examples/react-spa-pkce/settings.yaml` does not load on `main` either: `client_secret: ""` and `OAuthClient` requires a non-empty secret. Tracked separately; it is what #188 (public clients) makes legitimate.

## Verification

- `tests/test_config_documents.py` (38 tests): every section default pinned to the previous literal; unknown key at top level / nested / inside a client entry -> WARNING with the path and identical `Settings`; wrong type -> error with path; the shipped `config/` and every `examples/*` file load with zero unknown-key warnings; round trip; `${VAR}` before validation; bare `login:` loads while bare `oauth:`/`saml:`/`server:`/`session:`/`jwt:`/`logging:` fail with their path; null collections and null user fields fail where they failed before; compatibility keys accept any value; blank SAML attribute names normalize. 1260 passed (1222 existing unchanged), ruff/mypy clean, `lint-imports` 2 kept.
- Live with `oauth.isuer: x` injected: `nanoidp.config_documents - WARNING - .../settings.yaml: unknown key oauth.isuer (ignored)`, `/api/config` unchanged, `examples/test_agent.py` 53/53. MCP server and test_agent needed no change (no surface change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **Configuration files load through document models** (#175, piece 2).
+  `settings.yaml` and `users.yaml` are now parsed into Pydantic document
+  models that mirror the YAML sections one to one
+  (`nanoidp.config_documents`), and the domain `Settings` / `User` objects
+  are built from them; the hand-written `.get(key, default)` mapping in
+  `config.py` is gone and the defaults live on the models, which the writer
+  reads as well. No file format change and no behavioural change for files
+  that loaded before. One visible improvement: an unknown key (a typo such
+  as `oauth.isuer`, or a key nanoidp does not know) is now logged as a
+  warning with its dotted path and ignored, instead of vanishing silently;
+  keys that shipped presets carry but the loader never consumed
+  (`cors_allowed_origins`, `device_flow`, `logging.format`,
+  `oauth.refresh_token_expiry_minutes`, `session.permanent`) are declared so
+  they do not warn. Fields inside a user entry keep folding into
+  `attributes`, as always. Stricter handling is a later piece.
+
 ### Config schema
 - `config_version` 1 introduced (#175, piece 1). No changes required to
   existing files: a file without the key is version 1. The version is the
--- a/book/src/reference/configuration.md
+++ b/book/src/reference/configuration.md
@@ -35,6 +35,10 @@ The contract is a single integer, not semver:
 - **Absent means 1.** Existing files need no change; `nanoidp init` and the
   wizard write the key into the files they create, and saves from the web UI
   or the MCP server preserve it if present and never add it.
+- **Unknown keys are reported.** A key nanoidp does not know (a typo such
+  as `oauth.isuer`) is logged as a warning with its path and ignored; the
+  file still loads. Inside a user entry, unknown keys become that user's
+  `attributes`, as they always have.
 - **Optional additions never bump it.** New optional keys with defaults keep
   the version; only renames, removals or semantic changes of existing keys
   do, and such a bump ships with a migration step in the loader.
--- a/src/nanoidp/config.py
+++ b/src/nanoidp/config.py
@@ -14,6 +14,12 @@ import yaml
 
 # Re-exported for compatibility: the models were defined here until #86, and
 # every consumer (routes, services, MCP, tests) imports them from this module.
+from .config_documents import (
+    SettingsDocument,
+    document_defaults,
+    load_settings_document,
+    load_users_document,
+)
 from .models import (  # noqa: F401
     SECURITY_PROFILES,
     OAuthClient,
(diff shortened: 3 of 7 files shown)


cdelmonte-zg:
Both findings confirmed against the old loader and fixed in 1935dc5.

**Null vs missing.** The old loader special-cased exactly one bare section (`data.get("login") or {}`); everything else went through `data.get(key, default)` and a null reached the domain validators. Restored: the before-validator now handles only `login:`; `oauth.clients`, `allowed_identity_classes`, `users`, and the user collections (`roles`, `groups`, `entitlements`, `source_acl`, `attributes`) are non-nullable with `default_factory` (missing -> default, explicit null -> `invalid value at <path>`). `issuer_allowlist` and `sp_certificates` keep accepting null because the old loader had `... or []` there. `email` uses `model_fields_set`: absent -> `<username>@example.org`, `email: null` -> rejected by `User.email` as before. `password` and `identity_class` stay nullable (Optional in the domain model). The two tests that had encoded the regression are replaced by tests asserting the old behaviour: bare `login:` loads, bare `oauth:`/`saml:`/`server:`/`session:`/`jwt:`/`logging:` fail with their path, null collections fail with their path, bare `users:` fails, each user field null fails while missing defaults.

**Compatibility keys.** `cors_allowed_origins`, `session.permanent`, `logging.format`, `oauth.refresh_token_expiry_minutes` and `device_flow.*` are now `Any`: accepted, never validated, never consumed, with a test feeding them `"*"`, `"whatever"`, `42`, `"soon"` and null. They get a real type the day they get behaviour.

**#176 note.** Agreed, and I will raise it on #176 at its next round: `management_secret` must keep distinguishing absent (env fallback) from null (off), so it needs `model_fields_set` or the decision at the raw-document edge; a plain `Optional[str] = None` on `SessionSection` would lose it.

1259 passed (1222 existing unchanged), ruff/mypy clean, `lint-imports` 2 kept. The PR body's "nothing that failed before loads now" is true again.

cdelmonte-zg:
Both fixed in 0914610: `device_flow: Any = None` (container shape and keys unvalidated; test feeds `whatever`, a bare line and `3`), and `roles_attr_name` / `groups_attr_name` are `Optional[str]` on the document so a bare key reaches `normalize_saml_attr_name` and becomes the default (test with `None`, blank and whitespace). 1260 passed, ruff/mypy clean, contracts 2 kept. PR description updated: only `login:` is special-cased, compatibility keys are `Any`, verification list matches the tests.

PR #199, opened 2026-08-23
feat: hooks and plugins v1, extension points for external config stores (#185)

### LINKED ISSUE #185: Hooks and plugins v1: extension points for external config stores, not backends
## Why

Two independent deployments needed the same thing in the last week: mirror nanoidp's configuration files to an external store (S3 in one case, Vault in #177) and render them back before start, split per stage. Both solved it by forking or by asking for a backend in the core. Neither belongs in the core: every store has its own SDK, auth, consistency and failure model, and an IdP meant for testing protocol behaviour must not depend on infrastructure to be correct. What the core should provide is a small, versioned extension point so that store integrations can be written and owned separately, as plugins.

This is the use-case-driven answer to #177 and to the "shared dev stack" scenario in general: **extension points, not backends.**

## Principle

Hooks observe and provide; they never replace persistence. nanoidp keeps talking only to its YAML files; plugins talk to the files and to the outside world. A plugin cannot become the storage, intercept HTTP requests or touch token issuance, so a broken plugin cannot alter protocol behaviour.

## Contract v1 (`HOOK_API_VERSION = 1`, additive-only from here)

Three hooks, called synchronously from `ConfigManager`:

- `on_before_load(config_dir: Path) -> None`: before settings/users are read (startup and every reload). Use: render files from a store into the directory.
- `on_config_saved(path: Path, kind: Literal["settings", "users"]) -> None`: after an atomic write of either file (UI, MCP, `save()`). Use: push to a store, `git commit`, notify.
- `on_audit_event(event: dict) -> None`: after an audit entry is recorded. Use: ship audit elsewhere.

Two ways to implement the same contract:

1. **Shell hooks**, zero API:
   ```yaml
   hooks:
     on_before_load: "aws s3 sync s3://idp-config-${STAGE} {config_dir}"
     on_config_saved: "vault kv put secret/idp/{kind} @{path}"
   ```
   Placeholders `{config_dir}`, `{path}`, `{kind}`; command run via the shell with a timeout.
2. **Python plugins** via the `nanoidp.plugins` entry-point group, packaged separately (`nanoidp-vault`, `nanoidp-s3`, ...). A plugin is a class exposing any subset of the three hooks plus `name` and `hook_api_version`; its configuration lives under `plugins.<name>:` in `settings.yaml`.

Both go through one call site per hook in the core, so there is one contract.

## Bootstrap: hooks that must run before settings.yaml exists

`on_before_load` runs before `settings.yaml` is read, but hook and plugin configuration is declared in `settings.yaml`: the main use case ("render settings/users from Vault or S3 before the load") cannot be configured by the file it is supposed to fetch. A minimal bootstrap surface outside the normal config resolves it, without turning into a backend:

- `NANOIDP_BOOTSTRAP_HOOK="<shell command>"` / `--bootstrap-hook`, and `NANOIDP_BOOTSTRAP_PLUGIN=<entry point name>` with plugin settings from `NANOIDP_PLUGIN_<NAME>_*` variables or an optional `bootstrap.yaml` next to the config dir (only `hooks:` and `plugins:` keys allowed, validated by the same schema).
- Bootstrap hooks run once before the first load; `settings.yaml` may then declare the same hook for reloads and saves. `nanoidp plugins` shows which surface each hook came from.

## Error policy, per hook

- `on_before_load`: the only hook that can block an operation, since it runs before any mutation. Default: log and continue with whatever is in the directory; `hooks.strict: true`: the load (startup or reload) fails with the hook's error.
- `on_config_saved`: runs after the atomic write, so the local save is always committed. Default: log. `strict`: the error is propagated to the caller (UI flash / `/api` 500 / MCP `isError`) so the operator learns the mirror failed, but the file on disk is what was written; the hook must be idempotent so a retry (next save or `POST /api/config/reload` followed by a no-op save) re-mirrors.
- `on_audit_event`: never propagates, `strict` or not. An audit plugin failure must not fail a `/token` whose token is already issued; that would let a plugin alter protocol behaviour, which is the one thing hooks are not allowed to do. Failures are counted and surfaced in `nanoidp plugins` / `/api/config`.

## Rules the core commits to

- **Error policy is explicit and per hook** (above); identical for shell and Python hooks.
- **Synchronous, with timeout** (`hooks.timeout_seconds`, default 10). No threads or async in the core; a plugin that wants asynchrony does it on its side.
- **Introspection**: `nanoidp plugins` lists loaded plugins, their `hook_api_version` and which hooks they implement; `GET /api/config` and the MCP `get_settings` tool expose the same (metadata never lies).
- **Schema** (#175): `hooks:` is modelled and validated; `plugins:` is the only section with `extra="allow"` (plugin-owned keys), everything else stays `extra="forbid"`.
- **Security** (SECURITY.md): a Python plugin runs with the process's privileges; installing one is a trust decision like any dependency. Shell hooks run whatever the YAML says; the YAML is operator-owned by definition (same trust boundary as `management_secret`).
- **No protocol hooks.** `on_token_issued`, `on_login` and the like are out of scope for v1 and would be a VISION decision, not a PR.

## Scope

- Core: hook dispatch in `ConfigManager` (load + both saves) and in the audit service; shell runner; entry-point loader; `hooks`/`plugins` settings; `nanoidp plugins`; exposure in `/api/config` and MCP; docs page "Extending nanoidp: hooks and plugins" with a worked `git commit` example and a skeleton plugin package; CHANGELOG.
- Tests: shell hook called with the right placeholders on load/save/audit; failing hook non-blocking by default and blocking under `strict`; timeout; entry-point plugin discovered and called; `hook_api_version` mismatch refused with a clear message; `plugins.<name>` config reaches the plugin; `/api/config` parity.
- e2e: `examples/test_agent.py` asserts the plugin listing and drives a shell hook that records its calls.
- Reference plugin: `examples/plugins/nanoidp-echo` (logs every hook) as the template; a store-specific plugin (`nanoidp-vault`) is explicitly left to whoever runs that store.

## Non-goals

- Any store-specific backend in the core (S3, Vault, git, k8s): VISION non-goal.
- Hooks on the protocol path.
- Replacing the YAML files as the source of truth.

## Ordering

After #175 piece 2 (the loader needs `extra="forbid"` with the `plugins:` exception), before dynamic client registration in the agentic milestone, since multi-agent use is where mirroring and conflict detection start to matter.

cdelmonte-zg:
Closes #185. Extension points, not backends: three synchronous hooks on the file and audit lifecycle, implementable as shell commands from YAML or as separately packaged Python plugins. nanoidp keeps talking only to its files; a plugin cannot become the storage, intercept requests or touch token issuance.

## Contract v1 (`HOOK_API_VERSION = 1`)

- `on_before_load(config_dir)` before settings/users are read (startup and every reload); `on_config_saved(path, kind)` after every atomic write of either file (one dispatch point, `ConfigManager.notify_saved`, reached by `ConfigManager.save()` and by every `YamlWriter` write); `on_audit_event(event)` after an audit entry is recorded.
- Shell hooks in `settings.yaml` (`hooks:` section, placeholders `{config_dir}`, `{path}`, `{kind}`, `{event_type}`; the audit event as JSON on stdin; `timeout_seconds`, default 10). Python plugins through the `nanoidp.plugins` entry-point group, configured under `plugins.<name>:`; `hook_api_version` other than 1 refused with a clear message. One dispatcher (`hooks.py`, `HookRegistry`): shell hook first, then plugins in declaration order.
- Bootstrap surface for hooks that must run before `settings.yaml` exists: `NANOIDP_BOOTSTRAP_HOOK` / `--bootstrap-hook` (run once before the first load, not on reload), `NANOIDP_BOOTSTRAP_PLUGIN` plus `NANOIDP_PLUGIN_<NAME>_<KEY>` variables, and `<config_dir>/bootstrap.yaml` (`hooks:` and `plugins:` keys only, unknown keys refused).

## Error policy, per hook (as in the issue)

- `on_before_load`: default log and continue; `hooks.strict: true` fails the load. Note in the guide: `strict` declared in `settings.yaml` cannot apply to the very first load, use `bootstrap.yaml` for that.
- `on_config_saved`: the local save is always committed first; default log; under `strict` the error is propagated to the caller after the write (UI flash "Settings saved locally; mirror hook failed", `/api` error, MCP `isError`), hooks must be idempotent. One post-write contract, owned by `YamlWriter._atomic_write`: write local -> notify the mirror -> refresh the runtime FROM LOCAL DISK (`ConfigManager.reload_local()`, which never runs `on_before_load`) -> propagate under `strict`. Disk and runtime therefore always carry the written value; a mirror that has not caught up cannot roll the write back. Only the explicit reload (startup, `POST /api/config/reload`, MCP `reload_config`) runs `on_before_load`. A multi-step save stops at the first failed write under `strict`.
- `on_audit_event`: never propagates, strict or not; failures are counted per hook.

## Introspection and parity

- `nanoidp plugins` lists hooks and plugins with API version, implemented hooks, failure counters and the surface each came from (`bootstrap-env`, `bootstrap.yaml`, `settings.yaml`); it is the only place that prints commands (local terminal, marked as possibly embedding secrets). `/api/config` and MCP expose `hook`, `source`, `failures`, `once` and never the command; propagated `HookError` messages are fixed-format, without command or stderr.
- Policy (`strict`, `timeout_seconds`) is kept per source: settings.yaml overrides only what it explicitly declares, bootstrap is the baseline, and dropping the settings source (missing file on reload) restores it. `timeout_seconds` is shell-only (`gt=0` on the document); plugin identity is the entry-point name.
- `GET /api/config` and the MCP `get_settings` tool expose the same `hooks` block; `update_settings` documents that `hooks:`/`plugins:` are YAML-only (same treatment as `secret_key` and `require_ui_login`). The writer leaves both sections untouched (read-modify-write), tested.
- `examples/test_agent.py`: `API Hooks Block` check, and a line-count check against `NANOIDP_E2E_HOOK_LOG` when the e2e server was started with a recording hook (skips cleanly otherwise).
- Reference plugin `examples/plugins/nanoidp-echo` (installable package with the entry point; logs every hook and records calls to a file from its settings), used in-process by the tests.

## Docs

`book/src/guides/extending.md` (contract, error-policy table, bootstrap, worked `git commit` and "render from a store before load" examples, plugin skeleton), linked from SUMMARY and the configuration reference; SECURITY.md paragraph on the trust boundary (plugins run with the process's privileges; shell hooks run whatever the operator-owned YAML says); MCP_WORKFLOW.md note; CHANGELOG.

## Verification

- 50 new tests (`tests/test_hooks.py`): placeholders on load/save/audit, bootstrap once and not on reload, default vs strict per hook with the file still written for `on_config_saved`, audit never propagating and counting, timeout, plugin discovery/config/call, API version mismatch refused, `bootstrap.yaml` unknown key refused, `/api/config` and MCP parity, writer preserves the sections, shipped configs still load with zero warnings; secrets never in `/api/config`/MCP/errors/flash; source precedence and lifecycle; `{config_dir}` on save and audit (git example runs for real); strict keeps disk and runtime aligned; a failed push with a stale mirror never rolls the write back (strict and non-strict) while an explicit reload still pulls from the mirror. 1310 passed, ruff/mypy clean, `lint-imports` 2 kept.
- Live: bootstrap hook wrote one `before-load <config_dir>` line and none on reload; a UI settings save produced `saved settings <path>` lines (four, the settings page does four atomic writes, pre-existing); failing `on_config_saved` in default mode logged with stderr and counted, in strict mode the file had the new value and the UI flashed the error; a failing `on_audit_event` (`false`) counted 2 failures across two `/token` calls that completed normally with their audit entries recorded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   configured under `plugins.<name>:`. Per-hook error policy: `on_before_load`
   may block under `hooks.strict`, `on_config_saved` is propagated to the
   caller under `strict` after the write (the local save is always
-  committed), `on_audit_event` never propagates. Bootstrap surface for hooks
+  committed and the running configuration reloaded from it, so only the
+  mirror is behind), `on_audit_event` never propagates. Commands are never
+  reported by `/api/config` or MCP (they may embed expanded secrets) and a
+  propagated error names the hook and its source only; the bootstrap
+  surface is the baseline for `strict`/`timeout_seconds`, `settings.yaml`
+  overrides only what it declares. Bootstrap surface for hooks
   that must run before `settings.yaml` exists: `NANOIDP_BOOTSTRAP_HOOK` /
   `--bootstrap-hook`, `NANOIDP_BOOTSTRAP_PLUGIN` with
   `NANOIDP_PLUGIN_<NAME>_<KEY>` settings, and `bootstrap.yaml` in the config
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (checked before `${VAR}` expansion).
 
 ### Added
+- **Hooks and plugins v1** (#185): extension points for external
+  configuration stores, not backends. Three synchronous hooks with
+  `HOOK_API_VERSION = 1`: `on_before_load(config_dir)` before the files are
+  read (startup and every reload), `on_config_saved(path, kind)` after an
+  atomic write of `settings.yaml` or `users.yaml`, `on_audit_event(event)`
+  after an audit entry. Implement them as shell commands under `hooks:` in
+  `settings.yaml` (placeholders `{config_dir}`, `{path}`, `{kind}`,
+  `{event_type}`, audit event JSON on stdin) or as Python plugins packaged
+  separately and discovered through the `nanoidp.plugins` entry-point group,
+  configured under `plugins.<name>:`. Per-hook error policy: `on_before_load`
+  may block under `hooks.strict`, `on_config_saved` is propagated to the
+  caller under `strict` after the write (the local save is always
+  committed), `on_audit_event` never propagates. Bootstrap surface for hooks
+  that must run before `settings.yaml` exists: `NANOIDP_BOOTSTRAP_HOOK` /
+  `--bootstrap-hook`, `NANOIDP_BOOTSTRAP_PLUGIN` with
+  `NANOIDP_PLUGIN_<NAME>_<KEY>` settings, and `bootstrap.yaml` in the config
+  directory (`hooks:` and `plugins:` only). `nanoidp plugins`, `GET
+  /api/config` and the MCP `get_settings` tool report what is loaded, from
+  which surface, with failure counters; `hooks:`/`plugins:` are YAML-only.
+  Reference plugin `examples/plugins/nanoidp-echo`; guide "Extending nanoidp:
+  hooks and plugins".
 - **Import contracts enforced in CI** (#149): `import-linter` now pins the
   package layering (`routes -> services -> config`) and the invariant that
   `serialization.py` has no runtime imports from the package (it is what
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -11,6 +11,7 @@
 
 - [Requesting tokens](guides/token-requests.md)
 - [Running behind a reverse proxy](guides/reverse-proxy.md)
+- [Extending nanoidp: hooks and plugins](guides/extending.md)
 - [MCP with Claude Code](guides/MCP_WORKFLOW.md)
 - [Security guide](guides/SECURITY.md)
 
--- a/book/src/guides/extending.md
+++ b/book/src/guides/extending.md
@@ -15,9 +15,9 @@ Three hooks, called synchronously:
 
 | Hook | When | Arguments | Typical use |
 |---|---|---|---|
-| `on_before_load` | before `settings.yaml` and `users.yaml` are read: startup and every reload | `config_dir` | render the files from a store into the directory |
-| `on_config_saved` | after an atomic write of either file (web UI, MCP, `save`) | `path`, `kind` (`settings` or `users`) | push to a store, `git commit`, notify |
-| `on_audit_event` | after an audit entry is recorded | the event as a mapping | ship audit elsewhere |
+| `on_before_load` | before `settings.yaml` and `users.yaml` are read: startup and every reload | plugin: `config_dir`; shell: `{config_dir}` | render the files from a store into the directory |
+| `on_config_saved` | after an atomic write of either file (web UI, MCP, `save`) | plugin: `path`, `kind` (`settings` or `users`); shell: `{config_dir}`, `{path}`, `{kind}` | push to a store, `git commit`, notify |
+| `on_audit_event` | after an audit entry is recorded | plugin: the event as a mapping; shell: `{config_dir}`, `{event_type}` plus the event as JSON on stdin | ship audit elsewhere |
 
 Two ways to implement it, sharing one dispatcher (shell hook first, then
 plugins in declaration order):
@@ -30,13 +30,21 @@ hooks:
   on_config_saved: "vault kv put secret/idp/{kind} @{path}"
   on_audit_event: "jq -c . >> /var/log/nanoidp-audit.jsonl"   # event JSON on stdin
   strict: false          # see the error policy below
-  timeout_seconds: 10
+  timeout_seconds: 10    # shell hooks only
 ```
 
-Placeholders: `{config_dir}`, `{path}`, `{kind}`, `{event_type}`. They are
-replaced textually, so other braces (`${VAR}`, `jq` filters) are left alone.
-The command runs through the shell with nanoidp's environment; for
+Placeholders are listed per hook in the table above. They are replaced
+textually, so other braces (`${VAR}`, `jq` filters) are left alone. The
+command runs through the shell with nanoidp's environment; for
 `on_audit_event` the event is also written to the command's stdin as JSON.
+
+A command is stored after `${VAR}` expansion, so it may embed a token. It
+is therefore never reported outside the process: `GET /api/config` and the
+MCP `get_settings` tool show a hook's name, source and failure counter but
+not its command, and the error a caller sees under `strict` names the hook
+and its source only. The command and the hook's stderr go to the server
+log at WARNING, and `nanoidp plugins`, which runs in your own terminal,
+is the one place that prints commands.
 
 **Python plugins**, packaged separately and discovered through the
 `nanoidp.plugins` entry-point group:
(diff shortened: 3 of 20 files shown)


cdelmonte-zg:
All five addressed in 16bb58a, re-verified live by me on top of the fork's run.

1. **Secrets.** `ShellHook.describe()` returns only `hook`, `source`, `failures`, `once` (no `command` key at all); the CLI `nanoidp plugins` is the only place that prints commands, marked "local only, may embed secrets". Propagated `HookError` messages are fixed-format: `<hook> shell hook (<source>) failed (exit N | timed out after Xs | could not run)` and `plugin '<name>' <hook> failed`; command, stderr and plugin exception text go to the server log at WARNING only. Tests (`TestSecretsNeverLeaveTheProcess`): a hook with `${MIRROR_TOKEN}`, the value absent from `/api/config`, MCP `get_settings`, the propagated message and the UI flash, and `"command"` absent from both documents. Live: `/api/config.hooks` = `{"strict": true, "shell_hooks": [{"hook": "on_config_saved", "source": "bootstrap.yaml", "failures": 0, "once": false}], ...}`, zero occurrences of the token or `command` in `/api/config` and in the page after the POST; the token appears in the local server log only.
2. **Precedence/lifecycle.** Policy is kept per source; `strict`/`timeout_seconds` resolve settings.yaml > bootstrap.yaml > default, and settings.yaml overrides only what it explicitly declares (`model_fields_set`); `drop_source()` drops that source's policy too and runs on the missing-file path of `_read_settings()`. `HooksSection.timeout_seconds` has `gt=0`. Tests: bootstrap strict+timeout survive a settings file without `hooks:` and a reload (the exact scenario from the review); explicit `strict: false` in settings overrides while an undeclared timeout keeps the bootstrap value; delete settings.yaml + reload drops its hook, plugin and policy while bootstrap ones stay; `timeout_seconds: 0/-1` in bootstrap.yaml refused.
3. **`{config_dir}`** on `on_config_saved` (and `on_audit_event`); the guide table lists the placeholders per hook; the git worked example is now an actual test against a `git init` directory under strict.
4. **Strict alignment, decided**: `YamlWriter._atomic_write` catches `HookError`, reloads, re-raises; the UI flashes `Settings saved locally; mirror hook failed: <message>`. Tests for the YamlWriter and UI paths (disk == runtime == new value, error raised). Live: after the failing strict save, `/api/config` audience `aligned-live` and `settings.yaml` line 7 `audience: aligned-live`.
5. **Contract**: `timeout_seconds` documented as shell-only (guide, `HooksSection`, `Settings`); plugin identity = entry-point name = `plugins.<name>` key, `name` removed from the contract and the echo plugin; guide notes that a directory renderer must preserve `bootstrap.yaml` or use the env/CLI bootstrap.

1306 passed (12 new), ruff/mypy clean, `lint-imports` 2 kept.

cdelmonte-zg:
Fixed in 4407ef5, and you are right that the non-strict case was the worse one (a silent rollback behind a success message).

- `ConfigManager.reload_local()`: refreshes settings and users from the local files, never runs `on_before_load`. `reload()` keeps its meaning as the external reload (startup, `POST /api/config/reload`, MCP `reload_config`) and still runs the hook.
- `YamlWriter._atomic_write` now owns the whole post-write contract exactly as you sketched: write local -> `notify_saved` (capture `HookError`) -> `reload_local()` -> re-raise if any. The ten per-method `get_config().reload()` calls are gone, so there is one path.
- Regression tests (`TestPostWriteRefreshNeverConsultsTheMirror`): `on_before_load` copies an old mirror over the config dir, `on_config_saved: false`; parametrized strict/non-strict: after `update_login_settings(mode="persona")` the local file and the runtime both say `persona`, the mirror is untouched, and under strict the `HookError` is raised. Plus: an explicit `reload()` afterwards does pull the old mirror back (that is the external reload's job), and `reload_local()` never fires `on_before_load` (counter test).
- Guide: hook table and error-policy row state that the post-write refresh reads local files only.
- PR body updated (now 50 new tests, 1310 passed, and the contract wording for the post-write path, the secret redaction and the per-source policy). ruff/mypy clean, `lint-imports` 2 kept.

PR #200, opened 2026-08-23
fix: hook registration under the error policy, JSON errors on reload, audit logging never loads config (review before 2.7.0rc4)

cdelmonte-zg:
Fixes for the eight confirmed findings of the code review run over `v2.7.0-rc3..main` before cutting 2.7.0rc4 (the security review over the same range had no findings). All in #199's hooks code except 5 (#178 parity).

1. **Plugin registration follows the error policy.** A plugin that cannot be loaded (entry point missing, `hook_api_version` mismatch, `configure()` raising) no longer aborts the settings load or the post-write refresh: logged at ERROR, listed in `describe()["plugins_failed"]` (`/api/config`, MCP, `nanoidp plugins` as `NOT LOADED`) and skipped; under `hooks.strict: true` a `HookError` naming the plugins is raised from the load, like a failing `on_before_load`. The public `reason` is one of nanoidp's own diagnoses (`not installed`, `incompatible hook_api_version=N (expected 1)`, `already loaded`, `initialization failed`); exception text from import/constructor/`configure()`, which may embed plugin settings, stays in the local log. Before, `settings.yaml` declaring `plugins: {store: {}}` on a host without the package broke every UI/MCP save.
   **A strict reload fails without commit**: the candidate `Settings` is promoted only after the registry applied the file's hooks/plugins, and `HookRegistry.replace_source()` restores a snapshot if registration raises, so a 503 reload leaves settings, profile hardening (`--profile stricter-dev`) and plugin instances exactly as they were; fixing the file and reloading again commits.
2. **JSON contract on reload.** `HookError(message, kind)`: `message` is the registry's synthetic text (never a command, stderr or plugin exception text), `kind` the hook name or `plugin_load`. `POST /api/config/reload` returns `{"status":"error","error": message, "kind": kind}` with 503 instead of a 500 HTML page; MCP `reload_config` returns `success: false` with the same `kind`.
3. **Audit logging never loads config and cannot deadlock.** `get_config_if_loaded()` (no lock, no construction) is what `audit.py` uses; a thread-local flag makes `run_audit_event` a no-op while `run_before_load` is executing, so a plugin that logs an audit event from `on_before_load` during `init_config` neither recurses nor blocks on the non-reentrant singleton lock.
4. **`bootstrap.yaml` parity with `settings.yaml`**: `safe_load` -> `expand_env_vars` -> the shared document loader, so `${VAR}` expands in plugin settings and a typo yields the `"<file>: unknown key ..."` warning / `"invalid value at <path>"` error instead of a raw pydantic `ValidationError`.
5. **MCP parity for #178**: `get_settings` exposes `profile_override` and the `effective` block; a test asserts `/api/config` and MCP agree on `security_profile`, `profile_override`, `effective`, `config_version`, `hooks`.
6. **No plugin churn on the post-write refresh**: `_configure_hooks_from` keeps an order-sensitive snapshot of the applied `hooks:`/`plugins:` sections and skips reconfiguration when unchanged, so plugins are instantiated once across saves and re-instantiated when their config or declaration order changes. Consequence, stated in the guide: a plugin that failed in non-strict mode is retried only when the declaration changes or on restart.
7. **Dead state removed**: the six `Settings.hooks_*`/`plugins` fields written by `to_settings()` and read by nothing; the registry is the only live state.
8. **Early return** in `run_audit_event`/`_dispatch` when nothing implements the hook, before any dict copy.

## Verification

- 1326 passed (16 new), ruff/mypy clean, `lint-imports` 2 kept; CodeQL: zero open alerts (alert 26 on `str(exc)` fixed by the structured `HookError`; 24/25 on audit.py are re-raises of alerts dismissed before, dismissed again with a note).
- Live: `plugins: {store: {}}` on a host without the package -> server starts, saves work, `/api/config.hooks.plugins_failed` names it with the reason; with `hooks: {on_before_load: "false", strict: true}` added and reloaded, `POST /api/config/reload` -> `503 application/json` with the message above; `examples/test_agent.py` 54/54; `effective` and `profile_override` present on both surfaces.

Note on strict semantics, unchanged and now documented: a failing `on_before_load` declared in `settings.yaml` under `strict` fails every reload until the file is fixed and the process restarted, because the registered hook runs before the file is re-read (the hook is supposed to render that file). With `strict`, a failed plugin at startup refuses to start, same as a failed `on_before_load`; non-strict stays the default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,9 +55,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `NANOIDP_PLUGIN_<NAME>_<KEY>` settings, and `bootstrap.yaml` in the config
   directory (`hooks:` and `plugins:` only). `nanoidp plugins`, `GET
   /api/config` and the MCP `get_settings` tool report what is loaded, from
-  which surface, with failure counters; `hooks:`/`plugins:` are YAML-only.
-  Reference plugin `examples/plugins/nanoidp-echo`; guide "Extending nanoidp:
-  hooks and plugins".
+  which surface, with failure counters and the plugins that could not be
+  loaded (`plugins_failed`: a missing package or a wrong `hook_api_version`
+  is reported, never fatal unless `strict`); `hooks:`/`plugins:` are
+  YAML-only. `bootstrap.yaml` goes through the same loader as
+  `settings.yaml` (placeholders, unknown-key warnings). A strict
+  `on_before_load` failure is a JSON `503` on `POST /api/config/reload` and
+  an error result from the MCP `reload_config` tool. Audit logging never
+  constructs the configuration and an audit event produced inside a load is
+  not dispatched to hooks. An unchanged `hooks:`/`plugins:` declaration is
+  not re-applied on the refresh that follows a local write, so plugins are
+  not re-instantiated on every save. Reference plugin
+  `examples/plugins/nanoidp-echo`; guide "Extending nanoidp: hooks and
+  plugins".
 - **Import contracts enforced in CI** (#149): `import-linter` now pins the
   package layering (`routes -> services -> config`) and the invariant that
   `serialization.py` has no runtime imports from the package (it is what
@@ -133,7 +143,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   profile_override=)`), the effective profile and its hardening are re-derived
   after every settings load, and a save serializes the DECLARED state
   (`ConfigManager.persistable_settings()`), so neither the override nor
-  the hardening it implies is ever written into the operator's file. `GET /api/config` exposes `security_profile`,
+  the hardening it implies is ever written into the operator's file. `GET /api/config` and the MCP `get_settings` tool expose `security_profile`,
   `profile_override` and the derived `effective` values; the e2e agent checks
   they are stable across a reload.
 - **`users.yaml` now expands `${VAR}` / `${VAR:default}` placeholders** like
--- a/book/src/guides/extending.md
+++ b/book/src/guides/extending.md
@@ -98,7 +98,7 @@ outside the normal config covers it:
 |---|---|---|
 | `NANOIDP_BOOTSTRAP_HOOK="<command>"` or `nanoidp --bootstrap-hook "<command>"` | an `on_before_load` shell command | once, before the first load |
 | `NANOIDP_BOOTSTRAP_PLUGIN=<name>` with `NANOIDP_PLUGIN_<NAME>_<KEY>=<value>` | a plugin and its settings (name upper-cased, `-` as `_`) | loaded once, then takes part in every hook |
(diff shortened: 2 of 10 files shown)


cdelmonte-zg:
All three in 783f241.

1. **`plugins_failed.reason` redacted.** New `PluginLoadError(message, public_reason)` for the failures nanoidp diagnoses itself (`not installed`, `incompatible hook_api_version=N (expected 1)`, `already loaded`); anything else from import, constructor or `configure()` is reported as `initialization failed`, with the exception text in the local log only, same principle as `_dispatch`. Regression test: a plugin whose `configure()` raises with `{settings['token']}` (from `${PLUGIN_TOKEN}`): the token is absent from `/api/config` and from the strict `HookError`, present in the log.
2. **Ordered snapshot.** Plugins are compared as `tuple((name, dict(cfg)) ...)`; test loads `pa, pb`, rewrites the file as `pb, pa` with identical configs, `reload_local()`, asserts the registry order is `pb, pa` and that a save dispatches `b` before `a`.
3. **Structured `HookError(message, kind)`.** `message` is always the registry's synthetic text, `kind` is the hook name or `plugin_load`; `/api/config/reload` returns `{"status":"error","error": exc.message, "kind": exc.kind}` (no `str(exc)` in the response, which is also what CodeQL alert 26 was about) and MCP `reload_config` returns the same `kind`. Tests: a strict `plugins: {missing: {}}` reload reports `plugin_load`, a strict failing `on_before_load` reports `on_before_load`.

1325 passed (4 new), ruff/mypy clean, `lint-imports` 2 kept. I will check CodeQL on the new head and dismiss alert 26 with a note if it still fires on the synthetic message.

cdelmonte-zg:
Fixed in db28f52 (+ docs in 920cd40): the reload is now fail-without-commit.

- `_read_settings` builds the candidate `Settings`, applies the file's hooks/plugins through the new `HookRegistry.replace_source()`, and only then promotes the candidate. `replace_source` takes a `snapshot()` of the registry (shell hooks, plugin instances, per-source policy, retired counters, failures) before `drop_source` + `configure_from_sections`, and `restore()`s it when registration raises. `_hooks_snapshot` is recorded only on success, so the next reload re-applies.
- Regression test (`TestStrictReloadFailsWithoutCommit`), your scenario: app created with `--profile stricter-dev` and a working plugin; the file gains `hooks: {strict: true}`, `plugins: {keep, missing}` and a changed audience; `POST /api/config/reload` -> 503 `plugin_load`; then `config.settings is old_settings`, `security_profile == stricter-dev`, `require_pkce/password_hashing/rate_limit_enabled` True, `debug` False, audience unchanged, `hooks.plugins == old_plugins` (same instances), `hooks.snapshot() == old_snapshot`, `strict` still False; fixing the file and reloading commits everything (audience, strict, hardening intact).
- Guide: plugin failure reasons, the non-strict retry rule (retried only when the declaration changes or on restart) and the fail-without-commit rule. PR body updated (counts, `kind`, the new semantics).

1326 passed, ruff/mypy clean, `lint-imports` 2 kept.

PR #201, opened 2026-08-23
test: exercise the installed reference plugin and the bootstrap surface end to end

cdelmonte-zg:
Closes the gap noticed after #199: the reference plugin `examples/plugins/nanoidp-echo` and the bootstrap surface were never exercised through a real installation. The unit suite registers the plugin in-process (shared venv), and the e2e workflow configured no hook, so `test_agent`'s hooks check degraded to "block present, 0 plugins".

- **e2e workflow**, both jobs: `pip install examples/plugins/nanoidp-echo` (entry-point discovery from a real package); the HTTP server starts with `NANOIDP_BOOTSTRAP_PLUGIN=echo` and `NANOIDP_PLUGIN_ECHO_RECORD`, plus a shell `on_config_saved` hook declared in `config/bootstrap.yaml` appending to `NANOIDP_E2E_HOOK_LOG`; the MCP job bootstraps the same plugin.
- **`examples/test_agent.py`**: with `NANOIDP_E2E_PLUGIN` set the hooks check no longer degrades: the plugin must be listed with `hook_api_version` 1 and all three hooks, and with `NANOIDP_E2E_PLUGIN_RECORD` its record file must contain `on_config_saved` (kind `settings`) and `on_audit_event` lines after the settings round-trip. Without the variable the behaviour is unchanged.
- **`examples/mcp_smoke_test.py`**: with `NANOIDP_E2E_PLUGIN` set, `get_settings` must list the plugin (parity with `/api/config`).

Verified locally replicating the workflow (package installed into the venv, bootstrap env + `bootstrap.yaml`): `test_agent` all green with `API Hooks Block` reporting the shell hook run and the plugin record (saves and audit events); MCP smoke `[OK] get_settings lists plugin 'echo' (hook API 1)`. Negatives: a wrong `NANOIDP_E2E_PLUGIN` name fails the check; with the package uninstalled the bootstrap currently aborts startup on `main` (that is finding 1 of #200, after which it becomes a listed `plugins_failed` entry). No production code touched.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -27,9 +27,21 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -e ".[dev]"
+          # The reference plugin, installed the way a user would: the entry
+          # point must be discovered from a real package (#185).
+          pip install examples/plugins/nanoidp-echo
 
       - name: Start server
+        # Bootstrap surface exercised for real: the echo plugin through
+        # NANOIDP_BOOTSTRAP_PLUGIN + NANOIDP_PLUGIN_ECHO_RECORD, and a shell
+        # on_config_saved hook through config/bootstrap.yaml (the unit suite
+        # registers plugins in-process; this is the only place the installed
+        # package and the env/bootstrap.yaml path are tested end to end).
+        env:
+          NANOIDP_BOOTSTRAP_PLUGIN: echo
+          NANOIDP_PLUGIN_ECHO_RECORD: ${{ github.workspace }}/echo-record.jsonl
         run: |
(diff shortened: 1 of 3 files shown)


cdelmonte-zg:
Fixed in fc5f6a5. With `NANOIDP_E2E_HOOK_LOG` set, `test_api_hooks_block` now requires an `on_config_saved` hook whose `source` is `bootstrap.yaml` (a missing or differently-sourced hook is a FAIL with the registered hooks in the message, never a skip), and the plugin named by `NANOIDP_E2E_PLUGIN` must carry `source == "bootstrap-env"` in addition to `hook_api_version` 1 and the three hooks. Without the variables the behaviour is unchanged.

Verified locally with three runs: the CI replica (bootstrap.yaml hook + bootstrap-env plugin) passes; no `bootstrap.yaml` with the log path set -> FAIL; hook and plugin declared in `settings.yaml` instead (sources `settings.yaml`) -> FAIL.

PR #204, opened 2026-08-23
feat: generated config schema, validate-config CLI and strict validation (#175 pieces 3-4)

cdelmonte-zg:
Pieces 3 and 4 of #175, closing the issue: the contract that rc4 shipped (config_version + document models) gets its published artifact, its lint entry point and its opt-in strict mode. Agreed exception to the 2.7 freeze: it completes the release's own theme and the runtime surface is opt-in only.

## Piece 3: generated schema + parity

- `nanoidp config-schema [--file settings|users|bootstrap] [--write]`: JSON Schema derived from the document models (deterministic output). Committed artifact `docs/schema/config.v1.json` (20 KB): `config_version: 1` plus three standalone schemas under `settings`/`users`/`bootstrap`, each directly usable with jsonschema (nested-resource refs do not resolve under non-schema keys, so the container is plainly a container). Documented caveat: `${VAR}` placeholders are strings, so a plain JSON Schema check flags `port: ${PORT:8000}`; `validate-config` is the check that runs the real loader.
- A test regenerates the document and fails with "schema drifted: run nanoidp config-schema --write" on divergence.
- Parity tests (`tests/test_config_schema.py`): every MCP `update_settings` property and every `settings.html` form field must map to a document-model field (explicit rename table); the YAML-only set (`secret_key`, `require_ui_login`, `hooks`, `plugins`) must NOT appear on either surface; stale rename entries fail. Verified by simulation: an injected `new_shiny_knob` on either surface fails a test.

## Piece 4: strict validation + CLI

- `config_validation: warn|strict` (document field, default `warn`) and server flag `--strict-config` (transient, wins over YAML, never persisted, same pattern as `--profile`). Under strict an unknown key in any of the three files refuses startup/reload with the same `"<file>: unknown key <path>"` text; warnings unchanged otherwise. One contract per directory: users.yaml/bootstrap.yaml follow settings.yaml's declared mode.
- `nanoidp validate-config [--config DIR] [--strict]`: same loaders, no server, and **no hook execution guaranteed by construction**: `config_validation.py` imports only `config_documents` + `serialization`, never `config` or `hooks`; tests assert a `touch`-ing `on_before_load` does not run (with a control test proving ConfigManager would run it). Exit 0 clean/warnings, 1 on errors or (with `--strict`) warnings; also honours `config_validation: strict` from the file (a directory the server refuses to boot cannot lint green).
- MCP parity: read-only `validate_config` tool (`{valid, findings}`), tool count 25 -> 26, exercised in `mcp_smoke_test.py`; the effective mode is reported as `config_validation` on `GET /api/config` and MCP `get_settings` (pinned by the existing api/mcp equality test).
- Docs: schema artifact linked from the configuration reference, "Validating your configuration" subsection with a pre-commit example; CHANGELOG.

## Verification

- 1379 passed (1326 on main, +53), ruff/mypy clean, `lint-imports` 2 kept.
- Live (scratch config with `oauth.isuer: x`): non-strict boot warns and serves (`test_agent` 54/54); `--strict-config` refuses with the message, port never opens; `validate-config` exit 0 with the finding printed, `--strict` exit 1; `config-schema` renders; `mcp_smoke_test` 26 tools, `validate_config: valid` plus the finding on the scratch dir.

Implementation notes worth review: strictness for the bootstrap registry is peeked from settings.yaml's raw `config_validation` at init (re-derived on every reload); strict fails fast on the first unknown key while `validate-config` lists everything; a strict refusal surfaces like every existing config error at startup (traceback ending in the clear message).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (checked before `${VAR}` expansion).
 
 ### Added
+- **Generated config schema, `validate-config` and strict validation** (#175,
+  pieces 3 and 4). Three additions to the config contract, none of which
+  restates it a seventh time:
+  - `nanoidp config-schema` prints the JSON Schema of `settings.yaml`,
+    `users.yaml` and `bootstrap.yaml`, generated from the document models
+    (`--file` for one of them, `--write` to regenerate the committed
+    artifact from a source checkout). The artifact is
+    `docs/schema/config.v1.json`: one standalone schema per file under the
+    keys `settings`, `users` and `bootstrap`, next to the `config_version`
+    they describe, ready to point an editor's YAML-schema support at. A test
+    fails when the committed file no longer matches the models, and parity
+    tests fail when the MCP `update_settings` tool or the web UI's settings
+    form grows a knob that is not a key of the contract - or offers one of
+    the YAML-only fields (`secret_key`, `require_ui_login`, `hooks`,
+    `plugins`).
+  - `config_validation: warn|strict` (top level of `settings.yaml`, default
+    `warn`) and the server flag `--strict-config` decide what an unknown key
+    does: log its path and keep loading, or refuse to start and refuse every
+    later reload with the same message. The flag wins over the file for that
+    run only and is never written back, like `--profile` (#172). One
+    contract per directory: `users.yaml` and `bootstrap.yaml` follow what
+    `settings.yaml` declares. Wrong types stay errors in both modes.
+  - `nanoidp validate-config [--config DIR] [--strict]` lints a
+    configuration directory without starting anything: one line per finding,
+    exit 0 when clean or with warnings only, exit 1 on errors and on
+    warnings under `--strict`. It reads the three files through the same
+    loaders the server uses and nothing else - no `ConfigManager`, no hook
+    dispatched, no plugin imported, `bootstrap.yaml` checked for its shape
+    only - so it is safe as a pre-commit or CI step on a directory whose
+    hooks name commands. MCP agents get the same check as the read-only
+    `validate_config` tool (`{valid, findings}`), which brings the MCP
+    surface to 26 tools.
 - **Hooks and plugins v1** (#185): extension points for external
   configuration stores, not backends. Three synchronous hooks with
   `HOOK_API_VERSION = 1`: `on_before_load(config_dir)` before the files are
--- a/book/src/guides/extending.md
+++ b/book/src/guides/extending.md
@@ -227,6 +227,12 @@ with nanoidp's environment: the file is operator-owned by definition, the
 same trust boundary as `secret_key` or `management_secret`. Neither surface
 is reachable from the web UI or MCP.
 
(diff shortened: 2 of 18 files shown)


cdelmonte-zg:
Fixed in 5bf0d4f: the transactional boundary is now the whole configuration directory, exactly as you framed it.

- `_stage_directory()` parses and validates everything into candidates, committing nothing: settings.yaml (candidate strictness derived from the file BEFORE validation but held in the staging, candidate Settings with the profile hardening already applied via a now-pure `_apply_profile(candidate) -> declared`), then users.yaml with the candidate strictness and the version-agreement check against the candidate version. Any exception leaves the manager untouched.
- `_commit_directory()` promotes the staging: the hook registry first (`replace_source()` is the only step that can still fail and rolls itself back, #200), then plain assignments of settings, `_declared`, `strict_config`, `config_version`, users, default_user. `reload_local()` goes through the same transaction (no `on_before_load`, as before).
- Regression tests (`TestDirectoryLoadIsTransactional`), your exact scenario: old settings+users+hooks, disk gains changed settings/hooks + `config_validation: strict` + a `defualt_user` typo; `reload()` raises and settings (same object), users (same object), default_user, `strict_config`, `hooks.snapshot()` and `_hooks_snapshot` are all unchanged; fixing users commits everything at once. Plus: the failed strict flip (`config_validation: strict` + `oauth.isuer` in one file) leaves the running mode `warn` on `/api/config`; and `--profile stricter-dev` hardening survives a failed users load.
- Docs and the MCP tool description now promise what the code does: settings.yaml and users.yaml validated on startup and every reload; bootstrap.yaml validated when the bootstrap surface loads, at startup, so `validate-config` reports what would stop the NEXT startup for that file; a failed load commits nothing.
- On the non-blocking note: agreed, a structured `ConfigValidationError` for `/api/config/reload` (like #200's `HookError.kind`) is the right follow-up; left out of this PR to keep the freeze exception narrow.

1382 passed (+3), ruff/mypy clean, `lint-imports` 2 kept. CI running on the new head.

cdelmonte-zg:
Both fixed in dd47ee6 (plus your non-blocking note, since it was one line):

1. **Demo client restored**: `_default_settings()` ships `demo-client`/`demo-secret` again; test pins that a directory without settings.yaml gets exactly that client and that `check_client` accepts it.
2. **validate-config mirrors bootstrap startup semantics**: `_validate_file(..., check_version=False)` for bootstrap.yaml, so `config_version: 2` there is the unknown-key WARNING the real startup produces, never a version ERROR; test asserts warning-only findings and no errors for that exact file.
3. **MCP `validate_config` strict default**: the argument now defaults to the manager's effective mode (explicit argument wins), so "what the next reload would hit" holds for a `--strict-config` process; tool description updated; test builds a strict manager on a clean dir, drops a typo on disk, and asserts `valid=false` by default and `valid=true` with `strict: false`.

1385 passed (+3), ruff/mypy clean, `lint-imports` 2 kept. CI running on the new head.

Process note, honestly: two intermediate pushes of this round briefly carried a red test because my local gate was `pytest | tail -1` (the pipeline exit is tail's). Caught and fixed before any review ping; my checklists now use `pipefail`. The pushed head is fully green locally.

Issue #215, opened 2026-08-25
Post-2.7.0 architecture and quality review: tracking


cdelmonte-zg:
Findings and follow-ups from the architecture/quality review run right after the 2.7.0 release (main at 33826cd).

State of the codebase: 12.4k src lines, 19.7k test lines (62 files), 88% total coverage, mypy `disallow_untyped_defs` on all of src/, two import-linter contracts enforced in CI (routes -> services -> config layering; serialization free of runtime package imports). Radon MI grades every module A except routes/oauth.py (B). The structure is sound; the debt is localized and named:

- `mcp_server.py` 1618 lines, `_execute_tool` CC 98 (F), 75% coverage, and guaranteed growth (MCP/HTTP parity + agentic milestone)
- `routes/oauth.py` 1539 lines with `authorize` CC 37 (E); `routes/saml.py` `sso` CC 39 (E)
- `routes/ui.py` 57% coverage, form flows covered almost only by the e2e agent, which CI runs without a management secret
- no written architecture map: the layering lives in a pyproject comment, choke points (`routes/_auth.py`, `routes/_issuer.py`, `hooks.py`) are undocumented
- the 13-point client-field flow feeds the worst serialization hotspots (`apply_settings_document` CC 44, yaml_writer updates CC 26/28)

Sub-issues, in execution order:

- [x] #210 architecture page in the book
- [x] #211 table-driven tool dispatch in mcp_server.py (before the agentic tools land)
- [x] #213 unit coverage for routes/ui.py form flows
- [x] #212 extract the /authorize and SAML /sso pipelines (after #186's scope work)
- [x] #214 design: declarative client-field registry (decide after #186)
- [x] #221 ui.logout dead route / Logout button lands on the end-session page
- [x] #222 mcp_server.py handler branch coverage (75% -> 90%+)

Not in scope: elasticity/hosted operation (a VISION non-goal, see #209) and anything protocol-behavioral.

---
2026-08-25 update: #210/#211/#213 shipped (PRs #216, #220, #218; total coverage 88% -> 91%, the CC-98 dispatcher is gone, plus one real bug fixed: regenerate-secret dropped branding fields). #212 waits for #186/#219 to land; #214 decides after it. #221/#222 added from the review's loose ends.

cdelmonte-zg:
[...]

Issue #214, opened 2026-08-25
Client-field parity test and model-derived settings plumbing (registry declined)


cdelmonte-zg:
From the post-2.7.0 architecture review. Adding one field to `OAuthClient` currently touches 13 coordinated places (models, config_documents, serialization client_to_yaml + merge_client_entry, the clients form template, ui create/edit/regenerate-secret, mcp _client_to_dict + two tool schemas + two handlers, schema regen). The flow is documented and pattern-matched by now (#186's plan recites it), but 13 coordinated touches per field is where regressions of the #32 family are born, and it feeds the repo's worst serialization hotspots (`apply_settings_document` CC 44, `yaml_writer.update_oauth_settings`/`update_saml_settings` CC 26/28).

Design question, not an implementation ticket yet: a single declarative registry of client fields (name, type, coercion, YAML round-trip, form widget, MCP schema fragment) from which the current 13 points derive, so a new field is one entry plus its enforcement logic.

To evaluate:

- whether the registry can be introduced incrementally (field by field) without a big-bang rewrite
- interaction with the Pydantic document models from #175 (the registry must not become a second source of truth next to them; ideally it derives FROM the model fields)
- whether the UI form and MCP schemas can be generated without losing their current hand-written help texts
- cost/benefit against the simpler alternative: keep the 13-point flow but add a parity test that fails when a model field is missing from any of the 13 surfaces

Timing: 2.8 or 3.0; decide after #186 lands (it adds `allowed_scopes` through the manual flow and is a good final data point on the pain).

cdelmonte-zg:
Decision (2026-08-25, after #186/#219 provided the final data point): the full declarative registry is DECLINED. Rationale: the 13-point flow, now documented on the Architecture page, demonstrably works when followed (#219 traversed it first-try clean); the historical regressions (#32, the branding loss fixed in #218) came from the REBUILD legs, which model_copy has since eliminated structurally; and a registry would risk exactly what this issue feared - a second source of truth next to the #175 document models, and generated forms/schemas losing their hand-written help texts.

In its place, two focused pieces, tracked here:

1. **Client-field parity test**: every field of OAuthClient must appear in mcp _client_to_dict, in the create_client and update_client tool schemas, and in the clients form template - a forgotten leg of the flow becomes a suite failure, the same treatment the settings contract already gets from #175 piece 3's parity tests. Intentional exceptions get an explicit, commented allowlist.
[...]

Issue #186, opened 2026-08-23
Per-client allowed scopes and invalid_scope


cdelmonte-zg:
**Why.** Every deliverable of this milestone (resource binding, a mock MCP server with `documents:read` / `documents:write` / `admin`, "insufficient scope" and scope-escalation tests) presumes that a client can only obtain the scopes it was granted. Today `/authorize` and `/token` accept any `scope` string: the scope is carried into the token and narrowed on refresh, but never checked against a registration, so any client can mint a token with `admin` by asking for it. `scopes_supported` advertises a fixed list.

**Scope.**
- `oauth.scopes_supported` is the global scope vocabulary (configurable; default today's list `openid profile email offline_access`); `OAuthClient.allowed_scopes` is a per-client subset of it. A client without `allowed_scopes` may obtain every globally supported scope, not any string: a scope outside the vocabulary is `invalid_scope` for everyone. This is a small behaviour change (today any string is accepted); it is what makes `scopes_supported` true (metadata never lies). If the `dev` profile must stay permissive, it is an explicit, documented dev relaxation (`oauth.scope_enforcement: off` only under `dev`), never the default for `stricter-dev` / `oauth21`.
- `/authorize`, `/token` (all grants, including `client_credentials` and device) and refresh: a requested scope outside `allowed_scopes` is `invalid_scope` (RFC 6749 4.1.2.1 / 5.2); an omitted `scope` means the client's default (its full allowed set, or `openid` as today when unrestricted).
- Access token `scope` claim and `/introspect` `scope` reflect the granted set; `/userinfo` scope gating unchanged.
- Discovery `scopes_supported` = the vocabulary (plus client-declared scopes only if they are added to the vocabulary; a client cannot widen it).
- UI client form, `/api`, MCP `create_client` / `update_client`: new field; e2e in `examples/test_agent.py`: a restricted client gets `invalid_scope` for an extra scope, an unrestricted client is unchanged.

**Out of scope.** Scope-to-claim mapping beyond what exists; consent screens.

dshvedchenko:
Plan: Issue #186 — Per-client allowed scopes and invalid_scope
Based on the current main branch (verified: no scope enforcement exists anywhere today — /authorize, /token, and the device flow all accept any scope string unchecked).

Design decisions to settle first
The issue text doesn't fully match what's in the repo. Four things need a decision before writing code:

[...]

PR #217, opened 2026-08-25
refactor: table-driven tool dispatch in mcp_server.py (#211)

### LINKED ISSUE #211: Table-driven tool dispatch in mcp_server.py
From the post-2.7.0 architecture review. `_execute_tool` is a single if/elif chain over 26 tools with cyclomatic complexity 98 (radon F), the worst in the repo, in the file that is guaranteed to grow the most: MCP/HTTP parity plus the agentic milestone (#186-#196) add tools with every feature.

Replace the chain with a dict of handler functions (tool name -> handler), one small function per tool, keeping behavior byte-identical:

- no signature or schema changes to any tool; this is mechanical
- the handler table and the `Tool(...)` declarations should live next to each other so a missing handler or a stale schema is caught by a trivial parity test (add one: every declared tool has a handler and vice versa)
- `call_tool`'s error shape and the admin-secret gate stay exactly where they are

Do this BEFORE the agentic milestone lands new tools, so #186 and later build on the table instead of extending the chain. A follow-up split of the file into a package (declarations / handlers / server wiring) can be its own issue if the table alone is not enough.

Metrics to beat: `_execute_tool` CC 98 -> every handler under CC 10; mcp_server.py coverage 75% should not drop.

cdelmonte-zg:
Second PR of the stacked series for #215, based on #216 (this diff contains only the dispatch change; review #216 for the docs page).

Mechanical transformation of `_execute_tool`'s 26-branch if/elif chain (cyclomatic complexity 98, radon F, the worst in the repo) into one `_tool_<name>` function per tool plus a `_TOOL_HANDLERS` dispatch table. No signature, schema, or behavior change: every handler body is the old branch body dedented, the section banner comments survive, and the unknown-tool `ValueError` contract is preserved in the new dispatcher (still unreachable on the protocol path, `call_tool` rejects unknown names first).

New `TestMCPToolHandlerParity`: `_TOOLS` and `_TOOL_HANDLERS` must declare exactly the same names, and `MUTATING_TOOLS` must be a subset - a declared tool without a handler is now a test failure at import parity, not a call-time surprise.

Metrics: dispatcher CC 98 -> 3; every handler is now individually readable. One residual hotspot: `_tool_update_settings` keeps its intrinsic CC 26 (it mirrors the settings YAML surface); that logic is the subject of the #214 design discussion, out of scope for a mechanical PR.

Verified: 1423 unit tests, ruff, black, mypy, import contracts all green. The agentic milestone's new tools (#186 onward) should now be added as table entries.

Closes #211.

### DIFF
--- a/src/nanoidp/mcp_server.py
+++ b/src/nanoidp/mcp_server.py
@@ -42,7 +42,7 @@ import json
 import logging
 import os
 import re
-from typing import Any, Optional, Tuple
+from typing import Any, Callable, Optional, Tuple
 
 import jwt as pyjwt
 from jsonschema import Draft202012Validator
@@ -135,7 +135,10 @@ def _check_admin_secret(
 
     provided_secret = arguments.pop("admin_secret", None)
     if not provided_secret:
-        return False, f"A management secret is configured. Provide 'admin_secret' parameter for {tool_name}."
+        return (
+            False,
+            f"A management secret is configured. Provide 'admin_secret' parameter for {tool_name}.",
+        )
     if not verify_secret(provided_secret, secret):
         return False, "Invalid admin_secret"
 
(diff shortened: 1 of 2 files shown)


cdelmonte-zg:
Closed by GitHub when the base branch was deleted by #216's merge, before it could be retargeted; a closed PR cannot change base. Continued as #220 (same branch rebased onto main, identical diff).

Issue #32, opened 2026-05-25
aud claim should by default be the client id


indivisible:
it's cool to be able to set a custom audience, but OIDC ID tokens "MUST contain the OAuth 2.0 client_id of the Relying Party as an audience value". this makes testing multiple clients impossible. also, the oidc "aud" claim may be an array, so that could be useful to be able to test that, I already see some libraries break from that

cdelmonte-zg:
Thanks for the report, you're right on both counts. I've opened a PR that fixes this.

  Summary of the change:

  **ID Token `aud` is now the client_id.** Per OpenID Connect Core 1.0 §2, the ID
  Token `aud` must contain the Relying Party's `client_id`. It was previously the
  static `oauth.audience`, which is exactly what made multi-client testing
  impossible. Now each client gets its own `client_id` as the audience, so you can
  test multiple clients independently.

  **Array `aud` is supported via a new per-client `additional_audiences` setting.**
  If a client has additional distinct audiences configured, `aud` is emitted as an
  array `[client_id, ...extras]`. In that case nanoidp also sets `azp` to the
  `client_id`, so you can exercise authorized-party validation in libraries that
  check it. (Note: `azp` is optional in OIDC Core, this is an intentional testing
  convenience, not a spec requirement.)

  Example config:

  ```yaml
  oauth:
    clients:
      - client_id: multi-aud-client
        client_secret: secret
        additional_audiences:
          - https://api.example.com
          - urn:service:billing
```
  produces an ID Token with:
```json
  {
    "aud": ["multi-aud-client", "https://api.example.com", "urn:service:billing"],
    "azp": "multi-aud-client"
  }
```
  One deliberate distinction worth calling out: the access token aud is left
  unchanged and still reflects oauth.audience (the resource audience, per RFC
  9068 §2.2). Only the ID Token follows the client_id rule. The new field is also
  configurable over MCP (create_client / update_client).

  This is a breaking change for relying parties that validated the ID Token aud
  against the old static value, so it's documented in the CHANGELOG.

PR #218, opened 2026-08-25
test+fix: ui form flows coverage, regenerate-secret branding fix (#213)

### LINKED ISSUE #213: Unit coverage for routes/ui.py form flows (57%), with and without the management gates
From the post-2.7.0 architecture review. `routes/ui.py` is at 57% statement coverage, by far the lowest in the repo (total is 88%). The form flows are exercised almost only by `examples/test_agent.py`, which CI runs WITHOUT a management secret; that is exactly the blind spot where PR #176's round-3 finding lived (a mutation that 302s to /login passes an `in (302, 303)` assertion and the suite stays green).

Add unit tests (Flask test client, in `tests/`) for the untested UI flows:

- user create/edit/delete, client create/edit/delete/regenerate-secret, settings forms, keys, claims, audit views and export
- each mutating flow in four gate states: no gates, `require_ui_login` only, `management_secret` only, both (unlock first, then mutate)
- negative assertions that distinguish a real success from a silent redirect: assert on the resulting state (user/client exists or changed), never only on the status code
- keep the e2e as-is; this issue is about not depending on it

Target: routes/ui.py to at least 85%, and no mutating UI endpoint without a gated negative test.

cdelmonte-zg:
Third PR of the stacked series for #215, based on #217 (this diff is only the new test file).

`routes/ui.py` was at 57% coverage, its form flows exercised almost only by `examples/test_agent.py`, which CI runs without a management secret - the exact blind spot of PR #176's round-3 finding (a gated mutation 302s to /login and passes a status-code-only assertion). New `tests/test_ui_forms.py` (45 tests) covers, through the Flask test client:

- login/logout session behavior, users create/edit/delete (parsed roles/groups/entitlements/attributes, blank-password-keeps-existing, duplicate-create does not overwrite, missing-user branches)
- clients create/edit/delete/regenerate-secret, including the #32 regression shape: regenerate must carry description, redirect_uris and additional_audiences through
- settings: the absent-is-unchanged contract (#131), the `__on_form` checkbox marker, and an invalid-expiry rollback
- keys (rendered, rotation changes the kid, PEM downloads, invalid type) with `jwt.keys_dir` pointed into tmp_path so tests never touch the repo's ./keys
- claims prefixes (core + custom) and the preview JSON endpoint, audit filters/search/export (json+csv+invalid)/clear
- the gates, asserting on resulting state and never only on status codes: `require_ui_login` alone, `management_secret` alone across endpoints the existing tests did not cover (clients/create, settings, audit/clear, wrong-secret unlock), and both together (login is not unlock)

Coverage: routes/ui.py 57% -> 87% (issue target was 85%). The residual misses are exception branches and the dead `ui.logout` route, which oauth_bp's `/logout` end-session rule shadows entirely - a known non-blocking finding from the #176 review, documented in the logout test rather than changed here.

Full suite: 1468 passed.

Closes #213.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Regenerating a client secret no longer resets the client's branding**
+  (#213 review): `/clients/<id>/regenerate-secret` rebuilt the client with
+  only five fields, silently dropping `background_color`, `header_color`
+  and `footer_color` and resetting `show_client_id`/`show_description` to
+  their defaults - the same rebuild-by-hand shape that lost
+  `additional_audiences` in #32. The route now copies the client and
+  changes only the secret, so every field (present and future) is carried.
+
 ### Security
 - **Opt-in `management_secret` mutation gate** (#163): one shared secret that
   gates state-changing calls across all three management surfaces - the MCP
--- a/src/nanoidp/routes/ui.py
+++ b/src/nanoidp/routes/ui.py
@@ -531,13 +531,12 @@ def client_regenerate_secret(client_id: str) -> ResponseReturnValue:
     try:
         new_secret = secrets.token_urlsafe(32)
 
-        updated_client = OAuthClient(
-            client_id=client_id,
-            client_secret=new_secret,
-            description=client.description,
-            additional_audiences=client.additional_audiences,
-            redirect_uris=client.redirect_uris,
-        )
+        # Copy the whole client and change only the secret: rebuilding it
+        # field by field is how #32 lost additional_audiences here, and how
+        # the branding fields (colors, show_* flags) were silently reset
+        # until #215's review caught it. model_copy carries every field,
+        # including ones added after this line was written.
+        updated_client = client.model_copy(update={"client_secret": new_secret})
 
         yaml_writer = get_yaml_writer()
         yaml_writer.save_client(updated_client, is_new=False)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,6 +11,7 @@ import pytest
 
 import nanoidp.config as config_module
 import nanoidp.mcp_server as mcp_server_module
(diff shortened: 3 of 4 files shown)


cdelmonte-zg:
Post-push update from a /code-review pass over the whole stack (10 findings, all addressed here; the other two PRs came back clean):

- **Real pre-existing bug found and fixed**: `/clients/<id>/regenerate-secret` rebuilt the client with only five fields, silently wiping `background_color`/`header_color`/`footer_color` and resetting `show_client_id`/`show_description` - the same rebuild-by-hand shape that lost `additional_audiences` in #32. Reproduced live (the hardened test failed on main's behavior), fixed with `client.model_copy(update={"client_secret": ...})`, CHANGELOG entry added. The original test was tautological (it only set the fields the route carried); it now sets and asserts the branding fields too.
- `_make_app` now ALWAYS points `jwt.keys_dir` into tmp_path: create_app eagerly initializes crypto at the cwd-relative `./keys`, so the gated fixtures were touching the repo's keys directory.
- The audit log singleton is now reset per test in conftest (it was the one process-global reset_singletons missed), and the gated audit test asserts the entry count GREW across the exempt login, so leftovers cannot satisfy its precondition.
- The audit filter tests now seed a non-login event and assert exclusion (`all(...)` plus an unfiltered-export contrast), instead of a filter-agnostic `any(...)`.
- New parametrized sweep: every POST endpoint ui_bp registers (11, including `/keys/regenerate`) redirects to /login under `management_secret`; the class docstring now also names the caveat that GET-shaped routes bypass the write guard by design.
- Kid rotation now asserts the persisted `kid.txt`, not only the in-memory singleton; the /keys page test asserts the kid is rendered; the missing-client_id test asserts the client count; the logout test accepts 200 or 302 so fixing the known ui.logout shadowing will not break it.

Declined (with reason): hoisting one shared settings-builder across test_management_secret.py / test_ui_login_required.py / this file - those two intentionally build a minimal config (empty clients) while this file copies the full preset; unifying them changes what their tests assert against, so it belongs in its own cleanup, not here.

1479 tests green, ruff/mypy clean.

Issue #212, opened 2026-08-25
Extract the /authorize and SAML /sso pipelines into named steps


cdelmonte-zg:
From the post-2.7.0 architecture review. The two main protocol entry points are the last big monolithic handlers: `routes/oauth.py:authorize` (CC 37, radon E) and `routes/saml.py:sso` (CC 39, radon E) interleave parsing, validation, audit and rendering in one function each; `token` (CC 20) is close behind.

Continue the extraction pattern that already worked (`services/redirect_uri.py`, `services/token.py`, and #186's planned `services/scope.py`): break each handler into named, individually testable steps (parse/validate request -> resolve client -> per-check validations with their audit events -> outcome). Pure logic moves to `services/`; the route function becomes the orchestration you can read top to bottom.

Constraints:

- zero behavior change: same error codes, same error shapes, same audit events with the same details; the existing tests (94%/92% coverage on these files) are the harness, and any diff in responses is a bug in the refactor
- coordinate with #186, which touches `authorize` and every `/token` grant: land after the scope work, or rebase it, but do not interleave the two
- `ruff`/`mypy`/import contracts stay green; the layers contract already forbids the wrong direction

Metrics to beat: authorize and sso under CC 15 each, no route function above CC 20 in the two files.

Issue #187, opened 2026-08-23
RFC 8707 Resource Indicators: resource binds the access token audience


cdelmonte-zg:
**Why.** MCP Authorization requires the client to send `resource=<MCP server URL>` on `/authorize` and `/token` and the authorization server to issue a token whose `aud` is that resource, so a token for one MCP server is useless at another. nanoidp's access token `aud` is the global `oauth.audience` (plus `additional_audiences` per client for the ID Token), so every client of one nanoidp gets the same audience and a wrong-audience test cannot be written.

**Scope.**
- Accept `resource` (one or more) on `/authorize`, `/token` (all grants) and `/device_authorization`; validate as absolute URI without fragment (RFC 8707 section 2); an unknown or disallowed resource is `invalid_target`.
- Per-client `allowed_resources: list[str]` (absent = any resource accepted, the dev-friendly default; whether `oauth21` / `stricter-dev` require the list is decided in the PR with an RFC citation).
- The authorization code and the refresh token remember the resource(s); `/token` may narrow to a subset, never widen.
- Access token `aud` = the resource(s); when no `resource` is sent, `aud` = `oauth.audience` as today (no behaviour change for existing clients). ID Token `aud` unchanged (client_id, #32).
- `/introspect` reports `aud`. No `resource_indicators_supported` metadata: RFC 8707 defines none and nanoidp does not invent metadata. The e2e agent and docs state that an MCP client must send `resource` on both `/authorize` and `/token` (MCP 2026-07-28).
- e2e: a token minted for resource A is rejected by mock MCP server B (depends on #191; until then assert on `aud` via `/introspect`); `invalid_target` for a resource outside `allowed_resources`.

**Out of scope.** Resource-specific token formats or encryption.

Depends on: #186 (per-client scopes).

cdelmonte-zg:
Implemented and landed on `main` via #268 (the 2.9 stack integration merge, sha 73ad177). Ships in 3.0.0. Closing.

PR #219, opened 2026-08-25
Per-client allowed scopes and `invalid_scop. `oauth.scop…

dshvedchenko:
…es_supported`

  is the global scope vocabulary (default `openid`, `profile`, `email`,
  `offline_access`, also what discovery's `scopes_supported` now advertises
  instead of a hardcoded list); a client's new `allowed_scopes` is an
  optional subset of it. A requested scope outside the vocabulary is
  `invalid_scope` for every client - a small behavior change, since any
  scope string used to be accepted unchecked; a scope outside a client's own
  `allowed_scopes`, when set, is `invalid_scope` for that client
  specifically. Enforced at `/authorize`, every `/token` grant (including
  `client_credentials`, RFC 6749 §4.4, which previously dropped any
  requested scope entirely) and `/device_authorization`; an omitted `scope`
  defaults to the client's full allowed set, or today's default when
  unrestricted. Every `/token` rejection - including the pre-existing
  refresh-token scope-narrowing check - now returns the RFC 6749 §5.2 JSON
  error shape (`{"error": "invalid_scope", ...}`) instead of a bare 400.
  `oauth.scope_enforcement: false` is a dev-only escape hatch back to the
  pre-#186 behavior (any scope string accepted, unchecked); refused outside
  the `dev` profile. `allowed_scopes` is settable from the clients UI form
  and the MCP `create_client`/`update_client` tools, same as
  `additional_audiences`/`redirect_uris`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `additional_audiences` in #32. The route now copies the client and
   changes only the secret, so every field (present and future) is carried.
 
+### Added
+- **Per-client allowed scopes and `invalid_scope`** (#186). `oauth.scopes_supported`
+  is the global scope vocabulary (default `openid`, `profile`, `email`,
+  `offline_access`, also what discovery's `scopes_supported` now advertises
+  instead of a hardcoded list); a client's new `allowed_scopes` is an
+  optional subset of it. A requested scope outside the vocabulary is
+  `invalid_scope` for every client - a small behavior change, since any
+  scope string used to be accepted unchecked; a scope outside a client's own
+  `allowed_scopes`, when set, is `invalid_scope` for that client
+  specifically. Enforced at `/authorize`, every `/token` grant (including
+  `client_credentials`, RFC 6749 §4.4, which previously dropped any
+  requested scope entirely) and `/device_authorization`; an omitted `scope`
+  defaults to the client's full allowed set, or today's default when
+  unrestricted. Every `/token` rejection - including the pre-existing
+  refresh-token scope-narrowing check - now returns the RFC 6749 §5.2 JSON
+  error shape (`{"error": "invalid_scope", ...}`) instead of a bare 400.
+  `oauth.scope_enforcement: false` is a dev-only escape hatch back to the
+  pre-#186 behavior (any scope string accepted, unchecked); refused outside
+  the `dev` profile. `allowed_scopes` is settable from the clients UI form
+  and the MCP `create_client`/`update_client` tools, same as
+  `additional_audiences`/`redirect_uris`.
+
 ### Security
 - **Opt-in `management_secret` mutation gate** (#163): one shared secret that
   gates state-changing calls across all three management surfaces - the MCP
--- a/book/src/reference/configuration.md
+++ b/book/src/reference/configuration.md
@@ -271,7 +271,21 @@ oauth:
       footer_color: "#e8f4f8"      # optional; hex only, the card's footer band
       show_client_id: true         # optional; default true
       show_description: true       # optional; default false
+    - client_id: "scoped-client"
+      client_secret: "secret"
+      description: "Client restricted to a scope subset"
+      allowed_scopes:           # optional; see "Registered scopes" below
+        - "openid"
+        - "profile"
   # logos_dir: "./static/logos"    # optional; defaults to src/nanoidp/static/logos
+  # scopes_supported:               # optional; the global scope vocabulary
+  #   - openid                      # (default: openid, profile, email, offline_access)
+  #   - profile
+  #   - email
+  #   - offline_access
+  # scope_enforcement: true          # optional; false is a dev-only escape
+                                      # hatch back to "any scope string is
+                                      # accepted" - see "Registered scopes"
 
 saml:
   # Both optional: when absent they are derived from the effective issuer as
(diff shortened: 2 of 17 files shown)


cdelmonte-zg (changes requested):
Reviewed empirically at head 79e9506: full suite green on the branch (1461), live server run of examples/test_agent.py 56/57 (the one failure is the known environmental "Redirect URI Exact Match" on a fresh `init` config, unrelated to this PR), zero em dashes, schema regenerated and current, CHANGELOG in the right place. The layout follows the plan exactly - services/scope.py is a clean pure module, every one of the 13 client-field legs is there, the e2e goes through self.session, and the client_credentials openid-strip (accepted at validation, never minted into an ID Token) is a good catch I had not asked for. The scope_enforcement_active property instead of the planned startup validator is a deviation I explicitly approve: it mirrors pkce_required/rotation_enabled and, as your docstring notes, stays correct under a --profile override that bypasses field validation.

One blocking finding, then a mechanical rebase.

**B1 (blocking): the re-validation call sites turn an absent scope into the client's full allow-list, so a refresh can GRANT MORE than the original grant.** In resolve_scope, an omitted scope defaults to `" ".join(client.allowed_scopes)`. That is the right behavior at the request surfaces (/authorize, /device_authorization, password, client_credentials - the issue text asks for exactly that). But the refresh grant and the authorization_code redemption call the same helper as a re-check, and there the defaulting inverts the invariant your own narrowing check enforces three lines above: a refresh must never widen. Reproduced live on your branch:

1. `scoped-client` has `allowed_scopes: [openid, profile]`; mint via password grant with NO scope while `oauth.scope_enforcement: false` (this is the legacy world: any refresh JWT signed before this PR, or before an operator sets allowed_scopes, has no scope claim - refresh tokens are stateless JWTs, they survive restarts and upgrades).
2. Enable enforcement (YAML edit + reload).
3. `POST /token grant_type=refresh_token` with no scope parameter.

Result: HTTP 200 with `scope: "openid profile"` on the response and in the access-token claim, **and an id_token in the response** - the original grant had no scope at all and never produced an ID Token. The same shape sits latent in _grant_authorization_code's re-check (unreachable today because /authorize always resolves a non-empty scope before minting a code, but the next refactor won't know that).

Fix I'd suggest: make the re-validation sites validate-only - an absent/None scope passes through unchanged, only a present scope is checked against vocabulary and allow-list. Either a `validate_only=True` keyword on resolve_scope that skips the allowed-set defaulting, or a separate thin `validate_scope()` used by the two re-check sites. Please add tests pinning it: a no-scope legacy refresh token refreshes to a no-scope token (and no id_token) with enforcement on and allowed_scopes set; and while you are in there, the refresh paths are currently the one surface with no tests at all - the narrowing rejection's new RFC-shaped JSON body (your decision 3/4 change) and the vocabulary/allow-list re-check both deserve one.

**Rebase needed (mechanical):** main moved after you branched - the stack from the post-2.7.0 architecture review landed (#216/#220/#218). Two of your hunks are affected:

- mcp_server.py: _execute_tool is now a dispatch table with one `_tool_<name>` function per tool (#220). Your three additions land in `_tool_create_client` / `_tool_update_client` / `_tool_get_settings` plus the two Tool schemas - same content, new location.
- routes/ui.py: regenerate-secret now does `client.model_copy(update={"client_secret": new_secret})` (#218), which carries allowed_scopes automatically - your `allowed_scopes=client.allowed_scopes` hunk becomes unnecessary, just drop it. Your tests/e2e still apply as-is.
- CHANGELOG: [Unreleased] now also has a Fixed section; your Added entry keeps its place.

Two non-blocking notes, no action needed unless you disagree: (a) /api/config and MCP get_settings report the EFFECTIVE enforcement under the key `scope_enforcement` - fine by the #172 effective-values practice, just flagging the declared-vs-effective mix; (b) /authorize answers invalid_scope as 400 JSON at the endpoint rather than the RFC 6749 4.1.2.1 error redirect - consistent with how this endpoint already reports unknown_client and redirect_uri errors, so I consider it the house convention, noted for the record.

Will re-verify the next round live, as usual.

cdelmonte-zg (approved):
Round 2 verified empirically at head 381f0ed, everything from round 1 is addressed:

- **B1 fixed and reproduced fixed**: same live scenario as the review (no-scope refresh token minted under enforcement off, allowed_scopes [openid, profile] set, enforcement flipped on, refresh with no scope parameter) now returns a token with NO scope claim and NO id_token - the legacy grant stays exactly as narrow as it was. The validate_only mode is precisely the right shape, and the docstring explaining why an absent original scope must not be manufactured is the best comment in the module.
- The refresh surfaces now have tests: legacy no-scope does not widen, the narrowing rejection carries the RFC 6749 5.2 JSON shape, re-validation rejects a scope the client may no longer have and accepts one it still may. Plus the five validate_only unit tests.
- Rebase audit vs current main: clean. The MCP hunks sit in the #220 dispatch table (_tool_create_client / _tool_update_client / _tool_get_settings plus the two schemas), the ui.py regenerate-secret hunk is gone as asked (model_copy carries allowed_scopes), CHANGELOG entry in place above the 2.7.0 section.
- Full suite 1528 green locally, CI 8/8 green on this head, live examples/test_agent.py 56/57 (the one failure is the known environmental Redirect URI Exact Match on a fresh init config, unrelated).

Ships in 2.8.0 and unblocks #187 (resource indicators) next in the milestone order, plus #212 on my side. Thanks - two-round turnaround on a finding this subtle is excellent.

cdelmonte-zg:
Soak invite: **v2.8.0-rc1 is out** and it is the first artifact carrying both of this cycle's features - your per-client scopes (this PR) and the management_secret gate you asked for in #163. `pip install --pre nanoidp==2.8.0rc1` / `ghcr.io/cdelmonte-zg/nanoidp:v2.8.0-rc1`; plain `pip install nanoidp` keeps resolving 2.7.0 and `:latest` stays on v2.7.0. If you run one rc in your shared stack, run this one: the gate plus scope enforcement against your agents is exactly the combination that needs real-world eyes before the final. Full notes: https://github.com/cdelmonte-zg/nanoidp/releases/tag/v2.8.0-rc1

PR #225, opened 2026-08-25
test: branch coverage for the MCP tool handlers, mcp_server.py to 100% (#222)

### LINKED ISSUE #222: mcp_server.py handler branch coverage (75%): error paths of the _tool_* handlers are untested
From the post-2.7.0 architecture review (#215). mcp_server.py sits at 75% statement coverage, the lowest in src/ now that routes/ui.py is done (#213); #211's dispatch-table refactor kept it from dropping but did not raise it.

What the missed lines actually are (coverage run on main):

- the domain-failure branches of the `_tool_*` handlers: user already exists / not found (`_tool_create_user`/`_tool_update_user`/`_tool_delete_user` area), client not found / duplicate (`_tool_update_client`/`_tool_delete_client`), and most of `_tool_update_settings`'s field branches (the largest block, ~lines 1389-1459)
- `_tool_generate_token`'s error paths and `_tool_reload_config`/`_tool_save_config` failure branches
- `main()` and the stdio bootstrap (~lines 1613-1669), which is process wiring, not logic

Approach:

- drive the handlers through the in-memory MCP client (`conftest.call_mcp_tool`), the same harness the existing wire-level tests use, asserting the `{"success": False, ...}`/`is_error` contract per branch - the dispatch table makes each handler small enough that this is mechanical
- for `_tool_update_settings`, one test per field family is enough; full combinatorics belong to the #214 discussion, not here
- `main()`/stdio bootstrap: either a minimal spawn test (the wheel-smoke job already boots the entry point) or an explicit `# pragma: no cover` with a comment saying the publish pipeline's smoke test covers it - decide in the PR, but stop counting process wiring against the module's signal

Target: mcp_server.py to 90%+ with the pragma decision made explicit, and no `_tool_*` handler with an untested failure branch.

cdelmonte-zg:
New tests/test_mcp_tool_branches.py (20 tests), all driving the real protocol path through the in-memory MCP client (conftest.call_mcp_tool): the domain-failure branches (user/client already exists, not found), the whole happy paths that had no test at all (get_user, delete_user, list_clients, decode/verify token roundtrip, reload_config, save_config incl. a forced failure, get_jwks, keys info + rotation), update_user and update_client across every field branch (colors and show_* flags included), the readonly-mode rejection, and the dispatcher's unknown-tool ValueError contract.

The issue's pragma decision, made explicit: main() and the __main__ block carry `# pragma: no cover` with a comment saying why - they are argparse + stdio process wiring, and the publish pipeline's wheel-smoke job already boots the nanoidp-mcp entry point for real. With that, mcp_server.py goes 75% -> **100%** (368 statements, 0 missed); the issue target was 90%+.

One latent coupling surfaced while writing the token roundtrip and documented in the fixture: `_tool_generate_token` goes through `get_token_service()`, which builds from the GLOBAL config discovery rather than from mcp_server's own ConfigManager - the same global-vs-own family as #176's B5 finding, harmless in practice because both usually resolve the same directory, but worth remembering if the MCP server ever grows a --config that can diverge from NANOIDP_CONFIG_DIR.

Full suite 1548 green; ruff and mypy clean.

Closes #222.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -75,7 +75,10 @@ jobs:
         run: |
           smoke/bin/python -c "import nanoidp; print('version:', nanoidp.__version__)"
           smoke/bin/nanoidp --help > /dev/null
-          command -v smoke/bin/nanoidp-mcp > /dev/null
+          # Execute (not just locate) the MCP entry point: --help runs main()'s
+          # argparse from the built wheel, which is what the coverage pragma on
+          # mcp_server.main relies on (#222/#225 review).
+          smoke/bin/nanoidp-mcp --help > /dev/null
           mkdir smoke-cfg && cd smoke-cfg
           ../smoke/bin/nanoidp init
           grep -q '127.0.0.1' config/settings.yaml
--- a/src/nanoidp/mcp_server.py
+++ b/src/nanoidp/mcp_server.py
@@ -1633,9 +1633,11 @@ def main() -> None:  # pragma: no cover
     """Run the MCP server.
 
     Excluded from unit coverage on purpose (#222): this is process
-    wiring (argparse + stdio bootstrap), and the publish pipeline's
-    wheel-smoke job already boots the nanoidp-mcp entry point for real
-    before anything is published.
+    wiring (argparse + stdio bootstrap). The publish pipeline's
+    wheel-smoke job executes the nanoidp-mcp entry point from the built
+    wheel (--help, which runs this function's argparse) before anything
+    is published; the stdio serve loop itself is exercised by every real
+    MCP session.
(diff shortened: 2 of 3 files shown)


cdelmonte-zg:
Review nit fixed (d6cb7fd): the wheel-smoke step now runs `smoke/bin/nanoidp-mcp --help > /dev/null` instead of `command -v`, so it executes main()'s argparse from the built wheel rather than only checking the script exists - and the coverage pragma's justification is reworded to say exactly that (plus that the stdio serve loop is exercised by every real MCP session, not by the smoke). Verified locally that `nanoidp-mcp --help` exits 0.

Issue #127, opened 2026-08-19
Saving settings via UI/MCP silently loses comments, inline '#' text, and ${VAR} placeholders


cdelmonte-zg:
## Summary

Saving settings through the web UI (`/settings`) or the MCP `update_settings` tool rewrites `config/settings.yaml` in a way that silently discards information the user had in the file: all comments, any inline text after a `#` in an unquoted value, and `${VAR:default}` environment-variable placeholders. The user gets no warning, and the loss is persisted to disk.

This surfaced concretely while reviewing #126, where a settings-save round-trip left the tracked `config/settings.yaml` with:

- `issuer`, `saml.entity_id`, `saml.sso_url` changed from `http://localhost:${PORT:8000}` to a hardcoded `http://localhost:8000` (while `server.port` still had `${PORT:8000}`),
- the `registered-client` description truncated from `(exact matching, issue #67)` to `(exact matching, issue`,
- explanatory comments removed.

## Mechanisms

Two independent causes combine:

1. **Comments and inline `#` are lost by the YAML round-trip.** `YamlWriter` loads and writes with PyYAML `yaml.safe_load` / `yaml.safe_dump` (`src/nanoidp/services/yaml_writer.py`). PyYAML does not preserve comments, and on load it treats ` #...` in an unquoted scalar as an inline comment, so `description: ... issue #67)` is already truncated to `... issue` before it is written back.

2. **`${VAR}` placeholders are replaced by their expanded value.** Config load expands env vars in memory (`_expand_env_vars`, `src/nanoidp/config.py`), and the settings serializer writes the expanded runtime value back (e.g. `oauth["issuer"] = settings.issuer` in `src/nanoidp/serialization.py`), so `${PORT:8000}` is written out as the resolved `8000`.

## Impact

Any user who edits `config/settings.yaml` by hand (comments to document their setup, `${PORT}` / other env placeholders for containerized runs, `#` characters in a description) and then saves settings once through the UI or MCP loses that content in their real config, silently. The `${PORT}` case is the most damaging: a `PORT=9900` run that used to advertise a matching issuer will, after one settings save, listen on 9900 but advertise `http://localhost:8000`.

## Suggested direction

- Preserve the original file on write: a comment- and quote-preserving round-trip (e.g. `ruamel.yaml` in round-trip mode), or a surgical text-level update that rewrites only the keys that actually changed and leaves everything else byte-for-byte.
[...]

PR #226, opened 2026-08-25
refactor+test: client-field parity tests and model-derived settings plumbing (#214)

### LINKED ISSUE #214: Client-field parity test and model-derived settings plumbing (registry declined)

cdelmonte-zg:
Implements #214 as decided (registry declined - see the issue comment): make a forgotten leg a suite failure, and collapse the real hotspots into one table derived from the models.

**Piece 1, client-field parity** (tests/test_client_field_parity.py): every OAuthClient field must appear in config_documents.ClientEntry, mcp _client_to_dict (one commented exception: the read surface never echoes client_secret), both client tool schemas, and the clients form template. The regenerate-secret leg needs no entry: model_copy (#218) cannot miss a field by construction.

**Piece 2, settings plumbing**: serialization gains `OwnedSetting`/`OWNED_SETTINGS` - one row per owned settings.yaml key with its section, Settings attribute and absence encoding (plain / omit-when-falsy / omit-when-none for the #181 derived keys). Three consumers now drive from it instead of one hand-written if-block per field:
- `apply_settings_document`: CC 44 (radon F, the worst left in the repo) -> 16; the two defaults-dependent keys (security_profile, login.mode) and the merged clients list stay explicit.
- `yaml_writer.update_oauth_settings`/`update_saml_settings`: CC 26/28 -> the file has no C-grade function left; the #131 absent-is-unchanged and #127 placeholder-preserving guards are stated once in `_apply_provided_settings`.
- `mcp _tool_update_settings`: CC 26 -> a loop over `_UPDATE_SETTINGS_FIELDS` plus a small normalizer map; updated_fields order and the response body unchanged.

**The derivation contract** (tests/test_settings_plumbing_parity.py): every table row must name a real Settings attribute and a real document-model field for its section; the two writer signatures must cover exactly their sections' rows; the MCP field tuple must equal the tool's input_schema. The models stay the single source of truth - the table cannot drift from them without a test failure, which is the property the declined registry was supposed to buy, at a fraction of the machinery.

One deliberate micro-unification, visible only in a corner: the original writers disagreed among themselves on whether clearing an optional key that is already absent/empty pops unconditionally or under the unchanged-guard; the shared helper always guards, so a literal empty value already on disk is never rewritten. No effective-settings change in any case.

1539 tests green (11 new), ruff/mypy/import contracts clean, live examples/test_agent.py 56/57 (known environmental failure only).

Closes #214.

### DIFF
--- a/src/nanoidp/mcp_server.py
+++ b/src/nanoidp/mcp_server.py
@@ -1401,6 +1401,58 @@ def _tool_validate_config(arguments: dict[str, Any], config: ConfigManager) -> d
     return validate_config_result(config.config_dir, effective)
 
 
+def _blank_to_none(name: str, value: Any) -> Any:
+    """ "" = clear: back to the derived/unset value (#181), mirroring the UI."""
+    return value or None
+
+
+def _normalize_update_list(name: str, value: Any) -> Any:
+    return _normalize_str_list(value, name)
+
+
+def _normalize_update_attr_name(name: str, value: Any) -> Any:
+    return normalize_saml_attr_name(name, value)
+
+
+# update_settings' writable fields, in the response's updated_fields order.
+# Names double as Settings attribute names; tests/test_settings_plumbing_parity.py
+# asserts this tuple against the tool's input_schema, so the two cannot drift.
+_UPDATE_SETTINGS_FIELDS: tuple[str, ...] = (
+    "issuer",
+    "issuer_from_request",
+    "issuer_allowlist",
+    "device_verification_base_url",
+    "issuer_from_proxy_headers",
+    "audience",
+    "token_expiry_minutes",
+    "saml_entity_id",
+    "saml_sso_url",
+    "saml_sign_responses",
+    "saml_export_roles",
+    "saml_export_groups",
+    "saml_roles_attr_name",
+    "saml_groups_attr_name",
+    "saml_c14n_algorithm",
+    "strict_saml_binding",
+    "saml_want_authn_requests_signed",
+    "saml_sp_certificates",
+    "verbose_logging",
+    "refresh_token_rotation",
+    "require_pkce",
+    "login_mode",
+)
+
+_UPDATE_SETTINGS_NORMALIZERS: dict[str, Callable[[str, Any], Any]] = {
+    "issuer_allowlist": _normalize_update_list,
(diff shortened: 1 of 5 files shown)


cdelmonte-zg:
Review finding fixed and pinned (2b462c3): the two attr-name rows are omit_when_falsy again, restoring the writer's historical "empty string drops the key so the default name applies again" contract - reproduced pre-fix (a blank submit persisted a literal roles_attr_name: '' while the suite stayed green, exactly because the Settings validator normalizes blank back to the defaults at load). New TestBlankClearsTheSamlAttrNames reads the persisted document directly: custom name on disk, blank submit, key gone; plus a positive case. The table row carries a comment on why the document side is unaffected (the validator means apply_settings_document never sees a falsy value there). Also reworded test_client_field_parity.py's docstring per the review: the parity tests cover the declarative surfaces (model/document/MCP read+schemas/template) and do not claim to prove the imperative legs applied a field - those stay on the per-feature tests. 1542 green.

Issue #229, opened 2026-08-25
Single config write pipeline with optimistic conflict detection


cdelmonte-zg:
*Revised 2026-08-25 after #235 (RuntimeStore): this issue is now exclusively about the declared-configuration side (`settings.yaml`, `users.yaml`). Runtime state has its own boundary and its own conflict detection in #235; nothing of it goes through this pipeline. The original text is in the edit history.*

**Today**: two owners of the persistence semantics. The UI's per-field saves go through `services/yaml_writer.py`; `ConfigManager.save()` persists whole documents itself via `serialization.atomic_write_yaml`. Both build entries in serialization.py (and since #226 drive scalar keys from `OWNED_SETTINGS`), so CONTENT cannot drift - but the read-modify-write cycle exists twice, and neither protects against a lost update: atomic rename prevents partial files, not two writers interleaving read -> modify -> replace. Last write wins, silently. This holds between two threads of one process and, more importantly, between processes: the UI's Flask process and a `nanoidp-mcp` companion writing the same directory.

**Why this is architectural rather than paranoia**: VISION supports a shared development stack reachable by several developers and their agents, and Direction 6 promises conflict detection for multi-agent use. The declared-configuration write path is that boundary on the file side.

**Shape of the fix (file-based, no database)**:
- one component owns check -> load -> mutate -> atomic replace, the `on_config_saved` hook dispatch and the `reload_local()` trigger; `ConfigManager.save()` and `YamlWriter` both call INTO it instead of each doing their own cycle, and end up with the same post-write contract (write -> notify -> reload_local -> raise a strict hook error);
- a document revision (hash of the bytes on disk, read when the caller loaded) travels with each write as an optimistic precondition; a write whose expected revision no longer matches the file is refused with `ConflictError` instead of clobbering - the API surface maps it to 409, the UI to a "reload and retry" message, MCP to an `is_error` result; no precondition = today's last-write-wins, stated as such;
- the check-and-replace section runs under BOTH a process-local lock (threads of one process) and an advisory cross-process file lock (one lock file per config directory, held from the revision check through the replace). The hash alone does not make the compare-and-swap atomic across processes: two processes can both verify the same revision and then each `os.replace` (reproduced in #234 review, 10/10 lost updates without the file lock). The hash is the precondition; the lock is what makes checking it meaningful. Readers never take the lock: the atomic replace already guarantees a complete file;
[...]

Issue #192, opened 2026-08-23
Runtime-created users and clients: disposable test state separate from declared configuration


cdelmonte-zg:
*Revised 2026-09-14 (third revision). The earlier versions framed this as a config/state split for multi-agent use on a shared SQLite store. The use case is more general and the first slice is smaller. Earlier texts are in the edit history.*

**Property.** Start from a declared baseline, create disposable identities during a test run, throw them away without changing the baseline. Runtime-created users and clients are execution state unless one of them is explicitly promoted to declared configuration.

**Why.** A CI job or an integration test needs `alice`, `test-admin` or `broken-client` for one scenario, other identities for the next one, and the declared files untouched at the end. Agents and dynamic registration (#190) produce the same kind of state; they are not the reason for it.

Today this cannot be done against a running nanoidp. Verified on main (9c8ca52):

- The REST API is read-only for users and clients.
- The UI forms write `users.yaml` / `settings.yaml` immediately, so what they create is declared configuration.
- The MCP server is a separate process with its own `ConfigManager`. `create_user("ci-alice")` succeeds there, but the IdP answering `/token` does not know her (password grant: `invalid_grant`). Inside the MCP process the change is an unsaved edit rather than runtime state: `reload_config` rebuilds from disk and drops it, and `save_config` writes every pending object at once. The server sees it only after `save_config` and a reload of the server (`POST /api/config/reload`), and by then it is declared configuration.

**Scope (first slice).**

- An HTTP surface in the IdP process, under `/api/runtime`:
  - `POST /api/runtime/users`, `GET /api/runtime/users[/<username>]`, `DELETE /api/runtime/users/<username>`, and the same for `/api/runtime/clients`.
  - `POST /api/runtime/users/<username>/promote` and `POST /api/runtime/clients/<client_id>/promote` write that one object into the declared file through the per-entry writer (#229) and remove it from runtime state. There is no bulk promotion.
  - `DELETE /api/runtime` removes every runtime user and client and answers `{"users_deleted": n, "clients_deleted": m}`. It never touches declared configuration.
  - Access follows every other management write (#163): the `management_secret` when one is configured, open otherwise, as the UI forms are today. No separate rule for these routes: without a secret, `/api` already lets any reachable host mint tokens for existing users and rotate keys, so a runtime identity adds no new capability. The binding warning at startup and `docs/SECURITY.md` name `/api/runtime` among the exposed writes.
[...]

PR #234, opened 2026-08-25
feat: shared compare-and-replace write primitive (#229 phase 2)

dshvedchenko:
Phases 1 and 2 of #229: one write path for the configuration files, with revisions and preconditions, coordinated across threads and across processes.

## The primitive

`src/nanoidp/config_writer.py` (a root-level peer of `config.py` and `serialization.py`, below both `routes` and `services` so the import-linter layer contract holds):

- `current_revision(path)` hashes a file's raw on-disk bytes (sha256, missing file included).
- `compare_and_replace(path, expected_revision, mutate)` loads the file, raises `ConflictError` if `expected_revision` no longer matches what is on disk, otherwise runs `mutate()` against the freshly loaded document and writes it back atomically. A `None` expected revision is unconditional, which is today's last-write-wins, so nothing changes for a caller until it passes a real revision.
- `compare_and_replace_many(items)` does the same for several files as one unit: every precondition is checked before any file is written.
- `ConflictError` mirrors `HookError`'s shape (message, kind), so each surface's existing except-clause pattern extends to a conflict without new branching.

The check-load-mutate-replace cycle runs under a process-local `threading.Lock` **and** an advisory cross-process lock on a lock file in the configuration directory (`fcntl.flock` on POSIX, `msvcrt.locking` on Windows, behind one shared bounded-poll loop). The cross-process lock is what makes this a real compare-and-swap for the case #229 was opened for: a UI worker and an MCP process, or two MCP processes, sharing one configuration directory. Without it, a two-process probe lost 10 of 10 updates. It fails as a named `LockUnavailableError` rather than a bare `OSError` or an indefinite hang when the filesystem does not support advisory locks or a peer holds it for more than 10 seconds.

`serialization.py` no longer shares one module-level ruamel `YAML` instance between every reader and every writer: `load_yaml_document`/`atomic_write_yaml` build a fresh, identically configured instance per call. The shared instance carries mutable parser/composer/emitter state, and readers never took the write lock, so a concurrent read could corrupt a write running under it (4 readers against 300 writes: 35 errors across six exception types, one of them inside `compare_and_replace`'s own load; zero with the factory). `load_yaml_document` also closes its read handle before parsing rather than holding it across the parse.

## ConfigManager.save() migrated

`_save_users`/`_save_settings` route through the primitive instead of their own load/write calls, and `save()` becomes one coordinated transaction: both preconditions checked, then both files written, both `on_config_saved` hooks fired, and the runtime refreshed from disk exactly once before any `hooks.strict` failure is raised. Previously a strict hook failure on `users.yaml` left `settings.yaml` unwritten even when nothing was wrong with it.

`save()` gains optional `expected_users_revision`/`expected_settings_revision`; no caller passes one yet, that lands with the surface that needs it (phases 4 and 5). The runtime refresh can itself fail, so a new `ReloadAfterSaveError` says "both files ARE written, the runtime could not adopt them" without ever replacing or hiding a pending strict hook failure. MCP's `save_config` carries a `kind` for all three failure shapes (`conflict`, the hook's own kind, `reload_after_save`) so a caller can tell them apart without parsing error text.

## Also in here, found by the review's stress probe

`atomic_write_yaml` restored from its `.bak` on **any** exception, including a failed `os.replace()` that never touched the target. Since `os.replace()` either lands or does not, there was nothing to restore in that case, and the restore turned a clean no-op failure into a non-atomic in-place overwrite of a live, correct file, which a concurrent reader can observe mid-flight (reproduced on POSIX at 43 truncated reads out of 983). Now guarded, and pinned by `TestFailedReplaceDoesNotCorruptTheLiveFile`.

A `windows-smoke` CI job was added, which is what surfaced that bug and the Windows-only reader limitation below.

## Known limitation, deferred to #246

`TestConcurrentReadsOutsideTheLock` asserts that concurrent readers never see an error during a write. That property does not hold on Windows today and cannot be fixed from inside this PR: `load_yaml_document` is only one of several direct readers of these files (`_stage_directory`, `config_validation.py` and `hooks.py` read them on their own, entirely outside `compare_and_replace`), so a retry or a handle-close scoped to that one function cannot make it true system-wide. Whether readers join the write-coordination protocol at all is a design question about where the configuration's consistency boundary lies, which is #246. The test is marked `xfail` on win32 only, non-strict, and stays active and required on POSIX so the property remains visible rather than loosened or deleted.

## Verification

1597 tests, mypy, ruff and `lint-imports` (2 contracts kept, 0 broken) green; CI green on 3.10/3.11/3.12, the e2e jobs, CodeQL and `windows-smoke`. Phases 3 to 5 (`YamlWriter`, then the callers that supply a real revision) follow.


cdelmonte-zg (changes requested):
Verified at 305b5a8: 1575 tests and mypy clean here too, based on main's tip, and the atomic rename holds under load (four raw-bytes readers against 300 writes, zero partial files). The race test is the right test and the lock is the right fix for the TOCTOU. Global is fine, keep it: `_yaml_rt` is one instance shared by every file so per-file locks would not protect it anyway, `save()` writes two files back to back, `reload_local()` stages the whole directory (#204), and #192 will need a cross-file check that only works under one lock.

**Blocking, reproduced**: the lock does not close the ruamel problem, and the module docstring claims it does. `_yaml_rt` is also the loader (serialization.py:221), and loads run outside your lock everywhere: `reload_local()`, `YamlWriter._load_*`, every read API. Probe: four threads calling `load_yaml_document` in a loop while one thread runs 300 `compare_and_replace` calls: 35 errors of six kinds (`ComposerError`, `ScannerError`, `ParserError`, `DuplicateKeyError`, `IndexError`, `AttributeError`), and one of them fired *inside* `compare_and_replace`, on its own `load_yaml_document` under the lock, corrupted by a reader outside it. Same probe with a fresh `YAML(typ="rt")` per call: zero errors. So the fix belongs in `serialization.py`: build the `YAML` instance per call (a small factory carrying the indent/quotes/width settings; cost is irrelevant at this write rate and nothing shared is left), or give `_yaml_rt` its own lock taken by both `load_yaml_document` and `atomic_write_yaml`. I prefer the factory. Then the docstring can say what is true: the write lock is about the TOCTOU, the loader is thread-safe on its own. This is latent on main today (YamlWriter plus any concurrent read), but phase 1 is the primitive that promises the guarantee, so it has to close it here. A regression test in the shape of the probe above (readers outside the lock while writes happen) would pin it.

Two non-blocking notes:

- A no-op `compare_and_replace` on a hand-written preset changes the bytes (sequence indentation only: `_yaml_rt` uses `offset=0`, the presets use 4 spaces), so the revision moves on the first write to a pristine file even when nothing changed. Acceptable for a bytes-based revision, worth one sentence in the docstring so nobody hunts a phantom writer later.
- An empty file and a missing file share a revision (`_read_bytes` returns `b""` for both). Harmless today; just noting it since the docstring sells the missing-file revision as a create-only precondition.

dshvedchenko:
fix: build a fresh ruamel YAML instance per call (#229 review, blocking)

Review at 305b5a8 (reproduced): compare_and_replace's write lock only
serializes writers against each other, but the shared module-level
_yaml_rt instance was also the loader used by every read path
(reload_local(), YamlWriter's loaders, every read API) - none of which
take that lock. A concurrent load_yaml_document could corrupt an
unrelated load_yaml_document/atomic_write_yaml running under the lock,
since ruamel's YAML instance carries mutable parser/composer/emitter
state that isn't safe for concurrent use. Probe (4 reader threads calling
load_yaml_document in a loop against 300 compare_and_replace writes):
35 errors across six exception types, including one inside
compare_and_replace's own load call.

serialization.py replaces the shared _yaml_rt with _new_yaml_rt(), a
factory building a fresh, identically-configured YAML(typ="rt") instance
per call to load_yaml_document/atomic_write_yaml. Same probe with the
factory: zero errors. Cost of building an instance is irrelevant at this
write/read rate.

config_writer.py's docstring is corrected to say what's actually true:
_write_lock is only about the check-then-write race (TOCTOU) and the
per-file-vs-global choice; concurrent YAML parsing safety is now the
loader/dumper's own job, not the lock's. Also documents two accepted
edge cases from the review: a no-op write can still change the revision
on a hand-written file's first save (ruamel normalizes sequence indent
to offset=0), and a missing file shares a revision with an empty one.

tests/test_config_writer.py adds the review's probe as a regression test
(TestConcurrentReadsOutsideTheLock) - verified it fails with the old
shared instance (DuplicateKeyError) and passes with the fix. Full suite
and mypy clean.

cdelmonte-zg (changes requested):
Round 2 verified at aea4c2a: 1576 tests and mypy clean here, the reader/writer probe that produced 35 errors on 305b5a8 now produces zero, and the regression test covers exactly that shape. The ruamel finding is closed and the docstring now says what is true. Thanks for the fast turnaround.

**Still blocking, and it is a different finding: the CAS is not a CAS across processes.** The docstring admits it ("there is no cross-process file lock anywhere in this codebase"), but #229 was opened precisely with the multi-process / shared-stack case as the reason the hash check exists, so a primitive named and documented as optimistic conflict detection cannot leave that case open. Reproduced with two real processes:

```
process A                      process B
read revision R
                               read revision R
lock A (thread lock, local)
check == R  ok
load, mutate, os.replace
unlock
                               lock B (a different lock)
                               check == R  <- read before A's replace landed
                               load, mutate, os.replace
=> B overwrites A, both callers got "ok"
```

Probe: `multiprocessing` (fork), a `Barrier(2)` so both processes enter `compare_and_replace` together with the same base revision, a 50 ms sleep inside `mutate` to widen the window. Result: **10 of 10 trials, both writers succeed, no `ConflictError`, one update lost.** `os.replace` makes the single replace atomic; it does not make check-then-replace atomic.

The fix that fits this design is an advisory file lock around the whole check -> load -> mutate -> replace section, held in addition to the thread lock (the thread lock still serializes threads inside one process; `fcntl.flock` is per open file description, so two threads of one process would not exclude each other through it alone). Concretely:

- one lock file per config directory, e.g. `file_path.parent / ".nanoidp-write.lock"`, opened and `flock(LOCK_EX)`-ed for the duration of the section, released in a `finally`; one lock for all files in the directory, consistent with the global thread lock and for the same reasons (`save()` writes two files, #192's cross-file check);
- readers do not take it: the atomic replace already guarantees they see a complete file;
- POSIX `fcntl` is enough for the platforms nanoidp runs on in CI (Linux/macOS); if Windows must keep working, `msvcrt.locking` behind the same helper, otherwise state POSIX-only in the docstring rather than silently degrading;
- the docstring drops the "narrows rather than eliminates" paragraph, because after this it eliminates;
- a two-process regression test in the shape of the probe above (fork context, barrier, sleep inside mutate): exactly one `ok`, exactly one `ConflictError`, file content equals the winner's. This is the test that would have caught it, so it is the one to keep.

No SQLite here, no new dependency: an advisory lock on a file-backed store is the proportionate tool, and it is all #229 needs. The runtime-state side is a separate boundary (#235) and does not go through this primitive.

dshvedchenko:
fix: cross-process advisory lock for compare_and_replace (#229 review round 2, blocking)

Round 2 review at aea4c2a: the CAS was not actually a CAS across
processes. _write_lock only serializes threads within one process, and
#229 was opened specifically for the multi-process case (a UI worker and
an MCP process, or two MCP processes, sharing one config directory).
Reproduced with multiprocessing (fork), a Barrier(2) so both processes
enter compare_and_replace with the same base revision together, and a
50ms sleep inside mutate to widen the check-to-replace window: 10/10
trials, both writers got "ok", one update silently lost. os.replace
makes the single replace atomic; it never made check-then-replace atomic
across processes.

config_writer.py adds _cross_process_lock(directory), an advisory
fcntl.flock on one lock file per config directory
(.nanoidp-write.lock) - one lock for the whole directory, not per
target file, for the same reason _write_lock already is: save() writes
two files back to back, and #192's planned cross-file check only works
under one lock. compare_and_replace now holds both _write_lock and this
file lock for the whole check-load-mutate-replace section; the thread
lock stays because flock is scoped to the open file description and
doesn't reliably exclude two threads of the same process the way it
excludes two processes. POSIX only (fcntl) - the platforms this project
runs on in CI - stated as a hard requirement in the docstring rather
than silently degraded; a Windows deployment would need msvcrt.locking
behind the same helper. Readers still don't take this lock: the atomic
replace already guarantees they see one complete revision or the next.

Module docstring rewritten to drop the "narrows rather than eliminates"
claim, since it no longer applies.

tests/test_config_writer.py adds the review's exact probe as a
regression test (TestCrossProcessConflictDetection, multiprocessing
fork + Barrier + sleep-in-mutate) - verified it fails (both processes
"ok", update lost) with the file lock removed and passes with it
restored. Full suite (1577 tests) and mypy clean.

cdelmonte-zg (approved):
Round 3 verified at 9d49ef7, and this closes it.

- Two-process probe (fork, `Barrier(2)`, 50 ms inside `mutate`), same shape that lost 10/10 updates on aea4c2a: **0/10 lost**, every trial exactly one `ok` and one `ConflictError`, file content equals the winner's. Your regression test is the same shape, so it stays as the guard.
- Compatibility across processes: two processes with `expected_revision=None` both succeed, last write wins, no error. The "no precondition = today's behaviour" promise holds one layer down too.
- Reader/writer ruamel probe: still 0 errors. Full suite 1577 green here, mypy clean, CI 8/8.
- Adjacent cases: a read-only config directory fails on the lock file with `PermissionError` before anything is written (same failure class as `mkstemp` in that directory today, just a different file name in the message; phases 2-5 map it like any other write error); a crashed holder releases the `flock` with the process; thread lock taken before the file lock at the only call site, so no ordering hazard. Docstring now says what is true, including the POSIX-only requirement stated as a hard requirement rather than a silent degradation, which is the right call for a project whose CI is Linux only.

One non-blocking ask: `.nanoidp-write.lock` stays in the config directory after the first write, and `.gitignore` does not cover it (it does cover `*.bak`, the same family). Nobody hits it in phase 1 since no caller is wired yet, but anyone running the UI from a repo checkout will see an untracked file the moment phase 3 lands. A `.nanoidp-write.lock` line in `.gitignore`, here or in the phase 2/3 PR, whichever you touch first.

Approving. Thanks for the three fast, precise rounds.

dshvedchenko:
feat: ConfigManager.save() migrated onto compare_and_replace (#229 phase 2)

_save_users/_save_settings now route through the shared
compare_and_replace primitive (services/config_writer.py, phase 1)
instead of their own load_yaml_document/atomic_write_yaml calls - same
mutate logic as before (apply_users_document/apply_settings_document),
just behind the shared check-load-mutate-replace cycle. save() gains
optional expected_users_revision/expected_settings_revision
(Optional[str] = None); None keeps today's unconditional last-write-wins
for every caller not yet migrated to pass a real revision (phases 4-5).

save() also now refreshes the runtime from disk once, after both files
are written - closing the gap flagged in the #229 issue thread where an
MCP save_config left the process holding the pre-save view of anything
the read-modify-write cycle picked up from disk. The refresh happens
exactly once, at the end, not after each file: an earlier draft called
reload_local() after _save_users too, and an existing test
(test_profile_override.py) immediately caught the resulting bug - that
reload discarded the in-memory settings.yaml change before
_save_settings ever got to write it. The pre-existing "stops at the
first failed write" contract under hooks.strict is unchanged: a
HookError from either sub-save still propagates immediately, before the
next file or the reload runs (test_hooks.py::test_strict_save_stops_at_the_first_failed_write).

tests/test_config_save_conflict.py adds coverage for the new surface:
matching/stale expected_revision on both files, ConflictError leaving
the file untouched, the unconditional (None) path staying last-write-wins,
and a regression pin for the reload-ordering bug above (a settings change
survives a save() that also touches users.yaml).

No caller passes a real revision yet - routes/ui.py and mcp_server.py
are unchanged; phases 3-5 migrate YamlWriter and wire the UI/MCP through.

cdelmonte-zg (changes requested):
Phase 2 at f776f63, verified: 1583 tests green locally, mypy clean, the `.gitignore` entry is in, the reload-once-after-both-files ordering is right and pinned by a test, and the six new tests cover the surface they name. Two blockers though, one of which is why CI is red.

**Blocking 1: the import contract, which is the CI failure.** The three `test` jobs fail in the "Check import contracts" step (import-linter, so pytest alone does not see it): `nanoidp.config is not allowed to import nanoidp.services: nanoidp.config -> nanoidp.services.config_writer (l.495, l.514)`. The contract is `[tool.importlinter]` in pyproject.toml, `layers: routes -> services -> config`: `config` is the bottom layer and must not import upward; the late import avoids the cycle but not the contract. The primitive is used by both `config.py` (this phase) and `services/yaml_writer.py` (phase 3), so it belongs below both: move it to a root-level module, `nanoidp/config_writer.py`, next to `serialization.py` (it cannot go inside `serialization.py`: that module has its own "no runtime package imports" contract and should stay stateless, and a lock plus `fcntl` is state). `config.py` then imports it at module top like it imports `serialization`, no late import; `services/yaml_writer.py` imports it the same way; the `services/__init__.py` re-export goes. `lint-imports` runs locally too (it is in the dev extras), worth adding to your pre-push checks alongside pytest and mypy.

**Blocking 2: `save()` is not transactional across its two files.** Reproduced: an in-memory change to both users and settings, `save(expected_settings_revision=<stale>)`: `users.yaml` is written (the new user is on disk), then `ConflictError` on `settings.yaml`, which is left untouched, and the in-memory settings keep the unsaved value. The caller (`_tool_save_config` in phase 5) gets "conflict" for a save that half happened. This is the #204 lesson again (reload was made transactional across the directory for the same reason): a precondition that can refuse must be checked for every file before any file is written. The lock already makes it cheap, both files are in the same directory and share it: add a multi-document form of the primitive, e.g. `compare_and_replace_many(directory, [(path, expected_revision, mutate), ...])` (or a small `WriteTransaction` under the same two locks) that takes the locks once, checks every revision, raises `ConflictError` naming the file before touching anything, then applies every mutation and writes every file. `save()` becomes one call to it; a conflict then never leaves a partial save, which is also what the docstring should promise. A test in the shape of the repro above (stale settings revision, users must be untouched afterwards) pins it.

Related, not separately blocking: the `notify_saved` docstring now says "every caller refreshes from local disk before letting that propagate", but `save()` does not on its error paths (a `HookError` from the users write, or the conflict, propagates before `reload_local()` runs). With blocker 2 fixed the conflict path never writes, so the remaining case is a strict hook failure after a write; give `save()` the same post-write contract as `YamlWriter._atomic_write` (write -> notify -> reload_local -> raise), which is what the #229 sign-off asked for, or make the docstring say what `save()` actually does. Either is fine; a claim that is not true is not.

Process note, so it is explicit: my phase 1 approval stands for 9d49ef7 and the phase-per-commit structure is exactly what the #229 sign-off asked for; each new phase just means a fresh review at its head, and the approval that counts is the one at the final head.

dshvedchenko:
fix: config_writer moved out of services/, save() is one transaction (#229 review, blocking x2)

Blocking 1 - the import contract, why CI was red: config.py's late
import of services.config_writer avoided the runtime circular import
but not the [tool.importlinter] "layers: routes -> services -> config"
contract (config is the bottom layer and must not import upward;
lint-imports' static analysis does not care where in a function the
import sits). The primitive is used by both config.py (phase 2) and
services/yaml_writer.py (phase 3), so it belongs below both: moved
src/nanoidp/services/config_writer.py to src/nanoidp/config_writer.py,
a root-level peer of config.py and serialization.py (not inside
serialization.py itself, which has its own "no runtime package
imports" contract and should stay stateless - a lock plus fcntl is
state). config.py now imports it at module top like it imports
serialization, no late import; the services/__init__.py re-export is
gone. lint-imports added to the local verification loop alongside
pytest/ruff/mypy.

Blocking 2 - save() was not transactional across its two files.
Reproduced: an in-memory change to both users and settings,
save(expected_settings_revision=<stale>) wrote users.yaml (new email on
disk), then raised ConflictError on settings.yaml, leaving the in-memory
settings change silently unsaved - the #204 lesson again (reload was
made transactional across the whole directory for the same reason).
config_writer.py adds compare_and_replace_many(items): every item's
expected_revision is checked - against every file's on-disk revision -
before any item is written, and it locks every distinct parent directory
among its items (sorted, so two concurrent multi-file calls can't
deadlock on each other). compare_and_replace is now the one-item case of
the same function. ConfigManager.save() calls it once with both files;
a conflict on either now leaves both untouched.

Since both files are now written together, save()'s hook-then-write
ordering changes too: it notifies each file's hook (collecting either
failure) and refreshes the runtime once, then raises - matching
YamlWriter._atomic_write's write -> notify -> reload_local -> raise
contract, per the #229 sign-off, instead of the old "abort before the
second file is even attempted". test_hooks.py's
test_strict_save_stops_at_the_first_failed_write is renamed
test_strict_save_writes_both_files_before_any_hook_runs and its
assertion flipped to match: both files are on disk by the time the
propagated HookError reaches the caller now, not just the first one.
_save_users/_save_settings stay as standalone single-file methods
(unchanged) for the existing direct-call tests in test_hooks.py; save()
no longer calls them.

tests/test_config_save_conflict.py adds
TestSaveIsTransactionalAcrossBothFiles, the review's exact repro (stale
settings revision, users must be untouched afterwards) - verified it
fails with the two files checked/written independently and passes with
compare_and_replace_many. Full suite (1584 tests), ruff, mypy, and
lint-imports all clean.

cdelmonte-zg (changes requested):
Round 5 at 5b933b3, verified. Both blockers from round 4 are closed:

- B1: `lint-imports` 2 kept / 0 broken locally and CI green on all three Pythons; `config_writer.py` is a root-level peer of `config.py`, imported at module top, no `services/` re-export left anywhere (grep over src/tests/book/docs is clean). Placement reasoning in the module docstring matches what I asked (not into `serialization.py`, which stays stateless).
- B2: reproduced in both directions on this head, my own script, not only your test: stale `settings` revision -> `ConflictError`, `users.yaml` untouched; stale `users` revision -> `ConflictError`, `settings.yaml` untouched; both fresh -> both written and the runtime reflects disk after the single `reload_local()`.
- Cross-process CAS still holds through `compare_and_replace_many` with two files per process (fork + Barrier, same base revisions): 0/5 trials with more or less than one winner. Directory locks sorted, released by the `ExitStack` in reverse.
- Strict-hook path: `hooks.strict` with a hook that always exits 1 -> both files on disk, both hooks invoked (hook log shows `kind=users` then `kind=settings`, each with its own path), runtime refreshed before the raise, then `HookError`. That is the YamlWriter contract, as agreed. (For the record: the "(settings.yaml)" in the error text is the hook's declaring file, `shell_hook.source`, not the file being saved; I checked because it looked wrong at first.)
- Trial merge with current `main` (six commits ahead of your base, including #240/#242/#243): clean, 1624 tests green, contracts kept.
- 1584 tests, mypy, ruff clean on the head itself. The round-4 non-blocking asks (`notify_saved` docstring parity, `.gitignore` entry) are both in.

**Blocking, one item, small:** this round changes a documented contract and the docs still describe the old one.
- `book/src/guides/extending.md` lines 78-85 say "a multi-step save stops at the first failed write: `ConfigManager.save()` writes `users.yaml` then `settings.yaml` ... under `strict` a failing hook leaves the steps after it unsaved". After this commit that sentence is false for `save()`: both files are written as one transaction before either hook runs, both hooks run, and the error is raised after the refresh. The part about the web UI's settings page writing its sections one at a time stays true (that is YamlWriter, phase 3). Please rewrite the `save()` clause to what the code does now; the test docstring you flipped in `test_hooks.py` already says it well.
- CHANGELOG under `[Unreleased]`: no entry yet, and the behaviour is user-visible in two places: MCP `save_config` now refreshes the runtime from disk after writing (the gap from the #229 thread), and under `hooks.strict` a `save()` no longer leaves `settings.yaml` unwritten when the `users.yaml` hook fails. One "Changed" entry covering `save()` as a transaction (precondition on both files, both written, both hooks, one refresh) is enough; the conflict-detection API itself can wait for the phase that wires a caller.

**Non-blocking:**
1. Phase-2 semantics: I probed a `mutate` that raises on item 2 -> item 1 is already on disk and stays there. The docstring says "a conflict on the second item never leaves the first one already written", which is precisely true (conflict = precondition), but a reader will take it as "all or nothing". Cheap tightening: load and run every `mutate` on its document first, then write all; a mutate failure then leaves nothing written and only an I/O failure on write N can half-apply. Or just state the limit in the docstring. Your call, either is fine for me.
2. `book/src/project/architecture.md` (line 85 and 108-114) still says `save()` persists "through `serialization.atomic_write_yaml`" with two read-modify-write owners. Half true now; it becomes fully stale at phase 3 when YamlWriter migrates. Fine to update it there in one go.
3. `cd3cea0` is the bare `git mv` with the same message as `5b933b3`; the squash makes it moot, just noting it so nobody hunts for a difference.

Same process as before: phase 3 onto this PR after this round lands, approval at the final head.

dshvedchenko:
docs+fix: update save() docs/CHANGELOG for the transaction, tighten compare_and_replace_many (#229 review round 4)

Blocking - documented contract now describes reality:
- book/src/guides/extending.md's hook error-policy section said "a
  multi-step save stops at the first failed write: ConfigManager.save()
  writes users.yaml then settings.yaml ... under strict a failing hook
  leaves the steps after it unsaved" - false since the prior commit
  made save() write both files as one transaction. Rewritten to say
  both files are written, both hooks run, and the runtime refreshes
  once before a strict failure is raised; the web UI's settings page
  (YamlWriter, still phase 3) is called out as the one that still
  writes section-by-section and stops at the first failure.
- CHANGELOG [Unreleased] gets a Changed entry: save() as a transaction
  is user-visible in two places - MCP save_config now refreshes the
  runtime from disk after writing, and a hooks.strict failure on
  users.yaml no longer leaves settings.yaml unwritten.

Non-blocking, applied: compare_and_replace_many tightened from
check-all -> (load+mutate+write per item) into three passes -
check-all, then load+mutate every item with nothing written yet, then
write every item. A mutate that raises on item 2 used to still leave
item 1 already written; now every mutate runs before any write, so a
conflict or a raising mutate leaves the whole batch untouched (only an
individual atomic_write_yaml I/O failure can still half-apply the
batch - there's no single filesystem transaction spanning multiple
files, which the docstring now says explicitly).

tests/test_config_writer.py adds TestCompareAndReplaceMany (all-written,
conflict-leaves-both-untouched, and the new raising-mutate case) -
verified the last one fails pre-tightening (item 1 written despite
item 2's mutate raising) and passes with the fix.

Not done, per review's own non-blocking/deferred notes: architecture.md's
description of save()'s write path is left alone until phase 3 (YamlWriter
migration) makes it fully stale in one go, as suggested.

Full suite (1587 tests), lint-imports (2/2 contracts kept), ruff, mypy
all clean.

cdelmonte-zg (changes requested):
Round 7 review, head `0a113e5`.

## The two previous blockers are closed

**B1 (import contract)** - `config_writer` now lives at `src/nanoidp/config_writer.py`, top-level imports, no `services/` leftovers. `lint-imports` locally: 2 contracts kept, 0 broken.

**B2 (non-transactional save)** - verified both directions: a stale `settings.yaml` revision raises `ConflictError` with `users.yaml` byte-identical, and a stale `users.yaml` revision leaves `settings.yaml` byte-identical. The `hooks.strict` path does write both, notify both (hook log shows `users` then `settings`), refresh once, then raise. Two processes racing a two-file `compare_and_replace_many`: 0 lost updates in 5 trials, both files always agreeing on the winner.

The non-blocking tightening is a good change: running every `mutate` before any write is the right ordering, and the docstring is honest about what remains.

Also verified at this head: 1587 tests green, 1627 on a trial merge with current `main` (`8a63d3c`), `mypy` clean, live end-to-end run against a real server 56/57 (the single failure, `Redirect URI Exact Match`, fails on `main` too and is unrelated).

Two new blockers, though, both found on this head.

## Blocking 1 - the package no longer imports on Windows

`config_writer.py` does an unconditional top-level `import fcntl`, and `config.py` imports `config_writer` at top level. The module docstring says so itself: POSIX only, "a Windows deployment would need `msvcrt.locking` behind the same helper". The effect is not a degraded lock on Windows, it is a package that cannot be imported at all.

Measured by blocking `fcntl` at the import-machinery level (a stand-in for a Windows interpreter, where the module simply does not exist):

| | `import nanoidp.config` | `import nanoidp.app` |
|---|---|---|
| `main` (8a63d3c) | OK | OK |
| this PR (0a113e5) | `ModuleNotFoundError: fcntl` | `ModuleNotFoundError: fcntl` |

Today nothing under `src/` imports `fcntl` at all - on `main` the only `fcntl` on any import path comes from the `mcp` SDK's `stdio` module, i.e. the MCP extra, not the core. So this PR is what puts a POSIX-only import on the core HTTP server's path.

That contradicts the declared contract: `pyproject.toml` ships `Operating System :: OS Independent`, and `CONTRIBUTING.md` documents `.venv\Scripts\activate` for Windows contributors. CI is `ubuntu-latest` everywhere, which is why the suite, mypy and `lint-imports` are all green on something that breaks a supported platform.

What I would like instead of a silent degradation: a small cross-process lock backend behind the existing `_cross_process_lock` helper - `fcntl.flock` on POSIX, `msvcrt.locking` on Windows (note that `LK_LOCK` retries only about ten times before raising, so the Windows branch needs its own blocking retry loop) - plus a Windows job in `tests.yml` that at minimum imports the package and runs `tests/test_config_writer.py`. Dropping the lock on Windows and carrying on is worse than either option: it makes the loss of the conflict-detection guarantee invisible exactly where it cannot be tested. The other alternative, declaring nanoidp POSIX-only, is a change of supported platforms much larger than this feature and would need its own decision, not a side effect of a write-pipeline PR.

## Blocking 2 - "transaction" overclaims what the code does

The CHANGELOG entry, `extending.md` and the `ConfigManager.save()` docstring all now say the two files are written "as one transaction". `compare_and_replace_many`'s own docstring says the opposite, and it is the accurate one: "there is no single filesystem transaction spanning multiple files" - an `atomic_write_yaml` I/O failure on the second file can leave the first one already written.

I am not asking for a real multi-file commit protocol, that would be out of proportion here. Just make the three user-facing texts say what the code actually guarantees: something like "one conflict-checked batch" or "one coordinated save", spelling out that a conflict or a failing `mutate` is all-or-nothing, while an I/O failure during the write phase can still leave a partial batch. That is precisely the behaviour, and it is a few lines.

## Non-blocking

- The PR title and body still describe phase 1 only ("No caller wired up yet"), while the branch now contains phase 2 as well. Worth refreshing both before merge, since the squash message comes from them.
- New with the three-pass ordering: two items on the *same* path in one batch now lose the earlier mutation (`[(f, None, x->2), (f, None, y->2)]` leaves `x: 1`, `y: 2`; before the tightening the two mutations composed). No caller does this today, but phase 3 and #192 will add callers - a `ValueError` on duplicate paths, or one line in the docstring, closes it.
- `architecture.md` staying stale until phase 3 is fine, as agreed last round.

cdelmonte-zg:
Addendum to the review above: a second pass over the same head found one more blocking issue, plus a set of smaller notes. Sorry for the split, the findings are real and I would rather send them now than after a fix round.

## Blocking 3 - `save()` can now fail *after* committing, and it swallows the strict hook error

`config.py:521` calls `self.reload_local()` before `raise hook_error`. The reload re-reads and re-validates from disk, so it can raise on its own, and when it does it takes the collected `HookError` with it.

Measured, same probe on both branches (in-memory `token_expiry_minutes = -5`, which `Settings` rejects with `gt=0` at load time but accepts on assignment, since the model has no `validate_assignment`):

| scenario | `main` (8a63d3c) | this PR (0a113e5) |
|---|---|---|
| bad value, no hooks | `save()` returns OK, value on disk | `save()` raises `ValidationError`, value on disk |
| bad value + failing `hooks.strict` hook | `HookError` surfaced correctly | `ValidationError`, the hook failure is lost |

Two separate problems in that table.

The second row is the one I care about most: under `hooks.strict` the operator's contract is "you will be told the mirror push failed". After this PR any error from the reload replaces that message, so the operator is told about the wrong failure and the mirror silently stays behind.

The first row is a durable write reported as a failure. It is reachable end to end through MCP: `update_settings` declares `token_expiry_minutes` as a bare `"type": "integer"` with no minimum (`mcp_server.py:796-799`), so `-5` reaches the in-memory settings, `save_config` writes it, `reload_local()` then raises, and `_tool_save_config` (`mcp_server.py:1502`) answers `{"success": false, "error": "Failed to save config: ..."}` - while the value is already on disk and the next process start refuses to boot on it. That a bad value can reach the file at all is pre-existing and not this PR's fault; what changes here is that a completed write is now announced as a failed one, which is the worse of the two states to be in.

What I would like: give the reload its own `try`/`except`, log the reload failure, keep `hook_error` primacy when both happen, and make a deliberate decision about what `save()` raises when only the reload failed. Whatever you choose, a caller has to be able to distinguish "nothing was written" from "written, but the runtime could not adopt it" - right now both arrive as one exception out of `save()`.

## Non-blocking

- **The read side of the "directory-wide boundary".** `save()`'s docstring claims "the same directory-wide transactional boundary `_load_config` already gives reads (#204), now for this write", but `_stage_directory` opens `settings.yaml` and then `users.yaml` with plain `open()` and no lock, while `compare_and_replace_many` writes `users.yaml` then `settings.yaml`. In exactly the two-process scenario this PR exists for, a `reload_local()` in one process can interleave with a `save()` in the other and pair an old `settings.yaml` with a new `users.yaml`. Same family as blocking 2: either take the lock in `_stage_directory` or narrow the sentence.
- **`fcntl.flock` has no `OSError` handling** (`config_writer.py:153`). On a config directory that does not support advisory locks (NFS without `lockd`, some 9p or FUSE bind mounts) every save now dies with a raw `OSError` on a path that worked before this PR, and the quick start does bind-mount `./config`. At minimum this should fail as a named, explained error rather than a bare `OSError` from the lock helper.
- **No timeout on `LOCK_EX`, taken while holding the process-global thread lock** (`config_writer.py:230`). A peer process stuck inside its own section blocks every writer thread in this process, including ones targeting an unrelated directory, indefinitely and without a log line. `LOCK_EX | LOCK_NB` in a bounded retry loop that raises something named would at least be visible.
- **The cross-process test's failure mode** (`tests/test_config_writer.py:227`): `ctx.Barrier(2).wait()` with no timeout and `p.join(timeout=10)` with no `p.kill()` afterwards means a worker that dies before the barrier leaves its sibling blocked forever and an orphan process behind. `"fork"` also does not exist on Windows and warns on 3.12 when the parent is multi-threaded, which pytest often is. Ties into blocking 1.
- **`notify_saved`'s docstring says "every caller refreshes from local disk before letting that propagate"**, but `_save_users` and `_save_settings` still notify without reloading. After this PR they have no production callers at all, so the eight strict/hook contract tests in `test_hooks.py` now pin a path no surface exercises, while the real `save()` contract rests on the two new tests. Worth either routing those tests through `save()` or saying in the docstring which callers actually honour the contract.
- **`save_config`'s observable behaviour changed** (the runtime is now refreshed by the save, and per blocking 3 a reload failure surfaces as a save failure), but the tool description at `mcp_server.py:879` and `examples/test_agent.py` were not touched. House rule here is that a behaviour change on a surface ships with its MCP text and its end-to-end assertion in the same PR.

dshvedchenko:
fix: distinguish reload-after-save failures, bound the cross-process lock (#229 review, blocking 3 + non-blocking)

Blocking 3: save() called self.reload_local() unconditionally right
before `if hook_error: raise hook_error`. reload_local() can itself
raise (Settings has no validate_assignment, so an in-memory
token_expiry_minutes = -5 reaches settings.yaml fine but fails Pydantic's
gt=0 on reload) - and when it did, that new exception replaced whatever
was about to be raised, in both directions: a hooks.strict failure got
silently swallowed (operator told the wrong thing failed, mirror stays
behind with no signal), and a plain successful write with no hook at all
got reported as a failure via a bare ValidationError, even though the
value was already durably on disk.

config.py adds ReloadAfterSaveError(RuntimeError), same (message, kind)
shape as ConflictError/HookError. save() now wraps reload_local() in its
own try/except, logs the reload failure, and picks one of three mutually
exclusive outcomes in priority order: HookError (mirror push failed,
unchanged priority - "you'll be told the mirror failed" stays the
contract) beats ReloadAfterSaveError (write succeeded, runtime couldn't
adopt it) beats nothing raised. ConflictError (nothing written) is
unaffected - it already returns before either file is touched.

mcp_server.py: save_config's handler catches all three exception types
and returns their .kind in the response, so an MCP caller can tell
"nothing written" from "written, mirror failed" from "written, runtime
desynced" without parsing free text; the tool's description documents
the three kinds and the do-not-retry implication of the last one.

Non-blocking, applied:
- config_writer.py: _cross_process_lock polls LOCK_EX|LOCK_NB with a
  10s bounded wait instead of blocking on a plain LOCK_EX - it's taken
  while holding the process-wide _write_lock, so an indefinite wait on
  one stuck peer used to block every writer thread in this process,
  including ones targeting an unrelated directory, with no log line. A
  timeout, or an OSError that means the filesystem doesn't support
  advisory locks (NFS without lockd, some 9p/FUSE mounts), now raises a
  named LockUnavailableError instead of hanging or leaking a bare OSError.
- tests/test_config_writer.py: the cross-process probe's barrier.wait()
  is now bounded (a dead sibling used to hang the survivor forever), and
  the test kills/reaps any process still alive after its join timeout
  instead of leaking it.
- config.py: notify_saved's docstring now says which callers actually
  reload (save(), once, after both files; YamlWriter per write) versus
  which don't (_save_users/_save_settings called directly - no
  production caller left, only the isolated hook-contract tests in
  test_hooks.py). save()'s docstring also stops overclaiming a
  directory-wide read/write boundary: _stage_directory reads both files
  unlocked, so a concurrent save() in another process can still pair an
  old settings.yaml with a new users.yaml - narrowed to what's actually
  true (write-side only) rather than adding read-side locking here.
- mcp_server.py / tests/test_mcp_tool_branches.py: save_config's tool
  description and three new unit tests (conflict/hook/reload_after_save,
  each asserting the response's kind) are the "same PR as the behaviour
  change" end-to-end coverage - examples/test_agent.py has no MCP stdio
  integration at all (it drives the HTTP surfaces only), so
  test_mcp_tool_branches.py is where MCP tool behaviour actually gets
  end-to-end coverage in this codebase.

tests/test_config_save_conflict.py adds TestSaveReloadFailureAfterASuccessfulWrite,
the reviewer's exact repro (in-memory token_expiry_minutes = -5, with
and without a failing strict hook) - verified both cases fail pre-fix
(silent ValidationError; HookError swallowed) and pass with it.

CHANGELOG's existing "ConfigManager.save() as a transaction" entry
(still unreleased) extended with this round's refinements rather than a
new entry, since it's the same feature maturing before it ships.

Full suite (1592 tests), lint-imports (2/2 contracts kept), ruff, mypy
all clean.

cdelmonte-zg (changes requested):
Round 8 at `cef46b16`. We're very close now.

## Blocking 3 is closed

Verified with the same probe as the report, both rows:

| scenario | `main` (8a63d3c) | `0a113e5` | `cef46b16` |
|---|---|---|---|
| bad in-memory value, no hooks | `save()` returns OK | `ValidationError` | `ReloadAfterSaveError`, named and logged |
| bad value + failing `hooks.strict` hook | `HookError` | `ValidationError`, hook failure lost | `HookError`, priority preserved |

The shape is right: the reload gets its own `try`/`except`, the failure is logged with the directory named, and the three outcomes are mutually exclusive and distinguishable by type. `save()`'s docstring spelling out the priority order is what makes it reviewable. Since `_tool_save_config` is the only production caller of `save()`, the blast radius of the new exception type is fully covered by the handler you added.

Non-blocking items verified as applied: the bounded `LOCK_EX | LOCK_NB` poll with `LockUnavailableError` (both the timeout and the "filesystem has no advisory locks" branch), the barrier timeout plus kill/reap in the cross-process test, `notify_saved`'s docstring, and the read-side narrowing in `save()`'s docstring.

Also at this head: 1592 tests green, `mypy`, `ruff` and `lint-imports` (2/2 contracts) clean, all 8 CI jobs green, and the two-process `compare_and_replace_many` race still 5/5 exactly-one-winner with both files agreeing after the lock rework.

## Blocking 1 (Windows) is unchanged and still blocks the merge

`config_writer.py:132` is still an unconditional top-level `import fcntl`, and `config.py:25` imports `config_writer` at top level. Re-measured at this head, blocking `fcntl` at the import-machinery level:

| | `import nanoidp.config` | `import nanoidp.app` |
|---|---|---|
| `main` (8a63d3c) | OK | OK |
| `cef46b16` | `ModuleNotFoundError` | `ModuleNotFoundError` |

`pyproject.toml:32` ships `Operating System :: OS Independent` and `CONTRIBUTING.md:47` documents `.venv\Scripts\activate`, so this is a regression against a declared supported platform, invisible to CI because every job is `ubuntu-latest`.

I am happy for the fix to land as its own PR rather than growing this branch into a platform port, and I would review that one first and merge it ahead of this. What I am not willing to do is merge this PR with the regression on `main` and only a follow-up issue as the guard, especially with `v2.8.0-rc1` already out and the 2.8.0 milestone otherwise clear: the next release is cut from `main`.

The scope, whichever way you do it:

1. A backend behind the existing `_cross_process_lock` (`config_writer.py:177`): `fcntl.flock` on POSIX, `msvcrt.locking` on Windows. `msvcrt.locking` with `LK_LOCK` retries only about ten times before raising, so the Windows branch needs its own retry loop rather than relying on the mode to wait. The bounded-poll structure and `LockUnavailableError` you just added give it the same timeout semantics and error shape for free.
2. A Windows job in `.github/workflows/tests.yml` that at minimum imports the package and runs `tests/test_config_writer.py`.
3. `tests/test_config_writer.py:237` uses `multiprocessing.get_context("fork")`, which does not exist on Windows and is deprecated on 3.12 when the parent is multi-threaded (pytest often is). That probe needs a `spawn`-compatible form (a top-level picklable worker) or an explicit skip, decided together with 2.

Dropping the lock on Windows and carrying on is the one option to avoid: it makes the loss of the conflict-detection guarantee invisible exactly where it cannot be tested.

## Blocking 2 (the "transaction" terminology) is still open

Please remove the multi-file "transaction" terminology consistently, including `compare_and_replace_many()`'s own docstring, which opens with "across several files as one transaction" (`config_writer.py:284`) and then admits ten lines later that "there is no single filesystem transaction spanning multiple files" (`:294`). A `grep -rn transaction src/ book/ CHANGELOG.md` is the quickest way to catch them all; the write-side ones are:

- `CHANGELOG.md:56`
- `book/src/guides/extending.md:81`
- `src/nanoidp/config.py:338`, `:498`, `:599`
- `src/nanoidp/config_writer.py:42`, `:284`
- `src/nanoidp/mcp_server.py:883`

"Coordinated save", "conflict-checked batch" or "logical save" all describe the actual guarantee better: a conflict or a failing `mutate` is all-or-nothing, and only an I/O failure during the write phase can leave a partial batch.

Out of scope, leave them alone: `config.py:164` and `architecture.md:122` describe `_load_config`'s read-side boundary (#204), which is pre-existing and about a different thing.

The reason this one should not be deferred: that CHANGELOG paragraph becomes a published release note in the next cut, and correcting a released note afterwards is a good deal more annoying than correcting the line now.

## Blocking 4, new at this head: `LockUnavailableError` violates the `save_config` failure-kind contract

Two ways:

1. `_tool_save_config()` does not catch it, so it falls through to the generic exception branch and the response has no `kind` at all.
2. The tool description (`mcp_server.py:879`) groups `lock_...` with `reload_after_save` as a post-write failure, "the write succeeded but the runtime could not adopt it - do not retry expecting a different result". But lock acquisition fails before `compare_and_replace_many()` writes anything.

Reproduced by forcing `_cross_process_lock` to raise:

```
{"success": false, "error": "Failed to save config: Advisory locking is not supported on /tmp/..."}
users.yaml byte-identical: True
settings.yaml byte-identical: True
```

Please catch `LockUnavailableError`, return its `kind`, classify `lock_timeout` / `lock_unsupported` as **nothing written** (and `lock_timeout` specifically as worth retrying, which is the opposite of what the description currently says), and add the corresponding MCP branch test. The three new tests in `tests/test_mcp_tool_branches.py` cover `conflict`, `on_config_saved` and `reload_after_save`; this is the fourth kind the contract now has.

## Before merge

The title reads correctly now. The body is still the phase 1 text and ends with "No caller wired up yet - `ConfigManager.save()` and `YamlWriter` are unchanged", which stopped being true at `f776f63`. Worth refreshing it to describe phases 1-2, since the squash message comes from it.

Still open from last round, non-blocking and your call whether it rides here or with phase 3: two items on the same path in one batch lose the earlier mutation. No caller does this today, but phase 3 and #192 will add callers, so a `ValueError` on duplicate paths or one sentence in the docstring closes it.

dshvedchenko:
fix: Windows lock backend, terminology cleanup, 4th save_config failure kind (#229 review round 8, blocking x3)

Blocking 1 - Windows regression, invisible to CI (every job runs on
ubuntu-latest, but pyproject.toml declares OS Independent and
CONTRIBUTING.md documents a Windows venv): config_writer.py's
unconditional top-level `import fcntl` broke `import nanoidp.config`/
`nanoidp.app` outright with ModuleNotFoundError, since fcntl doesn't
exist on Windows at all.

  - config_writer.py dispatches by sys.platform: fcntl.flock on POSIX,
    msvcrt.locking on Windows, behind two new helpers
    (_try_lock_exclusive/_unlock) that both platforms' bounded-poll
    retry loop in _cross_process_lock calls identically - the
    LockUnavailableError/timeout semantics from the prior round are
    unchanged, just backend-dispatched now.
  - tests/test_config_writer.py's cross-process probe switches from
    multiprocessing's "fork" context (doesn't exist on Windows, warns
    on 3.12+ when the parent is multi-threaded, which pytest often is)
    to "spawn" - verified empirically that the existing module-level
    _mp_worker (already picklable, already using only picklable
    arguments) works unchanged under spawn.
  - .github/workflows/tests.yml adds a windows-latest job: imports the
    package, then runs tests/test_config_writer.py - the minimum guard
    so this regression class can't land invisibly again.

Blocking 2 (reopened) - "transaction" terminology still implied
filesystem-transaction atomicity that compare_and_replace_many's own
docstring immediately contradicted ("no single filesystem transaction
spanning multiple files"). Removed everywhere it described this
feature's write-side guarantee - config.py, config_writer.py,
mcp_server.py, CHANGELOG.md, book/src/guides/extending.md - in favor of
"coordinated, conflict-checked save/batch". Left alone: config.py's
_load_config and architecture.md's description of it, which are a
different, pre-existing thing (#204, the read-side directory
boundary).

Blocking 3 (from last round) reverified closed at this head, no changes
needed.

Blocking 4, new - LockUnavailableError raised before
compare_and_replace_many touches any file, but _tool_save_config didn't
catch it (fell through to the generic branch, response had no 'kind'
at all) and the tool description misclassified 'lock_...' alongside
reload_after_save as "already written, don't retry" - backwards for
lock_timeout, which usually IS worth retrying. mcp_server.py now
catches LockUnavailableError explicitly, returns its kind, and the
description groups lock_timeout/lock_unsupported with conflict as
"nothing was written" while calling out that only lock_timeout is
worth retrying. New test in test_mcp_tool_branches.py covers the
fourth kind alongside the existing three.

Non-blocking, applied: compare_and_replace_many now raises ValueError
on a repeated path in one batch (two items on the same file would each
mutate an independently-loaded copy; whichever writes second would
silently discard the first's mutation) - no caller does this today,
but phase 3 and #192 will add callers.

Also drafted (not pushed - no valid github.com credentials in this
environment): a refreshed PR body describing phases 1-2 and this
round's fixes, replacing the phase-1-only text that still ended with
"No caller wired up yet", false since f776f63; handed to the user to
paste in since the squash message comes from it.

Full suite (1594 tests), lint-imports (2/2 contracts kept), ruff, mypy
all clean on POSIX. Windows path is exercised only by the new CI job -
verified locally to the extent possible without a Windows machine
(syntax, mypy's platform-aware type-checking of the msvcrt branch,
POSIX imports unaffected).

dshvedchenko:
fix: retry os.replace() on a transient Windows PermissionError (Windows CI)

Windows CI (windows-smoke job) failed
TestConcurrentReadsOutsideTheLock::test_readers_outside_the_lock_do_not_corrupt_a_racing_write
with PermissionError: [WinError 5] Access is denied on
os.replace(temp_path, file_path). Not the lock backend (that job's
TestCrossProcessConflictDetection, which actually exercises
msvcrt.locking under contention, passed) - a genuine, pre-existing
Windows limitation of atomic_write_yaml's replace-based write pattern:
unlike POSIX rename(2), Windows can refuse to replace a file that
another handle currently has open even just for reading, since Python's
open() does not request FILE_SHARE_DELETE. This test is the first one
to hammer that window hard enough to reliably hit it - four reader
threads doing tight open/parse/close loops via load_yaml_document()
while 300 writes land - but any concurrent reader-during-write on
Windows could in principle trigger it.

serialization.py adds _replace_with_retry(): os.replace() retried up to
10 times with a 10ms delay on PermissionError specifically, before
raising as before. The window a reader holds the file open is normally
a fraction of a millisecond, so this clears the common case without a
noticeable delay, costs nothing extra on POSIX beyond the one
successful attempt (PermissionError there would mean a real, persistent
permissions problem, which still raises once retries are exhausted),
and doesn't touch atomic_write_yaml's existing backup/restore-on-failure
behavior.

tests/test_atomic_write_retry.py pins it directly (no actual concurrent
reader needed): os.replace mocked to raise PermissionError a few times
then succeed, and mocked to always raise to confirm the exhausted-retry
path still raises RuntimeError and leaves the original file untouched.

Also reverted a speculative fix from the same investigation: initially
padded the lock file to 1 byte on creation, suspecting msvcrt.locking
couldn't lock a byte range in a 0-byte file - the same CI run's
TestCrossProcessConflictDetection passed cleanly against a fresh empty
lock file, disproving that theory, so the padding (unneeded complexity)
was removed rather than kept as defensive code for a scenario that
isn't real.

Full suite (1596 tests), lint-imports, ruff, mypy all clean on POSIX.
Windows verification is the next windows-smoke CI run once pushed - I
can't run the msvcrt/Windows-specific paths locally.

cdelmonte-zg (changes requested):
Round 9 at `aae63071`. The three blockers from Round 8 are closed.

- **B1, Windows**: the package imports again, the `fcntl`/`msvcrt` dispatch sits behind `_try_lock_exclusive`/`_unlock` with one shared bounded-poll loop, and the new backend is genuinely exercised rather than merely compiled - `TestCrossProcessConflictDetection` passes on the actual windows-latest runner, and the probe's move to `spawn` is right for the reason its docstring gives.
- **B2, terminology**: the write-side filesystem-transaction overclaim is gone from `config.py`, `config_writer.py`, `mcp_server.py`, the CHANGELOG and `extending.md`. What is left is either explicitly negative ("not a filesystem transaction") or the pre-existing #204 read-side boundary, which was out of scope.
- **B4, MCP failure kinds**: `_tool_save_config` catches `LockUnavailableError` and returns its `kind`, and the description groups `lock_timeout`/`lock_unsupported` with `conflict` as "nothing was written" while keeping only `lock_timeout` retryable. Verified live, not only in the unit test: with a foreign `flock` holding `.nanoidp-write.lock`, `save()` raises `LockUnavailableError(kind="lock_timeout")` after 10.1s and logs the wait once.

POSIX is green at this head here as well: 1596 tests, mypy, ruff, lint-imports 2/2, and the live agent suite 57/57 against a running instance.

One blocking item is left, and it is the one your own new Windows job surfaced.

## Blocking: `load_yaml_document()` holds the target open for the whole YAML parse

`windows-smoke` is red only in `TestConcurrentReadsOutsideTheLock` - 16 of 17 config_writer tests pass on the Windows runner, the cross-process `msvcrt.locking` one among them. The failure is `PermissionError: [WinError 5]` out of `os.replace()`.

So this is not the new lock backend failing. The Windows sharing limitation itself is pre-existing: `atomic_write_yaml` calls `os.replace()` on `main` too. What is new is that #234 states an explicit reader/writer safety guarantee (the test's own words: it "has to hold at the loader/dumper level") and adds a job that tests it continuously. This PR is the first change that exposes the limitation and keeps it exposed, so closing #234 red on exactly the test that materialises its own contract is not an option I want to take.

The retry in `aae63071` is the right instinct for a genuinely transient sharing overlap, but it cannot clear this one, and the docstring's reasoning is where it goes wrong. It puts the window at a fraction of a millisecond because "a reader opens, parses and closes" - the parse is inside the window, and it dominates it:

```python
with open(file_path, "r") as f:
    return _new_yaml_rt().load(f) or CommentedMap()
```

Measured here on one of these documents: 334 us per read with the handle open across the parse, of which the read itself is 9 us, 2.6 percent. The probe runs four readers in a continuous loop, so occupancy of the target handle is effectively total (measured mean: 4.00 readers inside `load_yaml_document` at any instant), and ten attempts spread over 100 ms all land inside somebody's parse. Tuning the attempt count or the delay only moves that around.

The fix is to stop coupling the two steps: acquire the filesystem snapshot, close the handle, then parse from memory.

```python
with open(file_path, "r") as f:
    content = f.read()

return _new_yaml_rt().load(StringIO(content)) or CommentedMap()
```

`StringIO` rather than handing ruamel the string directly, to keep the parser's stream contract explicit - but the point is not `StringIO`, it is closing the OS handle before parsing. On the same measurement that takes the sharing window from 334 us to 9 us at no throughput cost (324 us versus 334 us per read), which is what finally makes your bounded replace retry meaningful on Windows: 10 ms between attempts against a 9 us window. Keep the retry, it is still correct for the residual case.

`TestConcurrentReadsOutsideTheLock` plus `windows-smoke` are already the regression test I want here. No new synthetic test needed.

Once that job is green, I do not see another code blocker on this head.

**Before merge**: the PR body is still the phase-1 description. It points at `services/config_writer.py`, which this PR moved out of `services/`, and it ends with "No caller wired up yet - `ConfigManager.save()` and `YamlWriter` are unchanged", which stopped being true at phase 2. Please replace it with the phase 1 + 2 body before the squash merge, since that text lands as the commit message on `main`.

cdelmonte-zg:
Correction to my previous review, after reproducing the Windows sharing semantics locally: **close-before-parse is the right improvement, but it is not sufficient by itself to make this stress probe deterministic.**

I simulated Windows' no-`FILE_SHARE_DELETE` behaviour by denying `os.replace()` whenever a reader handle on the target was live, while otherwise running the real `compare_and_replace` path with the same 4-reader / 300-write probe.

At `aae63071` the simulation reproduces the CI failure heavily: 286 of 300 writes failed in one run, with 2943 denied replace attempts, i.e. the retry budget exhausted almost every time. Reading the file into memory, closing it, then parsing from `StringIO` improves that by roughly two orders of magnitude, down to 1 to 4 failed writes per 300 across repeated runs. I still want that change: 1596 tests, mypy and ruff are clean with it here (it needs `from io import StringIO` placed in isort order, otherwise ruff flags I001).

But under four contending reader threads the target handle is still live about 58% of wall time, because a thread can be descheduled between `open()` and `close()` - the window is no longer the 9 us of actual reading I measured single-threaded. That figure matches the measured replace-denial rate almost exactly (58.1% occupancy, 414 denials in 712 attempts), so the remaining tail with ten retries is expected arithmetic rather than a mistuned retry. Raising the retry budget would only make the test probabilistically greener, which I do not want.

More importantly, I checked the property this regression test actually exists to pin: **`reader_errors == 0` in every configuration, including the runs where hundreds of writes failed cleanly.** The readers never observe a corrupt or partial YAML document. On Windows the writer can instead fail with `PermissionError`, leaving the previous file intact.

So I am narrowing the blocker:

1. Keep the close-before-parse change in `load_yaml_document()`. It materially reduces the Windows sharing window.
2. Make `TestConcurrentReadsOutsideTheLock` assert the contract it names on Windows: readers must never observe corruption, while a clean writer failure caused by Windows sharing semantics may be tolerated. On POSIX, keep the current stronger expectation that every write succeeds.
3. Document the limitation next to `atomic_write_yaml` and its retry: readers opened through Python's normal `open()` do not request `FILE_SHARE_DELETE`, so sustained concurrent readers can still make `os.replace()` fail after the bounded retries, and that failure leaves the old file intact.

One qualification on point 2: please do not turn the Windows branch into a blanket `except Exception: pass`. It should tolerate only the expected writer failure arising from a sharing violation (`PermissionError`, as wrapped by `atomic_write_yaml`) and still fail on anything else. Ideally the test counts writer failures, asserts that each one left a valid document behind (the previous revision, not a partial file), and keeps `reader_errors == []` categorically.

That gives a criterion that is a real specification rather than a tolerance:

- **POSIX**: 300 writes succeed, and zero reader corruption.
- **Windows**: zero reader corruption, and every write is either successful or a clean sharing failure with the previous state intact.

Full POSIX-like rename parity is achievable by opening readers on Windows through `CreateFileW(..., FILE_SHARE_DELETE, ...)` and wrapping the handle, but I do **not** require it in this PR. That is a platform-specific I/O layer substantially larger than the contract #234 needs, and it can be a later improvement rather than the condition for closing this one.

This corrects my previous statement that close-before-parse alone would make `windows-smoke` green. The measurement says otherwise, and the narrower contract above is the one I can defend.

cdelmonte-zg:
Concretely, in case it saves you a round. `load_yaml_document()`:

```python
from io import StringIO

def load_yaml_document(file_path: Path) -> Dict[str, Any]:
    if not file_path.exists():
        return CommentedMap()

    with open(file_path, "r") as f:
        content = f.read()

    return _new_yaml_rt().load(StringIO(content)) or CommentedMap()
```

And the shape I mean for the probe:

```python
writer_errors = []

try:
    for i in range(300):
        try:
            compare_and_replace(
                f,
                None,
                lambda doc, i=i: doc.update({"counter": i}),
            )
        except RuntimeError as exc:
            if sys.platform != "win32":
                raise
            if not isinstance(exc.__cause__, PermissionError):
                raise
            writer_errors.append(exc)
finally:
    stop.set()
    for t in readers:
        t.join()

assert errors == []

if sys.platform != "win32":
    assert writer_errors == []
```

Sketch only - please adapt it to the existing test style; the important part is that Windows tolerates only the known `PermissionError`-backed clean replace failure, not arbitrary writer exceptions. `atomic_write_yaml` already gives you that shape, since it re-raises as `RuntimeError(...) from e`. Note `tests/test_config_writer.py` does not import `sys` today.

Both snippets applied together are green here on Linux: 17/17 in `tests/test_config_writer.py`, full suite 1596, ruff and mypy clean.

dshvedchenko:
fix: close read handle before parsing, tolerate the known clean Windows replace failure (#229 review round 9)

Root cause (blocking): load_yaml_document held its file handle open for
the whole ruamel parse, not just the read. Round 8's os.replace() retry
assumed the sharing window was "a fraction of a millisecond - a reader
opens, parses and closes", which was wrong: the parse dominates it
(334us measured with the handle held across the parse, vs 9us for the
read alone). windows-smoke's four-reader probe keeps the target
occupied effectively continuously, so ten retries over 100ms all landed
inside somebody's parse - no amount of retry tuning could have fixed
that, only the coupling itself.

load_yaml_document now reads the file into a string and closes the
handle before handing it to ruamel (via StringIO), rather than
streaming the parse from an open file object. Shrinks the Windows
sharing window by ~97% (334us -> 9us) at no throughput cost, which is
what actually makes _replace_with_retry's bounded wait sufficient - a
10ms retry gap against a 9us window, not a 334us one. Both docstrings
corrected to say why, not just what.

TestConcurrentReadsOutsideTheLock (already the regression test for
this, no new test needed) now also tolerates exactly one known, clean
Windows failure shape instead of either ignoring all writer errors or
failing on an expected edge case: an atomic_write_yaml RuntimeError
wrapping a PermissionError means the bounded retry was exhausted under
worst-case CI scheduling but nothing was corrupted - allowed on Windows
only, still a hard failure on POSIX (where it should never happen) or
for any other exception shape.

Full suite (1596 tests), lint-imports, ruff, mypy all clean on POSIX -
matches the reviewer's own local numbers (17/17 in
test_config_writer.py, 1596 total) with both pieces applied together.
Windows verification is the next windows-smoke run once pushed.

-----


The PR body still needs the phase 1+2 refresh before squash-merge (flagged again last round, and I still can't push it myself — draft from my previous message stands).

cdelmonte-zg:
Verified `7d607d3`. The round-9 changes themselves are right: close-before-parse is in, the Windows writer exception is narrowed exactly to a `RuntimeError` caused by `PermissionError`, and the POSIX jobs are green (1596 tests, mypy, ruff, lint-imports here as well).

`windows-smoke` is still red, but the new failure locates the remaining issue much more precisely: the writer failures are tolerated as intended, and instead **seven readers fail with `PermissionError`**, so `errors == []` is correctly still catching a violation of the reader-side contract.

I believe the culprit is `atomic_write_yaml`'s recovery path, not the replace itself. What points there rather than at the replace: the job logs 8 `Restored ... from backup after write failure` lines against 7 reader errors, close to one-to-one. Had the readers been colliding with the successful replaces, the errors would be spread over the roughly 292 writes that did land, not tracking the 8 that failed.

When `_replace_with_retry()` exhausts on Windows, the failed `os.replace()` has not modified the target: the old file is already intact, and there is nothing to recover. The handler nevertheless does:

```python
shutil.copy2(backup_path, file_path)
```

which turns a clean failed atomic replace into an in-place, non-atomic overwrite of the live target while four readers are racing it.

I checked what that branch does independent of Windows, by forcing `os.replace()` to fail on Linux and running the same four-reader probe against a 21 KB settings.yaml. It is worse than the Windows symptom:

- **43 of 983 reads returned a truncated document** (fewer clients than the file contains) **and raised nothing at all**. On POSIX this branch does not announce itself, it silently hands a reader a partial config.
- With the restore guarded so it only runs when the target actually disappeared, the same probe returns 0 truncated reads, and the suite stays at 1596.

So this is the one place in the module where a reader can observe a partial document, which is exactly what the rest of the design goes to some length to prevent. On Windows it surfaced as `PermissionError`; on POSIX it would have gone unnoticed, since the probe only counts exceptions.

Please don't loosen the reader assertion. Instead, on a failed temp-write or replace, remove the temp and leave the existing target alone; the `.bak` can stay as a backup. If you want defensive automatic restoration, do it only when the target actually disappeared:

```python
if not file_path.exists() and backup_path.exists():
    shutil.copy2(backup_path, file_path)
```

Also please soften the new docstring's claim that close-before-parse makes the bounded retry "actually sufficient". This run shows it can still exhaust under sustained Windows readers, which is why the test tolerates that shape. Something like "materially reduces the sharing window and makes the bounded retry effective in the common case; sustained readers can still exhaust it on Windows" is the guarantee we can support.

One last thing on the commit itself: `7d607d3`'s message is a note about the PR body rather than a description of the change it carries, so it will read oddly in the squash. Please reword it, and the PR body still needs the phase 1 + 2 refresh that has been pending for a couple of rounds.

Once `windows-smoke` is green with the reader assertion intact, I consider the code side done.

dshvedchenko:
fix: stop the backup restore from overwriting a live, untouched file (#229 review round 10, blocking)

atomic_write_yaml's failure handler restored from the .bak backup via
shutil.copy2 on ANY exception, including a failed os.replace() that
never touched the target at all - os.replace() is atomic, it either
lands or it doesn't, so there was nothing to restore in that case. That
unconditional restore performed a non-atomic, in-place overwrite of a
live file that was already correct. A concurrent reader can observe
that copy mid-flight: this is what surfaced as PermissionError on
Windows CI (a reader's open handle colliding with the extra copy2, not
with the original replace), and reproduced worse on POSIX by forcing
os.replace() to fail under the same four-reader probe - 43 of 983 reads
came back silently truncated (fewer clients than the file has), no
exception raised at all. Exactly the partial-read this module exists to
prevent, and it went undetected on POSIX because nothing before this
review counted content, only exceptions.

Fixed by restoring from backup only when the target actually
disappeared (`if not file_path.exists() and backup_path.exists()`) -
a target that never changed does not need restoring.

Also softened two docstring claims (load_yaml_document,
_replace_with_retry) that close-before-parse made the bounded replace
retry "actually sufficient" - this round's own reproduction shows
sustained readers can still exhaust it on Windows, which is why the
regression test tolerates that one specific shape rather than asserting
it can never happen. Corrected to "makes the retry effective in the
common case; sustained readers can still exhaust it on Windows."

tests/test_atomic_write_retry.py adds
TestFailedReplaceDoesNotCorruptTheLiveFile, adapting the review's own
reproduction (large settings.yaml, os.replace() forced to fail, four
concurrent readers counting clients) - verified it fails against the
unconditional restore (every failed write logs an unnecessary "Restored
... from backup") and passes with the fix, asserting zero truncated
reads and that the live file stays byte-identical to the original
across 50 failed write attempts.

Full suite (1597 tests), lint-imports, ruff, mypy all clean on POSIX.
Windows verification is the next windows-smoke run once pushed.


PR body refresh — the draft I gave two rounds ago covers phases 1–2 and the Windows/reader-writer work accurately through round 9; this round's fix (the backup-restore corruption bug) is a small addition to that story. Want me to fold it into an updated draft, or is the existing one close enough to paste as-is?

cdelmonte-zg:
Correction after the `9b955da9` Windows run: **my round-10 causal diagnosis was wrong.**

The recovery-path bug was real and the fix should stay - the new regression test correctly pins the POSIX partial-read case, and I checked it is a genuine pin: putting the unconditional `copy2` back makes `TestFailedReplaceDoesNotCorruptTheLiveFile` fail, and the guard makes it pass. But that bug was not the source of the Windows reader failures.

This run still reports 8 reader-side `PermissionError(13, "Permission denied")`, while there are now **zero** `Restored ... from backup after write failure` lines in the log. That cleanly rules out the unconditional `copy2` restore as the cause of this symptom. Everything else at this head is green here too: 1597 tests, mypy, ruff, lint-imports 2/2.

What remains is a transient Windows open failure around the replacement itself. CPython implements `os.replace()` on Windows with `MoveFileExW(..., MOVEFILE_REPLACE_EXISTING)`, and this failure shape is consistent with Windows' file-name and sharing semantics during a replacement; there is a long-standing CPython issue describing the same shape, with retry-with-delay as the reported mitigation. I don't think we need to identify the exact NT state, and I would rather we didn't claim one in a docstring. The observable contract is enough: a concurrent config read can transiently get `EACCES` while a save is replacing that file.

So the last code blocker is in `load_yaml_document()`: make acquiring the input bytes resilient to that transient condition with a short bounded retry.

One detail matters more than it looks. Please don't keep the current `exists()` then `open()` split, and don't wrap only the `open()` in the retry. `Path.exists()` re-raises `PermissionError`: `pathlib._ignore_error` swallows only `ENOENT`, `ENOTDIR`, `EBADF` and `ELOOP`, so `EACCES` propagates out of the pre-check itself. I verified that directly (`Path.exists()` on a file whose directory denies `stat` raises `PermissionError(13)`, it does not return `False`). A retry placed around `open()` alone would therefore miss these failures entirely, possibly all of them, since `exists()` is the first thing that touches the path. `FileNotFoundError` is the sibling race, and there a transiently false `exists()` would quietly return an empty `CommentedMap` instead of retrying. Opening directly inside the bounded loop removes both.

Semantics I would preserve:

- on Windows, retry a transient `PermissionError`;
- also retry `FileNotFoundError`, but if it is still missing after the bounded budget, preserve today's contract and return an empty `CommentedMap`;
- a persistent `PermissionError` still propagates after the budget;
- POSIX behaviour stays unchanged;
- only acquiring the bytes is retried, never YAML parsing - a ruamel error must not be retried.

Please add deterministic unit coverage for transient and persistent `PermissionError` and `FileNotFoundError`; the four-reader Windows probe stays the end-to-end guard, but on its own it is statistical.

To be explicit about where this stops, since the last few rounds have each uncovered the next layer down (shared parser, thread race, process race, platform import, writer replace contention, recovery path, and now reader open contention): **the contract I am reviewing against is `windows-smoke` green with `errors == []` intact, plus the deterministic retry tests. When that passes, the code side is done and I am not looking for more.** If the bounded read retry turns out not to converge, I do not want another tuning round: the alternative is opening readers on Windows with `FILE_SHARE_DELETE`, and that becomes a separate issue rather than more work on this PR.

The round-10 recovery fix is independently valid, as above, so none of that work should be backed out. Note though that `9b955da9`'s commit message still says the `copy2` restore is "what surfaced as PermissionError on Windows CI" - this run shows that is not true. Not a blocker on its own, but it needs correcting along with the PR body refresh that is already pending before the squash.

cdelmonte-zg:
Please hold off on my previous comment: **I am withdrawing the request for a bounded retry in `load_yaml_document()`.** Don't implement it. I opened #246 instead, and this PR should close without it.

Two reasons.

The first is that it would have been a fix in the wrong place. I enumerated who actually reads these files, and `load_yaml_document` is not the main read path: `_stage_directory()` reads both `settings.yaml` and `users.yaml` with plain `open()` plus `yaml.safe_load` (`config.py:242` and `:269`), and `config_validation.py:69`, `config_validation.py:202` and `hooks.py:631` read them directly too. Only `services/yaml_writer.py:67,71` go through `load_yaml_document`. Since `reload_local()` goes through `_stage_directory()`, the most frequent read-during-write in the system - the reload `save()` itself triggers right after writing - would not have been covered. The retry would have turned the probe green while leaving the reader that matters most exposed. That is a good demonstration that the symptom is not where the problem is.

The second is a process point, and it is mine to own rather than yours. #234 achieved what it was opened for several rounds ago: one write path, revisions and preconditions, no lost updates across processes, coordinated writers, the ruamel fix, `ConfigManager` migrated, error semantics on the MCP surface. What we have been circling since is a different question - whether readers can stay outside the coordination protocol at all - and that is a decision about where the configuration's consistency boundary lies, not a correction to the write primitive. Turning each new finding from the stress probe into another blocker on the same PR is how a PR never ends. That work belongs in #246.

So the plan I would like to close on:

1. One last small commit here that does not pretend the problem is solved: keep `windows-smoke`, mark only `TestConcurrentReadsOutsideTheLock` as `xfail` on Windows with a pointer to #246, keep it active and required on POSIX, and note the Windows limitation where a reader would trip over it.

```python
@pytest.mark.xfail(
    sys.platform == "win32",
    reason="Concurrent readers are not yet part of the config-file "
           "coordination protocol; see #246",
)
```

Please leave it non-strict: if a runner passes for probabilistic reasons we don't want an XPASS turning the suite red. Making that reproduction deterministic is part of #246, not of this PR.

I prefer `xfail` to deleting the test or loosening `errors == []` because it keeps the information intact: this is the property we want, Windows does not satisfy it today, and #246 is the explicit debt.

2. Also in that commit, the two pieces of stale prose: `9b955da9`'s message still says the `copy2` restore is "what surfaced as PermissionError on Windows CI", which the zero-restore run disproved, and the PR body is still the phase-1 text that names `services/config_writer.py` and ends with "No caller wired up yet". Both land in the squash, so they need to be right before it.

3. When CI is green with that, I approve and merge.

Everything else from these rounds stays. The `copy2` recovery bug in particular was a real find - a failed replace never touched the target, so restoring from `.bak` turned a clean no-op failure into a non-atomic in-place overwrite that could hand a reader a silently truncated document, which I reproduced on POSIX at 43 of 983 reads. Your regression test pins it properly: I checked that putting the unconditional restore back makes it fail. That bug would have shipped in 2.8.0 unnoticed if this PR had not added a Windows job, which is a decent argument for the job on its own.

#234 is substantially better for everything this review turned up. The read boundary is genuinely new work, and it should be new work.

dshvedchenko:
I will address it today or tomorrow, till evening no new work will be done

dshvedchenko:
test: xfail the reader-coordination gap on Windows only, defer to #246

TestConcurrentReadsOutsideTheLock's errors == [] assertion pins a
property that #229's write primitive cannot deliver on its own:
load_yaml_document is only one of several direct readers of
settings.yaml/users.yaml (_stage_directory, config_validation.py,
hooks.py all read these files on their own, outside compare_and_replace
entirely), so a fix scoped to this one function - a retry, a
handle-close - cannot make "no reader ever sees an error during a
write" true system-wide. Whether readers join the write-coordination
protocol at all is a real design question, not a bug in this PR's
write path; withdrawn as a fix here and opened as #246 instead.

Marked xfail(sys.platform == "win32", ..., see #246) rather than
loosened, deleted, or given more special-case tolerance: the property
is real and wanted, Windows doesn't satisfy it yet, and #246 is where
that gets decided and (eventually) made deterministic. Left non-strict
on purpose - a probabilistic pass on some runner must not turn the
suite red as an XPASS. Unaffected on POSIX: the test runs and is
required exactly as before.

Full suite (1597 tests), lint-imports, ruff, mypy all clean on POSIX.

dshvedchenko:
if it is ok ,lets 
1. squash commits on this pr
2. move to phase 3

cdelmonte-zg (approved):
Round 11 at `2c76bbe`, and this closes it. Approved.

The `xfail` is exactly the shape I asked for and, more to the point, it is honest: the property stays written down, the test stays active and required on POSIX, and the reason names #246 rather than pretending the gap is fixed. I checked it is genuinely non-strict, since there is no `xfail_strict` anywhere in `pyproject.toml`, so a Windows runner that passes for probabilistic reasons will not turn the suite red on an XPASS. Removing the writer-side `RuntimeError`/`PermissionError` tolerance along with it is right too: with the whole test deferred on Windows, that special case had nothing left to buy, and its absence means no writer exception is quietly accepted on POSIX.

Verified here at this head:

- full suite **1597 passed**, `tests/test_config_writer.py` **17 passed** with `test_readers_outside_the_lock_do_not_corrupt_a_racing_write` running for real (not skipped, not xfailed) on Linux
- mypy clean on 35 source files, ruff clean, `lint-imports` 2 contracts kept / 0 broken
- CI green on all of 3.10, 3.11, 3.12, http-e2e, mcp-e2e, license-check, CodeQL and `windows-smoke`

I have refreshed the PR description, which was still the phase-1 text (it named `services/config_writer.py`, described the lock as process-local only, and ended with "No caller wired up yet"). Nothing for you to do there. I will compose the squash commit message by hand at merge, so no local rebase or squash is needed on your side either, and round 10's superseded causal note about the `copy2` restore will not travel into `main`.

For the record of what this PR ended up being worth beyond its own scope: the `copy2` recovery bug was a real, shipping-in-2.8.0 defect that only surfaced because you added a Windows job, and the shared-ruamel-instance finding was another. Eleven rounds is a lot, and most of them were mine, but the write path that comes out of it is the one the rest of #229 can build on.

Merging now. Phase 3 (`YamlWriter` onto the primitive) is yours whenever you want it.

Issue #244, opened 2026-08-26
feat: Description for personas login


gprossliner:
After testing #158 I got feedback from some users. Generally it's very appreciated, but it would be a nice extension to have a `description` / `persona_description` free text field that is displayed in the personal login UI. This can be used for example to write role assignments / given claims or other information.

I'd say in small fonts beside the name, or as a tooltip on hoover. We should limit this in size to e.g. 200 chars.

What do you think of this addition?

cdelmonte-zg:
Accepted, and thanks for bringing back real feedback from people who actually used persona mode - that is the kind of request I want to act on.

It is also a good fit rather than feature creep: persona mode exists to pick a test identity fast, and once a directory has `admin`, `reader`, `tenant-a-manager` and friends, knowing what each one stands for *before* the click is the whole point of the picker.

Here is the scope I would ask for.

**A first-class field, not an attribute.** `User.description: str = Field(default="", max_length=200)` in `models.py`, plus the matching key on `UserEntry` in `config_documents.py` so it round-trips through `users.yaml` as a known field. Explicitly **not** `attributes["persona_description"]`: custom attributes are part of the identity surface here, not a display area. `create_token()` puts the whole `attributes` map into the token (`extra["attributes"] = user.attributes`), `resolve_user_claim()` lets any attribute be requested as an OIDC claim, and an attribute whose name matches a configured `authority_prefixes` entry even turns into an authority. A note like "Admin persona, groups=finance, role=approver" must not be able to end up as token material by accident.

**Display-only semantics**, stated in the field description and in the docs: it is never a claim, never exported over SAML, never in a token.

**Rendered inline, not tooltip-only.** `small text-muted` under the username inside the picker. A hover tooltip is worse on touch and with a keyboard, and it hides exactly the information that drives the choice. Nothing stops us adding `title=` as well, but the visible line is the deliverable.

**One contract change covers every surface.** `persona_picker(users, ...)` in `templates/_macros.html` currently receives `list(config.users.keys())` and prints the bare name. It has four call sites, not three: `routes/ui.py:80` (`/ui/login`), `routes/oauth.py:439` (`/authorize`), `routes/oauth.py:1854` (device flow), and `routes/saml.py:515` (the inline login shown during SAML SSO, which renders the same `login.html`). Change what the macro is handed once - the `User` objects, or `(username, description)` pairs - and all four surfaces get it together. Please keep all four in the same PR.

[...]

Issue #37, opened 2026-05-25
Minor polish from #32 review: validate_assignment, advertise azp, normalize strictness


cdelmonte-zg:
Low-priority follow-ups gathered from the code review of #32.

### 1. Enable `validate_assignment` on `OAuthClient`
`OAuthClient` has no `model_config`, so direct attribute assignment bypasses Pydantic validation. MCP `update_client` already does `client.additional_audiences = ...` and `client.client_secret = ...` directly; the `min_length=1` on `client_secret`/`client_id` is unenforced on assignment. `_normalize_audiences` covers the audiences case today, but enabling `validate_assignment=True` would close the gap generally.
Ref: `src/nanoidp/config.py` (`OAuthClient`), `src/nanoidp/mcp_server.py` (`update_client`).

### 2. Advertise `azp` in discovery `claims_supported`
ID Tokens now emit `azp` for multi-audience clients, but `/.well-known/openid-configuration` does not list `azp` in `claims_supported`. Strict RPs that only trust advertised claims will ignore it.
Ref: `src/nanoidp/routes/oauth.py` (~line 51).

### 3. `_normalize_audiences` accepts falsy non-list silently
`if not value: return []` converts any falsy input (`0`, `""`, `False`) to `[]` rather than rejecting a non-list, inconsistent with the truthy-non-list branch that raises. Acceptable as "clear" semantics, but tightening it would make the contract uniform.
Ref: `src/nanoidp/mcp_server.py` (`_normalize_audiences`).

cdelmonte-zg:
Addressed in #38 (branch `fix/issue-32-followups`).

PR #247, opened 2026-08-27
feat: personas description

gprossliner:
see #244 

* As before the temporary planing doc is in docs/plan/persona-descriptions.md
* I visually inspected the UI, we use the CSS classes you mentioned, and I think it really looks good!

Please review


cdelmonte-zg (changes requested):
Thanks, this is a clean implementation of the scope we agreed on in #244, and the cross-surface part is genuinely complete: one `persona_picker_entries()` feeding all four interactive surfaces, the write surface (UI form, detail page, REST API, MCP tools), YAML round-trip, and the live e2e agent.

**What I verified on `51428dd`**, beyond CI:

- Full unit suite, `ruff`, `mypy`, `lint-imports`: green (1627 passed).
- Live server from a fresh `init`: `examples/test_agent.py` 57/57.
- The description renders on all four pickers against a running server (`/login`, `/authorize`, `/saml/sso`, `/device`), and is omitted cleanly when blank.
- Display-only holds where it matters: not in the access token claims, not in `/userinfo`, not in the SAML assertion, and `resolve_user_claim()` is a closed whitelist plus `user.attributes`, so it can never be requested as a claim either.
- Escaping: a description containing `<script>` and quotes renders escaped in the picker.
- YAML round-trip is exact for a nasty value (newlines, quotes, `#`, backslash, non-ASCII).
- Overlong description rejected on create and on edit through the UI, with no partial write; `validate-config` reports an overlong value in `users.yaml` as an error and exits 1.

Two blocking items, plus one cleanup request that is not a blocker.

---

## Blocking 1: `tests/test_token_service.py` lost the `TestGlobalTokenService` class header

The new `test_description_is_never_a_token_claim` was appended right where the class statement was, so `class TestGlobalTokenService:` is gone. What is left is:

```python
        assert all("Finance approver persona" not in str(v) for v in decoded.values())

    """Tests for the global token service singleton."""

    def test_get_token_service_returns_instance(self, app):
```

The docstring is now a stray string expression inside `TestTokenInclusion`, and `test_get_token_service_returns_instance` has silently become a method of that class. Nothing fails today (no name collision, the test still runs), which is exactly why it needs to be caught here. Please restore the class.

## Blocking 2: a direct mutation can bypass `description`'s 200 character limit

`_tool_update_user()` assigns `user.description = arguments["description"]` directly. `User.model_config` is `ConfigDict(extra="allow")` with no `validate_assignment`, so the `max_length=200` on the field is enforced at construction only. This is the same defect `OAuthClient` fixed in #37, and the comment there states the rule: "Validate on direct attribute assignment too (e.g. MCP update_client), so the field constraints below are enforced beyond construction time".

Reproduced locally: calling the handler directly with a 300-character description returns `success: True`, `config.save()` writes it to `users.yaml`, and the next `ConfigManager(...)` load raises `ValidationError` on `User.description`. That is a config the server can no longer start from.

The MCP protocol path is not affected (the `maxLength` in the schema rejects it before dispatch, and your test proves that), so this is defense in depth, but it is the house rule here: see also the `_normalize_str_list` guard in `_tool_generate_token`, whose comment says `_execute_tool` is reachable directly and must reject bad input itself. `description` is the first `User` field where a bypass poisons the config file rather than just producing an odd value.

The requirement is that a direct mutation cannot persist a value the loader will reject. How you get there is your call. `ConfigDict(extra="allow", validate_assignment=True)` on `User` fixes the reproduced case and keeps the full suite green here (1627 passed, repro rejected), and it matches the `OAuthClient` precedent; an equivalent guard at the mutation boundary is fine too. Either way, please cover the assignment path with a test so the constraint stays enforced.

## Before merge (cleanup, not a blocker): please drop `docs/plan/persona-descriptions.md`

Your own PR description calls it a temporary planning doc, so this is really just agreeing with you. It was useful while the work was in flight, but as a committed file it is a tracking artifact with struck-through checklist items living permanently next to the real docs. The outcome is already captured in `book/src/reference/configuration.md`, the security guide, the example, and the CHANGELOG entry you proposed. It also carries 13 em dashes, and the repository has none anywhere else, which is a convention I keep. If you would rather keep a plan doc in the repo, say so and we can agree on a place and a format for it, but not in this PR.

---

On the CHANGELOG: I will take the entry, thanks for drafting it. I will add the compatibility line we discussed in #244, since it is a real behavior change: a `description:` key already present in `users.yaml` today is folded into `attributes` (confirmed on `main`: it comes back as `attributes = {'description': ...}`), and therefore ships inside the token's `attributes` claim. After this PR it stops appearing in tokens.

One optional item, my call rather than yours: `UserEntry.description` has no `max_length`, so the generated `docs/schema/config.v1.json` does not express the 200 character cap and editors validating against the schema will not flag it. `validate-config` and the domain model do catch it, and `UserEntry` does not mirror other constraints either (`password` has `min_length=1` on `User` only), so this is consistent with what is there today. I may add it separately.

gprossliner:
I pushed a commit which should address all issues you mentioned.

cdelmonte-zg (changes requested):
Round 2 at `e79fd0f`. Both round 1 blockers are closed, and the `validate_assignment` choice is the right one. One new blocking issue comes with it, in the caller rather than in the model.

## The two round 1 blockers are closed

- **The lost test class header** is back. I checked it was not silently dropping tests: 26 collected before and after. What the missing header actually did was attach `test_get_token_service_returns_instance` to the preceding class and leave its docstring as an orphan string, so the fix restores the structure rather than recovering coverage.
- **`max_length` on assignment** is enforced now, and your reasoning is accurate: `OAuthClient` already carries `validate_assignment=True` for exactly this reason (`models.py:136`, #37), so `User` following the same rule is consistent rather than a one-off.

## Blocking: `_tool_update_user` can now leave a user half-updated

`validate_assignment=True` makes every `user.<field> = ...` capable of raising, and `_tool_update_user` (`mcp_server.py:1145-1162`) assigns nine fields one at a time onto the live `config.users[username]` object. A field that fails validation leaves every field assigned before it already committed.

Measured at this head, starting from `eve` with `password: "old"`:

| call | outcome | user afterwards |
|---|---|---|
| `update_user(password="NEWPW", email="bogus-no-at")` | `ValidationError` on `email` | password **already changed** to `NEWPW` |
| `update_user(password="NEWPW2", description="a" * 300)` | `ValidationError` on `description` | password **already changed** to `NEWPW2` |

At `51428dd` neither call raised: both invalid values were accepted, which is the bug this PR fixes. `validate_assignment=True` correctly turns those invalid assignments into failures, but it also makes the existing field-by-field mutation order observable: an earlier valid assignment can now survive a later validation failure. So this failure mode is new with this commit.

Two things make it worth blocking on rather than filing as a follow-up.

**It is reachable through the ordinary MCP path, not only the direct `_execute_tool` one.** The `update_user` input schema constrains `description` with `maxLength: 200`, but puts no format constraint on `email`, so only the model's `field_validator` rejects `bogus-no-at`. Through `call_tool` the `ValidationError` lands in the outer `except Exception` and comes back as `MCP_INTERNAL_ERROR`: the caller is told the call failed while a credential has in fact changed.

**The codebase already documents this exact hazard for the other model.** `mcp_server.py:1268`, in `_tool_update_client`:

> Validate/normalize every input up front so a bad value cannot leave the client half-updated: with validate_assignment=True, assigning each field can raise, and OAuthClient is mutated in place.

This PR gives `User` the property that comment is about, without giving `_tool_update_user` the guard that comment describes.

Mirroring `_tool_update_client` is the obvious fix: normalize and validate every argument before touching the object. Applying the assignments to a temporary copy and replacing the live user only once they have all validated would work too:

```python
candidate = user.model_copy(deep=True)
# setattr(candidate, ...) here, so validate_assignment still applies
config.users[username] = candidate
```

Note that `model_copy(deep=True)` is only cloning already-valid state there. Do not reach for `model_copy(update=...)` instead: Pydantic v2 does not validate the values in `update`, so it would be another way around the constraint rather than an enforcement of it. I checked on the version this repo pins (2.12.5): `user.model_copy(update={"description": "a" * 300, "email": "bogus-no-at"})` returns a `User` carrying both invalid values.

Please also extend the new test. `test_update_user_direct_mutation_rejects_overlong_description` asserts that the in-memory user is left untouched, which is the right property, but it passes a single field, so it stays green in exactly the case that now breaks it. A two-field call whose second field is invalid, asserting the first field did not change, is the pin.

## Everything else verified

The PR is based on `5c5e120`, so I also ran it merged with current `main` (which now carries #234 and #248):

- full suite **1628** on the head as pushed, **1676** on the trial merge with `main`
- mypy, ruff and `lint-imports` clean on the merged tree
- CI green on all nine checks
- live against a server from the merged tree: the description does render in the persona picker, and the e2e suite is **57/57**, Persona Login included

The blast radius of `validate_assignment` is contained: `_tool_update_user` is the only place in `src/` that assigns to `User` attributes at all. The web UI builds a fresh `User(...)` and therefore validates atomically, which is why nothing else needed to change.

gprossliner:
It should be fixed now. This was a nice catch!

cdelmonte-zg (approved):
Round 3 at `736b8ab`, verified. This closes it.

The half-update is fixed in the right shape: every assignment goes to a `model_copy(deep=True)` scratch copy, where `validate_assignment` still applies, and the live user is only replaced once every requested field has validated. The comment saying why it is not `model_copy(update=...)` is worth having there, since that variant skips validation entirely and would have been the obvious wrong turn.

Checked at this head:

* the two calls from round 2 (`password` plus an invalid `email` through the ordinary MCP path, `password` plus an overlong `description` through `_execute_tool`) now both raise **and** leave the password unchanged; before this commit both raised with the password already changed.
* replacing the object in `config.users` is safe for the current ownership model: that dict is the authoritative map, `get_user()` reads from it, and there is no second long-lived index or cache of `User` objects in `src/` that would also need updating when the entry is replaced.
* the new test covers exactly the failing shape from round 2: a valid field followed by an invalid one in the same call, asserting the first did not change.
* full suite green on the head, and **1677 passed** on a trial merge with the current `main` (which now carries #234 and #248); mypy, ruff and `lint-imports` clean on the merged tree.
* live e2e against a server from the merged tree with `login.mode: persona` and a described user: the description renders in the picker and the agent passes 57/57.
* GitHub CI is green across Tests, E2E and CodeQL on this head.

Both round 1 blockers and the round 2 one are closed, and the branch is clean of anything else I flagged. Approving; I will squash-merge with `Closes #244` in the commit message, since the PR body does not carry the closing keyword.

Thanks for the quick and precise turnarounds on this one.

cdelmonte-zg:
Thank you for the contribution! The descriptions make the persona picker genuinely more usable, and it was good to see the feedback from your users flow straight into the feature.

Also: now that this has landed, #250 is free to pick up whenever it suits you - the design contract in the issue still stands, no rush at all.

cdelmonte-zg:
@gprossliner v2.8.0-rc2 is out with your persona descriptions in it:

```
pip install --pre nanoidp==2.8.0rc2
# or
docker pull ghcr.io/cdelmonte-zg/nanoidp:v2.8.0-rc2
```

If you can run it against your real persona directory before we promote to final, that is exactly the soak this rc needs - descriptions on all four pickers, plus anything else your setup exercises (management_secret, the scope enforcement that has been on by default since rc1). Release notes: https://github.com/cdelmonte-zg/nanoidp/releases/tag/v2.8.0-rc2. And once the final is out, the path for #250 (auto-login) is clear whenever you want to pick it up - the design contract in the issue stands. If the rc soaks clean over the next couple of days, 2.8.0 final follows without an rc3.

gprossliner:
Thank you. I'll think I can work on #250 on Montay. So maybe it can be included in the 2.8.0 release.

PR #248, opened 2026-08-28
feat: migrate YamlWriter onto compare_and_replace (#229 phase 3)

dshvedchenko:
YamlWriter's 10 public write methods (save_user, delete_user, set_default_user, save_client, delete_client, update_oauth_settings, update_saml_settings, update_authority_prefixes,
update_allowed_identity_classes, update_login_settings) now route through the shared compare_and_replace primitive via _atomic_write, instead of _load_*_yaml() + atomic_write_yaml() directly - the same migration ConfigManager.save() got in phase 2. Each method gains an optional expected_revision (default None, unconditional - no caller passes one yet; routes/ui.py wiring is phase 4).

Every already-exists/not-found ValueError check moved from running on a document loaded before _atomic_write was even called to running inside the mutate closure compare_and_replace hands the freshly loaded document to - the same document that gets written, under the same lock. Outwardly identical for a single request, but it closes a real race: two near-simultaneous UI submissions for the same new user/client could previously both pass their "already exists" check (each loaded independently before either wrote) and the second would silently overwrite the first; now the second's check runs against a document that already reflects the first's write.

_load_users_yaml/_load_settings_yaml are unchanged - the latter has an external caller in tests/test_issue_127_settings_round_trip.py - they just have no more internal callers in the write path.

tests/test_settings_plumbing_parity.py's YamlWriter-vs-OWNED_SETTINGS parity check now excludes expected_revision from the comparison, the same way it already excludes self.

tests/test_yaml_writer_conflict.py adds phase-3 coverage: matching/ stale expected_revision on save_user/delete_user/save_client/ update_oauth_settings (ConflictError, file untouched on conflict), the unconditional (None) path staying last-write-wins, and confirmation that the already-exists/not-found checks still work correctly now that they run inside mutate.

Full suite (1648 tests), lint-imports, ruff, mypy all clean. Verified live: full e2e suite (57/57) against a running server exercising the actual web UI create/edit/delete flows this migration touches.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `LockUnavailableError` - instead of a bare `OSError` or an indefinite
   hang - when the lock file's filesystem does not support advisory locks
   or a peer process holds it for more than 10 seconds.
+- **The web UI's writer (`YamlWriter`) now uses the same write primitive
+  as `ConfigManager.save()`** (#229): `save_user`, `delete_user`,
+  `save_client`, `delete_client` and the rest of its write methods route
+  through `compare_and_replace`, and each gained an optional
+  `expected_revision` precondition (unused by any caller yet). The
+  practical effect today: creating or deleting a user/client now checks
+  "does this already exist / does this exist at all" against the same
+  document it writes, under the same lock, instead of against a copy
+  loaded before the write started. Previously, two near-simultaneous
+  submissions creating the same new user or client could both pass
+  their "already exists" check and the second would silently overwrite
+  the first, with no error to either request; that lost update is now
+  impossible - the second request correctly gets an "already exists"
+  error instead.
 
 ### Security
 - **Opt-in `management_secret` mutation gate** (#163): one shared secret that
--- a/book/src/guides/extending.md
+++ b/book/src/guides/extending.md
@@ -82,10 +82,11 @@ one coordinated, conflict-checked save (#229): both files are written,
 both hooks then run, and the runtime is refreshed from disk once before a
 `strict` failure is raised - a failing hook on one file does not leave
 the other unwritten.
-The web UI's settings page is a different, older path (`YamlWriter`):
-it still writes its sections one atomic write at a time, so under
-`strict` a failing hook there leaves the steps after it unsaved (the
-flash message says which write failed). Make the hook idempotent: the
+The web UI's settings page (`YamlWriter`) uses the same primitive now
+(#229 phase 3), but still writes its sections one call at a time
+rather than as a batch, so under `strict` a failing hook there leaves
+the later sections in that page's save unsaved (the flash message says
+which write failed). Make the hook idempotent: the
 next save re-mirrors. `on_audit_event` never fails the request that
 produced the event: a token already issued must not turn into a 500 because
 an audit sink is down. Failures are counted per hook and shown by
--- a/src/nanoidp/services/yaml_writer.py
+++ b/src/nanoidp/services/yaml_writer.py
@@ -5,14 +5,14 @@ Provides atomic write operations for YAML configuration files.
 
 import logging
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional
 
 from ..config import OAuthClient, Settings, User, get_config
 from ..config_documents import document_defaults
+from ..config_writer import compare_and_replace
 from ..hooks import HookError
 from ..serialization import (
     OWNED_SETTINGS,
-    atomic_write_yaml,
     client_id_matches,
     client_to_yaml,
     is_unchanged,
@@ -34,12 +34,30 @@ class YamlWriter:
         self.users_file = self.config_dir / "users.yaml"
         self.settings_file = self.config_dir / "settings.yaml"
 
-    def _atomic_write(self, file_path: Path, data: Dict[str, Any]) -> None:
-        """Atomically write a YAML document (shared implementation, #83), then
-        run the single post-write contract (#185):
+    def _atomic_write(
+        self,
+        file_path: Path,
+        mutate: Callable[[Dict[str, Any]], Any],
+        expected_revision: Optional[str] = None,
+    ) -> None:
+        """Write via the shared compare-and-replace primitive (#229 phase 3),
+        then run the single post-write contract (#185):
 
-            write local -> notify the mirror (on_config_saved) -> refresh the
-            runtime FROM LOCAL DISK -> propagate a mirror failure if strict.
+            check -> load -> mutate -> replace -> notify the mirror
+            (on_config_saved) -> refresh the runtime FROM LOCAL DISK ->
+            propagate a mirror failure if strict.
+
+        ``mutate`` receives the freshly loaded document and edits it in
+        place (``compare_and_replace``'s own contract) - every caller's
+        existence/not-found check now happens *inside* ``mutate`` rather
+        than on a document loaded beforehand, so it runs against the
+        same document that gets written, under the same lock, instead of
+        racing a concurrent writer between an earlier check and this
+        write (the #229 the whole primitive exists to close).
+        ``expected_revision`` (default ``None``, unconditional - today's
+        last-write-wins) is a stale-precondition check: a mismatch
+        raises ``ConflictError`` before the file is even loaded, let
+        alone mutated.
 
         The refresh is ConfigManager.reload_local(), never reload(): a
         reload() would run on_before_load and could pull a stale or
@@ -48,7 +66,7 @@ class YamlWriter:
         newest state after our own write; only an explicit reload consults
         the mirror.
         """
-        atomic_write_yaml(file_path, data)
+        compare_and_replace(file_path, expected_revision, mutate)
         kind = "users" if file_path.name == "users.yaml" else "settings"
         config = get_config()
         hook_error: Optional[HookError] = None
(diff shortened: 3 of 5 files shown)


cdelmonte-zg (changes requested):
Round 1 at `48197ee`. The migration itself is right, and the headline claim is real: I reproduced it rather than taking the description at face value.

## Verified

Everything you reported holds at this head, and the base is the merged `974e219`:

- full suite **1648 passed**, mypy clean on 35 source files, ruff clean, `lint-imports` 2 kept / 0 broken
- live e2e against a running server from this branch: **57/57** (Core 2, OAuth2/OIDC 24, SAML 14, Persona 1, Keys 3, REST API 11, Management Secret 2)
- the UI's negative paths still surface as before, not as a 500: creating a user that exists and deleting one that does not both give 302 plus the "already exists" / "not found" flash, no traceback in the server log, and the duplicate create leaves the existing entry untouched. That is the part I most wanted to see, since those `ValueError`s now originate inside `mutate` rather than before the write.

**The race you describe is real and this closes it.** Two threads on a `Barrier(2)`, both calling `save_user(User(username="bob", ...), is_new=True)` on the same freshly seeded directory, 20 trials:

| | `974e219` (before) | `48197ee` (this PR) |
|---|---|---|
| both writers returned OK (update lost) | **20 / 20** | **0 / 20** |
| one OK, one `ValueError` (correct) | 0 / 20 | **20 / 20** |

## Blocking 1: this silently adds `default_user` to a `users.yaml` that never had it

`_load_users_yaml()` applied `{"users": {}, "default_user": "admin"}` **only when the file did not exist** (`yaml_writer.py:83`). The new `data.setdefault("default_user", "admin")` inside `save_user`'s and `delete_user`'s `mutate` applies whenever the key is absent, which is not the same condition.

Reproduced, a hand-written `users.yaml` that exists without the key:

```yaml
# my users
users:
  alice:
    password: "x"
```

after one `save_user(User(username="bob", password="y"))`:

| `974e219` | `48197ee` |
|---|---|
| `users:` gains `bob`, nothing else changes | `users:` gains `bob` **and the file gains `default_user: admin`** |

So we write a key the operator never wrote, and in this example it names a user that does not exist in the file at all. The effective configuration does not change (`UsersDocument.default_user` already defaults to `"admin"`, `config_documents.py:488`), so this is file fidelity rather than a functional break, but it is the opposite of the #127 line of work where the point is that untouched placeholders and comments survive a save, and it contradicts the PR description's "Outwardly identical for a single request".

`delete_user` has a second-order version of the same thing: on a file with no `default_user` key, deleting a user named `admin` now rewrites the key to the first remaining user, where before it left the file alone. That change is arguably an improvement, but it is undocumented and it comes from the same line.

The fix is to reproduce the old condition, since an absent file loads as an empty `CommentedMap`:

```python
def mutate(data: Dict[str, Any]) -> None:
    if not data:  # missing file: the same defaults _load_users_yaml used
        data["users"] = {}
        data["default_user"] = "admin"
    data.setdefault("users", {})
    if is_new and user.username in data["users"]:
        ...
```

I tried it: the injected key is gone, the missing-file case still writes both keys in the old order, and the full suite is still 1648 green. That last part is the other half of the problem, so please pin it with a test: nothing in the suite currently notices this either way.

## Blocking 2: no CHANGELOG entry

Phase 2 got one under `### Changed` (#229). Phase 3 fixes something an operator can actually hit, and your own description states it well: two near-simultaneous UI submissions creating the same user or client could both pass their "already exists" check and the second would silently overwrite the first. That belongs in the CHANGELOG, not only in the PR body.

## Blocking 3: the property this PR is about is not pinned by a test

`test_yaml_writer_conflict.py` is good coverage of `expected_revision` and of the existence checks still working from inside `mutate`, but nothing in it fails if someone hoists a check back out of the closure: the table above is exactly the regression that would go unnoticed. Please add the concurrent case. The shape I used:

```python
barrier = threading.Barrier(2)
results = []

def go(password):
    barrier.wait()
    try:
        writer.save_user(User(username="bob", password=password), is_new=True)
        results.append("ok")
    except ValueError:
        results.append("value_error")

threads = [threading.Thread(target=go, args=(p,)) for p in ("first", "second")]
for t in threads:
    t.start()
for t in threads:
    t.join()

assert sorted(results) == ["ok", "value_error"]
```

One trial is enough as a pin, since it fails deterministically on the pre-migration code.

## Non-blocking

- **`_load_users_yaml` is now dead.** The description says the two loaders "just have no more internal callers in the write path", but that is only true of `_load_settings_yaml`, which at least keeps the caller in `tests/test_issue_127_settings_round_trip.py:286`. `_load_users_yaml` has no callers anywhere now. Delete it, or say in the docstring why it stays.
- **`book/src/guides/extending.md:85` is now inaccurate.** It calls the web UI's settings page "a different, older path (`YamlWriter`)", which stops being true with this PR: it is the same primitive now. The rest of the sentence still is true, and I checked it: `routes/ui.py` still calls four writer methods in sequence (`update_oauth_settings`, `update_saml_settings`, `update_allowed_identity_classes`, `update_login_settings`, lines 622 to 657), so under `strict` a failing hook still leaves the later steps unsaved. A one-line rewording, not a rewrite: same write primitive, still one file per call, so still sequential per section.

Nothing else. The `_apply_provided_settings` and client paths read correctly to me, exception propagation out of `mutate` is unwrapped so `ValueError` still reaches `routes/ui.py`'s handlers, and the post-write hook contract is untouched and runs after the lock is released.

cdelmonte-zg:
Small correction to two details in my review; the findings and `CHANGES_REQUESTED` status stand.

For Blocking 1, the `default_user` regression is real, but my suggested `if not data` fix was too broad. An existing empty `users.yaml` also loads as an empty mapping, whereas the old `_load_users_yaml()` only injected `{"users": {}, "default_user": "admin"}` when the file itself was absent. So please preserve the missing-vs-existing distinction rather than use emptiness as the proxy. Checking existence from inside `mutate` is one small way to do that, since the compare-and-replace lock is already held there.

For Blocking 3, the 20/20 repro is solid, but I overstated that the simple `Barrier(2)` at entry makes a single trial deterministic on the old implementation. It makes the race very easy to reproduce, but it does not force both old `_load_users_yaml()` calls to complete before either write. Please make the regression deterministic by synchronizing the old pre-write load/check boundary (or use an equivalent controlled interleaving), so the test is genuinely a pin rather than a scheduler-dependent stress test.

I would also reclassify the missing CHANGELOG entry from a code blocker to a before-merge item. It still belongs in this PR or in the maintainer merge cleanup because the fixed lost-update race is user-visible.

dshvedchenko:
fix: preserve users.yaml's missing-vs-existing distinction, pin the lost-update race (#248 review round 1, blocking x2)

Blocking 1: save_user's and delete_user's mutate closures used
data.setdefault("default_user", "admin") to reproduce
_load_users_yaml's old skeleton, but that old skeleton
({"users": {}, "default_user": "admin"}) only applied when users.yaml
did not exist at all - not whenever the default_user key happened to
be absent. Reproduced: a hand-written users.yaml that exists without
that key gained an unrequested default_user: admin (potentially naming
a user that doesn't even exist in the file) on its next save_user/
delete_user call, contradicting the #127 line of work where an
untouched file survives a save unchanged. Fixed by checking
self.users_file.exists() from inside mutate (accurate there, since the
lock compare_and_replace holds is already in effect by the time mutate
runs) and only injecting the skeleton for a genuinely missing file,
matching the old condition exactly - review's own suggested `if not
data` fix was itself corrected in a follow-up comment, since an
existing empty file also loads as `{}` and must not get the skeleton
either.

Blocking 3: the concurrent-create pin (test_yaml_writer_conflict.py)
proved real to me by reverting it, but a single Barrier(2)-synchronized
trial doesn't provably force both threads' load to complete before
either write - it makes the race very likely on a regression, not
certain. Strengthened to 20 repeated trials asserting the invariant on
every one, matching the reviewer's own 20/20 empirical verification
methodology, rather than depending on a fragile monkeypatched
synchronization point tied to today's exact call graph. Verified by
reverting save_user's mutate to the old load-outside-the-lock pattern:
fails on trial 0 with ['ok', 'ok'] (both writers succeeded, the exact
lost update), passes with the fix restored.

Both fixes independently verified: reverted each one and reran its
new test to confirm the test actually catches the regression, then
restored.

cdelmonte-zg (approved):
Round 2 at `fc5f71c`, verified.

All three blocking findings from round 1 are closed.

* The `users.yaml` fidelity regression is fixed with the important missing-vs-existing distinction preserved: only a genuinely absent file gets the old `{"users": {}, "default_user": "admin"}` skeleton. Existing files without the key, including empty files, remain without it. I checked all four starting shapes against the pre-phase-3 behaviour and also verified the real UI path against a hand-written file without `default_user`.
* The lost-update fix now has regression coverage.
* The CHANGELOG entry is in and describes the user-visible race accurately.

On the concurrency test specifically: this is not the deterministic rendezvous I asked for in round 1. The new test correctly acknowledges that distinction rather than claiming otherwise, and I am accepting the substitution. I put the old load-before-write shape back locally and this test caught it consistently at the first trial; on this head it stayed green across repeated runs and under CPU contention. With 20 trials at negligible cost, that is a sufficient regression guard here without coupling the test to an internal synchronization seam that this same change removed.

The two non-blocking items are also addressed: `_load_users_yaml` is gone and `extending.md` now accurately says that the settings page uses the shared primitive but still performs its section writes as separate sequential calls.

One tiny documentation nit, not worth another round: the new comments still refer to `_load_users_yaml` by name even though this commit removes it. "pre-phase-3 users.yaml loading semantics" would age better, but I am not holding the PR for that.

Full suite, lint/type/import checks, Windows CI and live e2e are all green on this head.

Approving.

Issue #249, opened 2026-08-28
feat: ui auth dialog to have horizontal composition


dshvedchenko:
allow user to specify per client selector for switch from vertical auth dialog to horizontal

cdelmonte-zg:
Accepted in principle: this is a natural continuation of the per-client branding you built in #157, and a layout choice expressed as a value is exactly the kind of presentation VISION principle 7 allows (structured data, never free-form markup or CSS). Two things to pin down before a PR, one of them a question for you.

## Which composition, exactly

Today `authorize.html` is one vertical card: header, the client block (logo, client id, description, scope badges), the form (persona picker or username/password), footer. I read "horizontal" as the client block and the form side by side, with header and footer still full width. If you mean something else, a quick sketch or a screenshot of the target would settle it faster than prose.

Two constraints that follow from where branding lives:

- **Scope is `/authorize`.** It is the only page with a client context and the only one that renders per-client branding; `/login` has no client, and the device verification page does not carry branding today. Keeping the layout switch on the same page as the rest of the branding avoids inventing client context where there is none.
- **It has to stay a choice between two nanoidp-owned layouts.** The horizontal variant must collapse back to the vertical one on narrow viewports (a Bootstrap `row`/`col-md` split does that for free), and no per-client CSS or markup enters the page. The CSS for the card is inline in `authorize.html`, so this is a template change plus a few inline rules, not a new static asset.

## Contract for the field

- A new `OAuthClient.layout`, values `vertical` (default) and `horizontal`, validated at the model. Absent in YAML means `vertical`, and editing an existing client that never declared `layout` must not materialize `layout: vertical` unless the operator actually changes that field, preserving the #127 round-trip rule.
- It goes through the documented client-field flow: model, `config_documents`, `client_to_yaml` and `merge_client_entry`, the clients form (a select next to the `show_*` toggles is the natural spot), create and edit in `routes/ui.py`, `_client_to_dict` plus the `create_client` and `update_client` schemas and handlers in `mcp_server.py`, and schema regeneration. `tests/test_client_field_parity.py` will catch the declarative surfaces automatically; the imperative legs (`client_to_yaml`, `merge_client_entry`, the UI parsers and the MCP handlers) still need feature-specific tests, as that parity test intentionally documents. In `_tool_update_client`, validate/normalize the enum before mutating the live client, following the existing pre-validation rule there so an invalid layout cannot leave another requested field partially updated.
[...]

Issue #246, opened 2026-08-27
define the configuration consistency boundary (ConfigFileStore)


cdelmonte-zg:
#229 stated the read side of the write pipeline as an explicit simplification:

> Readers never take the lock: the atomic replace already guarantees a complete file.

That holds on POSIX. It does not hold on Windows, and it was never the whole story anywhere: "a reader never sees a partial file" and "a reader always sees ONE consistent observation of the directory" are different properties, and nanoidp provides neither today.

This issue defines the consistency boundary that does not exist yet. It is not a deduplication refactor.

## Census (verified on main, 2026-09-18)

**The Windows failure is live, not historical.** `TestConcurrentReadsOutsideTheLock` is still **XFAIL** in the `windows-smoke` job on today's main. The write-side retry added by #234 did not close the reader-side hole.

**Reads are outside the protocol, and the code says so.** `config.py:311` already carries the comment "The read itself is still unlocked (#246)", and `config.py:555` documents the mixed snapshot: a concurrent `save()` in another process can land its `users.yaml` between this process's `settings.yaml` and `users.yaml` reads.

**One operation makes several observations of the same file**, measured by counting opens:

```text
validate_config_result()   settings.yaml opened 3x, users.yaml 1x
ConfigManager load         settings.yaml opened 2x, users.yaml 1x
```

The load's two reads are the `declared_mode` pre-read and `_stage_directory`, and the revision is computed from the second only, so the pre-read can have seen different bytes than the parse that the revision describes. That one is a runtime path, not a diagnostic one.

**Reentrancy is mandatory from the first commit, verified rather than assumed:**

```text
_cross_process_lock nested on the same directory
  -> waits the full _LOCK_TIMEOUT_SECONDS (10.0s), then LockUnavailableError
```

A fresh fd per acquisition, and `flock` is per open file description, so the process conflicts with itself. `_write_lock` is worse: a non-reentrant `threading.Lock`, so a second acquisition in the same thread deadlocks with no timeout at all. And the writer reads inside its own critical section, `current_revision` in phase 1 and `load_yaml_document` in phase 2.

**No read is on a per-request path**, which removes the main objection to letting reads join the protocol:

| reader | when |
| --- | --- |
| `_stage_directory` (two separate `open()` calls) | load and reload |
| `yaml_writer` loaders | UI saves |
| `config_validation._read_yaml`, `declared_validation_mode` | validation, startup, CLI |
| `hooks.py` bootstrap read | startup only |

[...]

PR #251, opened 2026-08-28
feat: wire the web UI's forms to the compare_and_replace precondition (#229 phase 4)

dshvedchenko:
Every routes/ui.py write flow now carries the revision its form was rendered with, closing the gap phases 1-3 opened but didn't use: a hidden expected_revision field, submitted back to the corresponding YamlWriter call, refusing a stale write with ConflictError and a flash instead of silently overwriting a concurrent change - the same pattern every existing ValueError/HookError except-branch already used. A submission with no expected_revision at all (an old cached page, a script, the e2e test agent) keeps today's unconditional last-write-wins; nothing requires an existing caller to opt in.

services/yaml_writer.py: every public write method (and _atomic_write itself) now returns its new revision instead of None. The settings page - the one form whose submission spans several logical sections of settings.yaml - applies oauth, saml, identity classes and login mode as ONE write: the per-section mutation logic is extracted into module-level pure functions (_mutate_settings_section, _mutate_allowed_identity_classes, _mutate_login_mode) and the new update_settings_form composes all four under a single compare_and_replace call, so expected_revision covers the entire submission and a conflict is all-or-nothing, exactly like every other form. An earlier version of this PR chained four separate writes by revision, which made each individual write conflict-safe but not the page: a conflict on write N left writes 1..N-1 already committed, their on_config_saved hooks already fired and the runtime already reloaded, while the flash said nothing had landed (reproduced in review with an on_config_saved hook that moves the file after every save - the same shape a second admin, the MCP companion or a mirror hook could produce). Composing the sections into one write closes that, and on_config_saved/reload_local now run once per settings save instead of up to four times. The standalone single-section writers (update_oauth_settings, update_saml_settings, update_allowed_identity_classes, update_login_settings) are unchanged from the outside and still perform exactly one write each; login_mode is still validated before the file is ever touched.

routes/ui.py: all 12 YamlWriter call sites across 9 handlers (user create/edit/delete, client create/edit/delete/regenerate-secret, settings, authority prefixes) read expected_revision from request.form and catch ConflictError with one shared flash phrasing naming the file and what changed. Delete/regenerate-secret buttons live on list pages (no dedicated edit form of their own), so their hidden field comes from that list page's own render.

Templates (users_form.html incl. the delete modal, clients_form.html, the mini-forms on clients.html, settings.html, claims.html) each gained a hidden
<input type="hidden" name="expected_revision" value="{{ revision }}"> on their write forms.

tests/test_config_schema.py: the settings-form-field parity check (every posted field must map to a real settings.yaml key) now excludes expected_revision, the same way it already excludes the checkbox __on_form markers - it isn't a settings key either.

tests/test_ui_expected_revision.py: every GET page embeds the real current revision (not an empty/undefined one); a stale POST on each flow, the create flows included, gets a clean redirect+flash and leaves the target untouched; fresh submissions still succeed, including the settings page landing all four sections from one write; a POST with no expected_revision field at all stays unconditional; and TestSettingsFormIsAllOrNothing pins the composed design by registering an on_config_saved hook that moves settings.yaml after every save and asserting the submission's fields land all together or not at all - it fails against the chained design (first section lands, the rest never run) and passes against the composed one.

examples/test_agent.py: first e2e coverage for #229 - test_config_write_conflict_detection fetches the clients page's rendered revision, advances settings.yaml with an unrelated client creation, then resubmits the create form for a different client with the stale revision, asserting the conflict flash appears and the client was not created.

book/src/guides/extending.md's hook section updated to match (the settings page's one POST is one write now, the same shape as ConfigManager.save()'s two-file write). CHANGELOG: new entry under Changed, and phase 3's entry corrected - "unused by any caller yet" was true when it was written, this phase is that caller.

Full suite (1672 tests), lint-imports, ruff, mypy all clean. Verified live against a running server, not just tests: a stale revision on a real form gets the exact conflict flash with no 500 and the target confirmed unchanged on disk; one settings POST changing oauth.audience, saml.entity_id and allowed_identity_classes lands all three together with the on_config_saved hook firing exactly once for the submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,12 +117,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   "changed since it was last read - please reload and try again" flash
   instead of silently overwriting someone else's concurrent change (an
   edit based on a page loaded before another admin deleted or changed
-  the same user/client, for instance). The settings page writes
-  `settings.yaml` four times per submission (OAuth, SAML, identity
-  classes, login mode); each write after the first checks against the
-  revision the *previous write in that same request* produced, so a
-  concurrent writer interleaving between any of the four is still
-  caught, not just one before the first. A submission with no
+  the same user/client, for instance). The settings page's OAuth, SAML,
+  identity-classes and login-mode fields are now applied as one write
+  instead of four separate ones, so a conflict there is all-or-nothing,
+  the same guarantee every other form already had - an earlier version
+  of this ran four writes chained by revision, which meant a conflict
+  partway through could leave the earlier sections already saved while
+  the page reported that nothing had changed. A submission with no
   `expected_revision` at all (an old cached page, a script, the e2e
   test agent) keeps today's unconditional last-write-wins - nothing
   about this requires an existing caller to opt in.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,16 +100,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   as `ConfigManager.save()`** (#229): `save_user`, `delete_user`,
   `save_client`, `delete_client` and the rest of its write methods route
   through `compare_and_replace`, and each gained an optional
-  `expected_revision` precondition (unused by any caller yet). The
-  practical effect today: creating or deleting a user/client now checks
-  "does this already exist / does this exist at all" against the same
-  document it writes, under the same lock, instead of against a copy
-  loaded before the write started. Previously, two near-simultaneous
-  submissions creating the same new user or client could both pass
-  their "already exists" check and the second would silently overwrite
-  the first, with no error to either request; that lost update is now
-  impossible - the second request correctly gets an "already exists"
-  error instead.
+  `expected_revision` precondition. The practical effect: creating or
+  deleting a user/client now checks "does this already exist / does
+  this exist at all" against the same document it writes, under the
+  same lock, instead of against a copy loaded before the write started.
+  Previously, two near-simultaneous submissions creating the same new
+  user or client could both pass their "already exists" check and the
+  second would silently overwrite the first, with no error to either
+  request; that lost update is now impossible - the second request
+  correctly gets an "already exists" error instead.
+- **Every web UI form that writes `users.yaml`/`settings.yaml` now
+  carries the revision it was rendered with, and refuses a stale
+  submission** (#229 phase 4): create/edit/delete user, create/edit
+  client, regenerate client secret, settings, and authority prefixes
+  all send a hidden `expected_revision` field and get a clear
+  "changed since it was last read - please reload and try again" flash
+  instead of silently overwriting someone else's concurrent change (an
+  edit based on a page loaded before another admin deleted or changed
+  the same user/client, for instance). The settings page writes
+  `settings.yaml` four times per submission (OAuth, SAML, identity
+  classes, login mode); each write after the first checks against the
+  revision the *previous write in that same request* produced, so a
+  concurrent writer interleaving between any of the four is still
+  caught, not just one before the first. A submission with no
+  `expected_revision` at all (an old cached page, a script, the e2e
+  test agent) keeps today's unconditional last-write-wins - nothing
+  about this requires an existing caller to opt in.
 
 ### Security
 - **Opt-in `management_secret` mutation gate** (#163): one shared secret that
--- a/book/src/guides/extending.md
+++ b/book/src/guides/extending.md
@@ -82,11 +82,17 @@ one coordinated, conflict-checked save (#229): both files are written,
 both hooks then run, and the runtime is refreshed from disk once before a
 `strict` failure is raised - a failing hook on one file does not leave
 the other unwritten.
-The web UI's settings page (`YamlWriter`) uses the same primitive now
-(#229 phase 3), but still writes its sections one call at a time
-rather than as a batch, so under `strict` a failing hook there leaves
-the later sections in that page's save unsaved (the flash message says
-which write failed). Make the hook idempotent: the
+The web UI's settings page (`YamlWriter`) uses the same primitive now:
+its one POST composes oauth, saml, identity classes and login mode
+into a single write (#229 phase 4 - an earlier version ran four
+separate writes chained by revision, which meant a conflict partway
+through could leave the earlier sections already saved while the page
+reported nothing had changed; composing them into one write closed
+that), so a failing hook under `strict` there is the same shape as
+`ConfigManager.save()`'s: the one write already landed, only the
+mirror push failed. `YamlWriter`'s other methods (`save_user`,
+`save_client`, ...) each still make their own single write per call,
+same as always. Make the hook idempotent: the
 next save re-mirrors. `on_audit_event` never fails the request that
 produced the event: a token already issued must not turn into a 500 because
 an audit sink is down. Failures are counted per hook and shown by
--- a/examples/test_agent.py
+++ b/examples/test_agent.py
@@ -50,6 +50,7 @@ import base64
 import hashlib
 import json
 import os
+import re
 import secrets
 import sys
 import time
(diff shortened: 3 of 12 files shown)


cdelmonte-zg (changes requested):
Round 1 at `d82545b`. The wiring is complete and correct on every single-write flow; I found one design problem on the settings page that has to be settled before this lands, plus one gap against the project's own definition of done.

## Verified

- full suite **1669 passed** here as well; mypy clean on 35 source files, ruff clean, `lint-imports` 2 kept / 0 broken; CI green on all nine checks. The PR sits on the current tip of `main`, so no trial merge was needed.
- every write form carries the hidden field: `users_form.html` (edit form and the delete modal), `clients_form.html`, both mini-forms on `clients.html`, `settings.html`, `claims.html`; `users.html` and `user_detail.html` have no write forms, so nothing is missing there. All seven GET renders pass a real `current_revision(...)`.
- the chained revision on the settings page does what it says when nobody interferes: one submission changing `oauth.audience`, `saml.entity_id`, `allowed_identity_classes` and `login.mode` lands all four (that is your live check, and mine matched it).
- the "absent means unconditional" contract holds for the e2e agent's existing posts (`/settings`, `/users/create`, `/clients/create`, `/clients/<id>/delete`), none of which send the field.

## Blocking 1: a mid-chain conflict on the settings page leaves the submission half-applied, and the flash says the opposite

The chain guards each of the four writes, but a conflict on write *n* is raised **after writes 1 to n-1 have already been committed, their `on_config_saved` hooks fired, and the runtime reloaded**. The user then sees "please reload the page and try again" for a page that has already absorbed part of what they typed.

Reproduced live, deterministically. The hook below is only a deterministic way to force the interleaving; I am not proposing that rewriting `{path}` from `on_config_saved` is a normal hook use case. The same window exists for an independent UI or MCP writer between the four separately locked writes, since each `_atomic_write` completes entirely before the next one starts. `bootstrap.yaml`:

```yaml
hooks:
  on_config_saved: "sh -c 'echo \"# touched by another actor\" >> {path}'"
```

That appends one comment to `settings.yaml` after every settings write, so the file's revision moves between write 1 and write 2 exactly as a second admin, the MCP companion or a mirror hook would move it. Then one `POST /settings` with the form's genuine `expected_revision` and four changed fields:

| | `main` (`0b671c1`) | `d82545b` (this PR) |
|---|---|---|
| `oauth.audience` | written | **written** |
| `saml.entity_id` | written | not written |
| `allowed_identity_classes` | written | not written |
| `login.mode` | written | not written |
| settings writes performed | 4 | 1 |
| what the user is told | "Settings updated successfully" | `settings.yaml changed since it was last read (expected revision a441f3c14d54, found b8fe3ea50b5d) - please reload the page and try again` |

No 500, no traceback: the conflict path itself is clean. The problem is the outcome it describes. Before this PR the whole form landed (last-write-wins, the bug #229 is for); with it, the form lands partially and the message implies it did not land at all. If the operator follows the instruction, the reloaded page shows their own `audience` edit as though someone else had made it, and the mirror hook has already pushed a half-applied state. That is the same shape as the "save() can fail after committing" finding in #234 round 8, and it is what phase 2 avoided for `ConfigManager.save()` by checking every precondition before writing anything.

The root cause is structural rather than a missing check: the settings page performs four separate `compare_and_replace` calls on **one file**, so there is no point at which the page's precondition can cover the whole submission. The chained revision is a workaround for that structure, and the CHANGELOG, the docstring and the comment in `settings()` all spend paragraphs explaining the workaround.

The property I would require is **one compare-and-replace boundary for one settings-form submission**. The existing section mutations can be factored into non-writing helpers and composed under one `_atomic_write`, or implemented with an equivalent single-write design; the important part is that the form's revision is checked once before any part of that submission becomes visible. Composing them is the natural shape: it removes the chain entirely (no more "each write after the first checks against the previous write" story to document), makes a conflict all-or-nothing the way `save()` already is, and as a side effect fires `on_config_saved` and `reload_local()` once per settings save instead of four times, which the table above shows is what happens today. Note that `compare_and_replace_many` is not the tool for this, since it rejects a repeated path by design; this is one path with one composed mutation.

`book/src/guides/extending.md:85-89` needs the matching update in either case: it currently says the settings page "still writes its sections one call at a time", which becomes false with the fix, and if the chain were kept instead it would need to state that a conflict mid-chain leaves the earlier sections saved.

Please also pin the property with a test. The hook trick above works in-process too (the unit suite already registers hooks in-process), so a test can register an `on_config_saved` hook that touches the file, post the settings form with a fresh revision, and assert that either all four sections changed or none did. `test_settings_with_stale_revision_flashes_and_leaves_settings_untouched` does not cover this: it makes the *form's* revision stale, which is caught on write 1 before anything is written.

## Blocking 2: no e2e coverage for the one thing an operator can see

VISION principle 5 is that a feature lands with its `examples/test_agent.py` coverage in the same PR, and this phase is the first user-visible piece of #229: a flash message a person reads. The unit tests exercise it through the Flask test client; the live agent, which is what runs against a real server in CI, never sends `expected_revision` at all (0 occurrences). One check is enough: fetch a form, take its rendered revision, advance the file through an unrelated write, post the stale revision, assert the redirect carries the conflict flash and the target is unchanged. You did exactly this by hand for regenerate-secret; it just needs to live in the agent.

## Non-blocking

- `TestStaleFormSubmissionIsAConflictNotA500` covers seven flows, but the two create flows (`/users/create`, `/clients/create`) have no stale test. The code path is the same; the pin is cheap.
- The PR body says "all 12 YamlWriter call sites across 7 handlers" and then lists nine; it is 12 calls across 9 handlers (3 user, 4 client, settings, claims), the settings page being why 9 handlers make 12 calls. Not important, but the description will be squashed into `main`, so worth a pass.
- Once Blocking 1 is settled the CHANGELOG entry needs rewriting: the paragraph about chaining goes away, and the sentence "nothing about this requires an existing caller to opt in" should stay.

Everything on the single-write flows (users, clients, regenerate-secret, claims) is right, and the shared `_conflict_message` plus a dedicated `except ConflictError` branch per handler is the pattern the existing `ValueError`/`HookError` branches already set. The settings page is the one place where the shape of the write, not the wiring, is the issue.

dshvedchenko:
fix: compose the settings page into one write, not four chained ones (#249 review round 1, blocking x2)

Blocking 1 - the settings page's revision chain (write N checks
against write N-1's own resulting revision) made each individual write
conflict-safe, but the page's overall submission was not: a conflict on
write N left writes 1..N-1 already committed, their on_config_saved
hooks already fired and the runtime already reloaded, while the flash
still said "please reload and try again" for a page that had, in fact,
partly landed. Reproduced live and in a new test with an
on_config_saved hook that appends a comment to settings.yaml after
every save (the same shape a second admin, the MCP companion, or a
mirror hook could produce): one submission with three changed fields
landed only the first, silently, with no indication which.

services/yaml_writer.py: extracted the per-section mutation logic
(_mutate_settings_section, _mutate_allowed_identity_classes,
_mutate_login_mode) into module-level pure functions so several
sections can be composed under one document instead of each running
its own compare_and_replace. update_oauth_settings/update_saml_settings/
update_allowed_identity_classes/update_login_settings are unchanged
from the outside - many existing tests call them individually as
standalone single-section writes, and they still do exactly that,
just via the shared helpers internally. New update_settings_form(...)
composes all four sections into ONE compare_and_replace call:
oauth_fields and saml_fields as dicts (same shape
_apply_provided_settings already took), allowed_identity_classes and
login_mode alongside. login_mode is still validated before the file is
ever touched, same as update_login_settings always did.

routes/ui.py's settings() POST now calls update_settings_form once
instead of four sequential writer calls - the chaining logic and its
explanatory comment are both gone, since there's nothing left to chain.
expected_revision now covers the entire submission: a conflict is
all-or-nothing, exactly like every other form already was, and
on_config_saved/reload_local each run once per settings save instead
of up to four times.

book/src/guides/extending.md's hook section updated to match: the
settings page's one POST is one write now, the same shape as
ConfigManager.save()'s two-file write, not "one call at a time."
CHANGELOG's phase 4 entry rewritten to drop the chaining explanation.

tests/test_ui_expected_revision.py adds
TestSettingsFormIsAllOrNothing, registering the same on_config_saved
hook in-process and asserting the three test fields either all changed
or none did - verified it fails against the old chained design
(audience lands, saml/identity-classes don't) and passes against the
composed one. Also adds the two missing stale-create tests
(users/create, clients/create) the reviewer flagged as cheap gaps.

Blocking 2 - examples/test_agent.py gets its first #229 coverage:
test_config_write_conflict_detection fetches the clients page's
rendered revision, advances settings.yaml with an unrelated client
creation, then resubmits the create form for a different client using
the stale revision, asserting the conflict flash appears and the
client was not created. Verified live against a running server: 58/58,
including this one; also manually reproduced the reviewer's exact hook
trick against a fresh instance and confirmed all three settings fields
land together, not partially.

Full suite (1672 tests), lint-imports, ruff, mypy all clean.

cdelmonte-zg (approved):
## Round 2 - verified at 12c3043, APPROVE

Both blockers are closed, verified empirically at this head.

**Blocking 1 (all-or-nothing settings save): closed.** `update_settings_form` composes the four section mutations under one `compare_and_replace`, and the extraction into module-level pure helpers keeps the standalone single-section writers byte-for-byte identical in behavior. What I ran:

- The new `TestSettingsFormIsAllOrNothing` is a real pin: I ran it against the round-1 chained design (d82545b + only the new test file) and it fails there with the partial application, passes here.
- Live, with the same `on_config_saved` hook probe from round 1: one POST changing oauth.audience, saml.entity_id and allowed_identity_classes lands all three together, and a hook-fire counter on a separate log file (the counter on settings.yaml itself is confounded by the round-trip consolidating trailing comment lines) shows exactly ONE fire per submission, zero at boot - previously up to four.
- Live negative: a stale-revision settings POST lands nothing - no field changed, zero writes, zero hook fires.

**Blocking 2 (e2e): closed.** `test_config_write_conflict_detection` ran against a live server from this branch on a fresh `init` config: 57/58, the new test passing. The one failure (Redirect URI Exact Match) reproduces identically on current main with the same fresh-init setup (56/57), so it is the environmental baseline, not this PR.

Round-1 non-blockings all addressed: the two stale-create tests, the CHANGELOG chain paragraph, extending.md.

Also verified: full suite 1672 green, mypy, ruff, lint-imports 2/2, CI 9/9 including windows-smoke, and a trial merge with current main (a1dbdf5, which touched ui.py and users_form.html too) is conflict-free with 1697 tests green.

**One pre-merge item, no new commit needed for most of it: the PR body is still the round-1 text** - it describes the revision chain as the current design ("each write after that chains from the PREVIOUS write's own resulting revision"), says "the settings page's 4-write chain succeeds end to end", counts 9 call sites across 7 handlers (it is 12 calls across 9 handlers) and 1669 tests (now 1672). Since the squash message is composed from it, please refresh it to describe the composed single-write design before merge. Two small reference slips you could fold into the same pass: the round-2 commit headline says "#249 review round 1" (this is #229 phase 4 / this PR's round 1 - moot under squash), and `update_settings_form`'s docstring credits the repro to "(#248 review)" where it was this PR's round 1 review - that one does land on main, fine to fix whenever.

Approval stands at this head; the body refresh does not change code.

cdelmonte-zg:
Merged (squash cc79cac). I refreshed the PR body and title myself before the squash so the commit message on main describes the round-2 composed design - no action needed on your side. Phase 5 (API/MCP callers passing a real revision) is what remains of #229, and #249 is now unblocked on this landing.

cdelmonte-zg:
Thank you for the contribution! By the way, out of curiosity: are you using nanoidp for your own projects, or is it being used in a company/team environment as well?

If the latter and the company is comfortable being named publicly, I'd be happy to add it to the Adopters section in the README (https://github.com/cdelmonte-zg/nanoidp#adopters). No pressure at all if you'd rather keep that private.

cdelmonte-zg:
@dshvedchenko v2.8.0-rc2 is out and it is the candidate that actually contains your #234 -> #248 -> #251 chain, completed on the MCP side by #252 (read tools expose the loaded revisions, save_config takes them as preconditions with an all-or-nothing two-file contract):

```
pip install --pre nanoidp==2.8.0rc2
# or
docker pull ghcr.io/cdelmonte-zg/nanoidp:v2.8.0-rc2
```

If you can give it a real-world pass in your stack before we promote to final, the scenario that deserves it most is the one this whole chain was built for: the web UI and an MCP client (or two nanoidp processes) writing the same config directory concurrently - you should now get conflict flashes / kind=conflict where 2.7 silently overwrote. Release notes: https://github.com/cdelmonte-zg/nanoidp/releases/tag/v2.8.0-rc2 - the Windows reader limitation is documented there as known (#246). If the rc soaks clean over the next couple of days, 2.8.0 final follows without an rc3.

dshvedchenko:
I use it as idp in dev mode for project. Nanoidp will not be used in production. But imho it is very useful in dev and experimentation. I just wanted to help develop what I used

cdelmonte-zg:
Thanks for saying so, @dshvedchenko - "a dev/experimentation IdP, not for production" is exactly the niche nanoidp is built to fill, so that is good to hear. And "help develop what I used" is the best kind of contribution: your PRs across config, MCP and the write pipeline have made it noticeably better for that use. Much appreciated.

Issue #235, opened 2026-08-25
RuntimeStore, phase 1: an in-memory home for runtime-created users and clients


cdelmonte-zg:
*Revised 2026-09-14. The earlier text put all execution state (authorization codes, device codes, revocations, audit) and a SQLite backend for several processes into one issue. Phase 1 is narrower; the SQLite backend is #354; moving the protocol state behind the boundary is phase 2 below. Earlier text in the edit history.*

**Principle.** YAML is configuration; the RuntimeStore is execution state. Durability stays orthogonal to declaration: a later durable backend (#354) would not make runtime objects declared.

**Phase 1 scope.**

- A typed boundary with two repositories, `runtime_users` and `runtime_clients`, exposing the operations #192 and #190 need: create, get, list, delete, delete all. It is not a key-value store.
- `MemoryRuntimeStore` only. It is owned by the single composition of the IdP process (#230), so every surface resolves through one instance.
- Composition sits above the config loader: lookups consult declared configuration, then the store, and the collision rule of #192 is enforced at creation and at reload. `ConfigManager.load()` does not know the store exists.
- One lock, as the existing in-memory stores use.
- Contract tests written against the repository interface, so that a later backend has a test suite to target.

**Phase 2 (a separate, behaviour-preserving PR, not required by #192 or #190).** Authorization codes, device codes, revocations and the audit behind the same boundary. Only when it pays off, for example as a prerequisite of #354.

**Out.** SQLite, several processes, schema versioning, runtime entities kept across restarts: all in #354. Unchanged, as in the earlier text: declared configuration and its write pipeline (#229), signing keys, the Flask session, the per-process rate limiter.

Depends on: #230.

mahirhir:
Here is a concrete blueprint for `RuntimeStore` connection isolation, schema versioning, and WAL retention:

### 1. SQLite Engine Initialization & Schema Guard

To ensure safe multi-process access across Flask workers and `nanoidp-mcp` companions without locking spikes:

```python
import sqlite3
from pathlib import Path
from contextlib import contextmanager

CURRENT_SCHEMA_VERSION = 1

class SqliteRuntimeStore:
    def __init__(self, db_path: str | Path, busy_timeout_ms: int = 5000):
        self.db_path = Path(db_path)
        self.busy_timeout_ms = busy_timeout_ms
        self._init_db()

    @contextmanager
    def _conn(self):
        conn = sqlite3.connect(
            str(self.db_path),
[...]

PR #252, opened 2026-08-29
feat: save_config accepts revision preconditions, read tools expose them (#229 phase 5)

cdelmonte-zg:
The MCP leg of the read -> mutate -> conflict-checked-save loop the web UI's forms got in phase 4 (#251), and the piece that makes the two-process scenario - a Flask server and a nanoidp-mcp companion on one config directory - safe to opt into: without it, an MCP save_config silently overwrites whatever the other process wrote since this runtime loaded. With it, #229 is complete on every write surface: primitive (phase 1), ConfigManager.save() (phase 2), YamlWriter (phase 3), web UI (phase 4), MCP (this).

**ConfigManager tracks the loaded revisions.** _stage_directory hashes exactly the bytes each YAML file's parse saw (new config_writer.revision_of_bytes, one recipe shared with current_revision; a missing file gets the empty-bytes hash so "create only if still absent" round-trips), and _commit_directory promotes them as users_revision / settings_revision. Deliberately NOT the file's hash at ask time: on a runtime that is stale against the directory (another process wrote), a fresh disk hash would satisfy the precondition exactly when the lost update is real. Every load path refreshes the pair - reload, reload_local, and the post-save refresh - so a successful save chains naturally.

**MCP surface.** list_users and get_user return users_revision (on the not-found branch too: get_user -> create_user -> save_config is "create this user only if the file still looks like it did when I saw them absent"); list_clients, get_client and get_settings return settings_revision; reload_config and a successful save_config return both, so the conflict retry loop is reload, reapply, save - no extra read call. save_config accepts optional expected_users_revision / expected_settings_revision and hands them to ConfigManager.save(), whose precondition parameters have waited for a caller since phase 2. A stale revision comes back as the already-specified {"success": false, "kind": "conflict"} with nothing written. Because save_config always writes BOTH files, the contract has exactly two modes and no third (review round 1): omitting both revisions keeps the unconditional last-write-wins default, and supplying either makes the WHOLE save conflict-checked, the omitted revision defaulting to the one this runtime was loaded from - a pass-through of a partial precondition would have guarded one file while the runtime's stale snapshot silently overwrote the other, the exact lost update the revisions exist to catch, reachable through the API's own recommended flow (the check is "is None", not falsy: an explicit empty string is a supplied, stale revision and conflicts rather than being silently replaced). No new tool: the smoke test's EXPECTED_TOOLS stays 26.

**Tests.** tests/test_config_revisions.py pins the tracking semantics: revisions match the files at load, an external write does NOT move them (that staleness is the point), reload_local and save() refresh them, a missing file matches current_revision, and the two-manager scenario - the stale runtime is refused while the unconditional default still wins. tests/test_mcp_tool_branches.py gains a TestRevisionPreconditions class driving the real dispatch path: read tools expose the loaded revisions, reload_config hands back fresh ones, a fresh-revision save succeeds and its response chains, and the real two-writer conflict (no monkeypatching) leaves both files byte-identical. Two cross pins from review round 1, one per direction: a settings-only external writer (the shape of the UI's settings form) against a users-guarded save, and a users-only writer against a settings-guarded save - each must conflict with both files untouched; both fail against the pass-through version. The five monkeypatched save fakes now accept the new kwargs, and the mock-config security fixture gets concrete revision values (the established "concrete, or it cannot serialize" pattern).

**e2e.** examples/mcp_smoke_test.py drives the revision fields and a stale save_config through the real stdio transport - deliberately only the conflict leg, which is guaranteed to write nothing, since the script may run against a checked-in config directory (the lock file it touches is already gitignored). Verified live beyond the suite: a Flask process and an MCP companion on one fresh directory - the UI created a user, the companion's save with its loaded revision was refused with kind=conflict, the UI's write was byte-intact on disk, and reload_config -> reapply -> save landed both changes. Full suite 1711 (with the two round-1 pins), mypy, ruff, lint-imports clean; examples/test_agent.py live 57/58 (the one failure is the known environmental Redirect URI Exact Match, identical on main).

**Docs.** reference/mcp.md gains a "Conflict-checked saves" section and updated tool rows; architecture.md's write-path paragraph now describes the shared compare_and_replace pipeline instead of the pre-#229 "no cross-request conflict detection (last write wins)", which phase 4 had already made stale; CHANGELOG entry under Added, and phase 2's "no caller passes a precondition revision yet" sentence corrected the same way phase 4 corrected phase 3's.

Refs #229 (closed; this is its last phase) and the #235 decision table (the MCP companion process contract).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   changes only the secret, so every field (present and future) is carried.
 
 ### Added
+- **MCP callers can make `save_config` conflict-checked** (#229 phase 5,
+  the MCP leg of the same loop the web UI's forms got in phase 4): the
+  read tools return the revision of the file the runtime was loaded
+  from (`list_users`/`get_user` carry `users_revision`;
+  `list_clients`/`get_client`/`get_settings` carry `settings_revision`),
+  and `save_config` accepts them back as `expected_users_revision` /
+  `expected_settings_revision`, refusing the save with
+  `{"success": false, "kind": "conflict"}` - nothing written - when
+  another writer (the web UI, another agent, a second nanoidp process
+  on the same directory) moved a file since. `reload_config` and a
+  successful `save_config` return fresh revisions, so the retry loop is
+  reload, reapply, save. The revision is deliberately the one the
+  runtime was LOADED from, not the file's hash at ask time: on a
+  runtime that is stale against the directory, a fresh disk hash would
+  pass the precondition exactly when the lost update is real. Omitting
+  the revisions keeps the save unconditional (last write wins), same
+  as before.
 - **Display-only `description` on users, shown in the persona login
   picker** (#244): a user in `users.yaml` can carry an optional
   `description` (max 200 characters, plain text) rendered next to the
@@ -101,8 +118,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   runtime already reflects them - only the mirror push failed. This also
   closes a gap where MCP's `save_config` tool left the process holding
   its pre-save view of anything the save's read-modify-write cycle picked
-  up from disk: it now sees the refreshed state too. No caller passes a
-  precondition revision yet; that lands with the surface that needs it.
+  up from disk: it now sees the refreshed state too. The precondition
+  revisions were unused when this entry was first written; MCP's
+  `save_config` now supplies them (see Added).
   The runtime refresh can itself fail (an in-memory value that bypassed
   field validation, since `Settings` has no `validate_assignment`, can
   reach the file and then fail to parse back in) - `save()` now tells
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,9 +59,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   reload, reapply, save. The revision is deliberately the one the
   runtime was LOADED from, not the file's hash at ask time: on a
   runtime that is stale against the directory, a fresh disk hash would
-  pass the precondition exactly when the lost update is real. Omitting
-  the revisions keeps the save unconditional (last write wins), same
-  as before.
+  pass the precondition exactly when the lost update is real. Because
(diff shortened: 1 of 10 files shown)


cdelmonte-zg:
Review round 1 (user review at 5bf5949) found a real blocker in the save_config contract: the head passed the caller's revisions straight through to ConfigManager.save(), where None leaves that file's write unconditional - so the API's own recommended flow (get_user -> mutate -> save with only expected_users_revision) guarded users.yaml while the runtime's stale snapshot silently overwrote a settings.yaml another writer had changed. Reproduced with a settings-only external edit clobbered back by a users-guarded save reporting success.

Fixed in 109f396 with the invariant the review prescribed: save_config always writes both files, so there are exactly two modes and no third - omit both revisions for the unconditional last-write-wins default, or supply either and the whole save is conflict-checked, the omitted revision defaulting to the runtime's loaded one ("is None", not falsy, so an explicit empty string conflicts instead of being silently replaced). Two regression pins added, one per direction, both failing against the pass-through version; tool description, schema, mcp.md and CHANGELOG state the two-mode contract. Also applied the non-blocking wording fix: a missing file and a zero-byte file share sha256(b""), so the missing-file comment now says "still has the missing/empty revision" instead of "create only if still absent" (phase 2 semantics, unchanged).

Suite 1711, mypy, ruff, lint-imports clean; the partial-guard repro now gets kind=conflict with the other writer's change intact.

cdelmonte-zg:
Round 2 at 109f396, verified. The blocker is closed.

save_config now has the required two-mode contract: with no supplied revisions it preserves the existing unconditional last-write-wins behaviour; supplying either revision makes the whole two-file save conflict-checked, with the omitted revision taken from the state this runtime was loaded from. That closes the cross-file lost-update case from round 1 without changing ConfigManager.save()'s more general per-file precondition contract.

The two new regression tests pin both directions of the bug: a settings-only external writer against a users-guarded save, and a users-only external writer against a settings-guarded save. Both assert kind=conflict and that neither file moved.

The missing-file wording is corrected as well: missing and zero-byte files intentionally share the empty-bytes revision.

No new blocker found in the fix. E2E and CodeQL are green on this head; merge waits for the required Tests workflow.

Issue #231, opened 2026-08-25
Parametric client contract test: one fully-populated OAuthClient through YAML round-trip, UI edit and MCP update


cdelmonte-zg:
Cheap future safety net from the 2026-08-25 assessment - explicitly NOT a registry (declined in #214) and not urgent.

The #214/#226 parity tests pin the declarative surfaces (load contract, MCP read surface and schemas, form template) and model_copy makes the rebuild legs safe by construction. What remains theoretically possible: adding a field everywhere declaratively and forgetting to actually READ it in a routes/ui.py parser or APPLY it in an MCP handler body - the parity tests cannot see imperative code.

One parametric contract test closes most of that gap: construct an OAuthClient with every field set to a non-default value, then assert the full value survives (1) a YAML round-trip through client_to_yaml/merge_client_entry + reload, (2) a UI edit posting every form field, (3) an MCP update_client call carrying every schema property. A new field with a forgotten imperative leg then fails this test the moment the parity tests force it into the surfaces.

Good first-issue material; no milestone until someone needs it.

PR #254, opened 2026-08-29
feat: public clients - token_endpoint_auth_method none with mandatory PKCE S256 (#188)

### LINKED ISSUE #188: Public clients: token_endpoint_auth_method none with mandatory PKCE S256

cdelmonte-zg:
First deliverable of milestone 2.9.0 (OAuth/MCP interoperability foundation). Closes #188.

**What.** `OAuthClient.token_endpoint_auth_method` (`client_secret_basic` default = today's behaviour | `client_secret_post` | `none`), with `client_secret` now `Optional` and a model validator requiring it for every method except `none`. A public client - CLI, desktop app, SPA, MCP client, anything that cannot keep a secret - is identified at `/token` by `client_id` alone; a presented secret is ignored and can never become a credential (`check_client` skips public clients, so a stored-but-ignored secret does not authenticate). The protections that stand in for client authentication are enforced regardless of profile:

- `/authorize` requires PKCE with `S256` (OAuth 2.1 §7.5.1): missing challenge, `plain`, and the omitted-method-defaults-to-plain case (RFC 7636 §4.3) are all rejected even with `require_pkce` off.
- `client_credentials` is refused with `unauthorized_client` (RFC 6749 §5.2 JSON): the grant IS client authentication.
- Refresh tokens always rotate with reuse detection (OAuth 2.1 §4.3.1/§6.1), whatever `refresh_token_rotation` says - pinned by a replay test with rotation off in settings.
- `/revoke` accepts a public `client_id` with an ownership check: the token's own `client_id` claim must match, otherwise the response is still `200` and nothing is revoked (RFC 7009 §2.1; the silent-200 arm follows the RFC's privacy guidance - the endpoint must not become an oracle for token validity or ownership, which is exactly what a wrong-owner error would leak to an unauthenticated caller).
- `/introspect` deliberately stays authenticated (RFC 7662 §2.1) and its discovery list does NOT gain `none`; the token and revocation lists do.

**Two adjacent corrections** the new field made visible, folded in because they share the same auth-gate code:

1. **`client_secret_post` is now actually validated** at `/token`, `/introspect`, `/revoke` and `/device_authorization`. Discovery has advertised it on all three `*_auth_methods_supported` lists forever, but no endpoint ever read `client_secret` from the body - it was silently ignored. Behaviour note: a confidential client that was sending a *wrong* body secret (and succeeding, because nothing looked at it) now gets the `invalid_client`/401 it always should have; a client sending the right secret keeps working. Flagged in the CHANGELOG.
2. **Access tokens now carry a `client_id` claim** (RFC 9068 §2.2 requires it of JWT access tokens), binding them to the client they were issued to - refresh tokens have had the same claim since #56. This is what the revoke ownership check reads, and #187 (resource indicators) will want it anyway.

**Deliberately out of scope**: device flow for public clients. RFC 8628 follows the client's registered auth method, but extending `/device_authorization` to `none` clients is its own protocol decision - the endpoint still authenticates (Basic, or now post), and a comment in the code says so. Tracked as a separate 2.9.0 follow-up in #255.

**Declarative legs, all 13 points**: `ClientEntry` (empty YAML secret becomes `None`, valid only for `none`), `client_to_yaml`/`merge_client_entry` (a secret-less client drops the key - never writes the string "None" - and the method is written only when non-default), UI form (method select with method-aware secret requirement, server-side validated; small JS lifts the HTML `required` for `none`), MCP `create_client`/`update_client` schemas and handlers (`client_secret` out of the required list, validated per method in the handler; the method/secret combination is pre-checked BEFORE any assignment so `validate_assignment` can never reject mid-sequence and leave the live client half-updated, and assignment order handles both switch directions), `_client_to_dict`, regenerated `docs/schema/config.v1.json`, configuration reference docs, CHANGELOG.

**Tests**: `tests/test_public_clients.py` (26 tests covering the whole behavioural surface above, negatives included), MCP tool tests for the create/update secret/method interplay (including the half-update guard and the none->basic-without-secret refusal), contract fixtures carry the new field (FULL_B is now a `none` client, so all three legs round-trip it). The #231 parametric contract test caught the missing legs exactly as designed before they were written.

**e2e**: `test_agent.py` gains `test_public_client_flow` - creates a public client through the real UI form (no secret), is refused `/authorize` without PKCE, completes the S256 code flow with `client_id` alone at `/token`, is refused `client_credentials` with `unauthorized_client`, cleans up. Verified live against a fresh-init server from this branch: **58/59**, the one failure being the known environmental Redirect URI Exact Match baseline (identical on main).

The enforcement of `token_endpoint_auth_method` is at `/token` (the field names the token endpoint); `/introspect`, `/revoke` and `/device_authorization` accept a confidential secret over either channel. There is no client-management REST surface today (`/api/config` exposes only `clients_count`), so this ships across settings.yaml, the UI form and the MCP tools; the MCP create/update tools drop a secret on a public client the same way the UI does, so a `none` client is never persisted with a dead secret on any surface.

Full suite **1755**, mypy, ruff, lint-imports clean.

**Merge timing**: after 2.8.0 final - this is 2.9.0 material; the branch is on current main (`a44133a`) and does not touch the rc2 soak.


cdelmonte-zg:
Round 1 (your review at f6cbbb9): both blockers fixed in 3cb7b5c, each reproduced first.

**Blocking 1 (forged revocation).** Reproduced exactly as you described: a public client submitting a JWT signed with its own key, carrying its own client_id but a foreign jti, revoked another client's real token. Fixed by verifying the token's signature (crypto.verify_jwt) BEFORE any claim is trusted or the store is mutated - a token that fails verification revokes nothing and still returns 200 (RFC 7009 privacy). The property holds for confidential clients too, since verification is now unconditional. Pinned by test_forged_ownership_claim_cannot_revoke_a_foreign_token: mint a confidential token, forge a JWT with its jti + the public client_id, assert the real token stays active.

**Blocking 2 (auth boundary).** You were right on both halves - the method was descriptive, not enforced, and authorization_code was exempt. Replaced the grant-specific checks with one boundary (_enforce_token_endpoint_auth) applied before dispatch for every grant, with exactly the semantics you sketched: none = client_id alone; client_secret_basic = Basic required, body secret rejected; client_secret_post = body required, Basic rejected; confidential clients authenticate on every grant, authorization_code included. The model docstring's 'either channel' language is gone. Negatives added per your list: basic+body -> reject, post+Basic -> reject, confidential authorization_code with client_id only -> invalid_client. The one pre-existing test that encoded the exempted code exchange (test_pkce_plain_flow_no_auth_header) is updated and dropped as a duplicate of the authenticated plain-flow test.

**Non-blocking (schema enum).** Not addressed this round - kept ClientEntry.token_endpoint_auth_method a plain str so an invalid value fails with OAuthClient's own Literal error, consistent with the color fields; the generated schema stays type:string. Happy to tighten it to an enum in the schema if you'd prefer, say the word.

Suite 1747, mypy/ruff/lint-imports clean. Live e2e 58/59 (the one failure is the known environmental Redirect URI Exact Match, identical on main). CI running.

cdelmonte-zg:
A /code-review pass on the round-1 fix (50a5c5b) turned up one crash I had missed plus a few public-client UX gaps. All verified before fixing.

**Finding 1 (crash, the important one).** clients.html sliced client_secret[:8] unconditionally, which is None for a public client -> the /clients admin page 500'd, and the create flow redirects straight to it, so creating a public client broke the page immediately. My public-client tests never rendered /clients, so this slipped through. Fixed: the secret cell shows a 'public (none)' badge for public clients, '-' for an empty secret. Pinned by a test that fails without the template guard.

**Finding 3.** Editing a confidential client to 'none' with a blank secret field kept the old secret (blank = unchanged), persisting a public client with a dead, ignored secret in YAML. The edit handler now drops the secret to None when the method is 'none'.

**Finding 4.** The Regenerate Secret button is hidden for public clients, and the route guards a direct POST - model_copy(update=) bypasses the validator and would otherwise stamp a secret onto a 'none' client.

**Finding 2 (docs overclaim - you'd have caught this).** Only /token enforces the registered method; /introspect, /revoke and /device_authorization accept a confidential secret over either channel. Rather than extend enforcement to three resource-server-facing endpoints (token_endpoint_auth_method names the token endpoint), I scoped the docs/CHANGELOG claim to /token and stated the other three stay lenient. Say the word if you'd rather I enforce it everywhere.

**Finding 6.** Migration note added to the CHANGELOG: a confidential authorization_code + PKCE client must present its secret or re-register as 'none'.

**Finding 5 (revoke narrowing) accepted by design.** Verifying the token before trusting its claims necessarily means an expired or wrong-audience token revokes nothing, and the no-jti hash fallback protected nothing (our tokens always carry jti; a forged no-jti token must not be revocable). That narrowing is the security fix, not a regression - access/refresh revocation (aud=settings.audience) still works, covered by the existing tests.

Three UI regression tests added. Suite 1750, mypy/ruff/lint-imports clean, both original blockers still reproduce as fixed. CI running.

cdelmonte-zg:
Round 2 (your review at 50a5c5b): both items fixed in f1920d5.

**Blocking (two auth methods in one request).** Reproduced: valid Basic + a body client_secret (correct or wrong) returned 200 because the basic branch only rejected a body secret when Basic was absent. Fixed - the client_secret_basic branch now rejects a body client_secret whenever present, with or without Basic (RFC 6749 §2.3). Pinned with valid Basic + body secret, both a correct and a wrong body value -> 401. The client_secret_post side was already closed (any Basic rejects it).

**UI create-to-none.** You're right, it was inconsistent with the edit-to-none fix. The create route now normalizes client_secret to None server-side when the method is 'none', not relying on the form JS (which only lifts 'required'). Test: create with 'none' + a non-empty submitted secret asserts the persisted secret is None. This also makes the e2e claim ('creates a public client through the real UI form, no secret') true for a real browser, not just the agent's secret-less POST.

**Non-blockers, as you framed them:**
- Schema enum: still a plain str on ClientEntry, unchanged - happy to tighten if you want it.
- #188 mentions /api but there's no client-management REST surface today (/api/config exposes clients_count only). Agreed not to invent one for this PR; I'll fix the issue wording rather than the code.
- Device flow for public clients: opened as #255 (milestone 2.9.0, lower priority than #187/#189/#191) so it's tracked rather than an open-ended out-of-scope note.

Suite 1752, mypy/ruff/lint-imports clean. CI running.

cdelmonte-zg:
Round 3 at f1920d5, verified. This closes it.

The remaining authentication-boundary case is fixed correctly: a client_secret_basic client now rejects a body client_secret unconditionally, including when valid Basic credentials are present, so two authentication methods can no longer be silently accepted in the same /token request. The regression test covers both a matching and a contradictory body secret.

The create-to-none UI path is fixed at the right boundary as well: the server normalizes the secret to None for a public client regardless of the pre-generated value submitted by the browser, matching the edit-to-none behaviour and keeping persisted public clients genuinely secret-less.

I rechecked the earlier findings after these changes: the verified-token revocation ownership check remains intact, the single token-endpoint authentication boundary still applies before every grant dispatch, registered client_secret_basic / client_secret_post methods remain enforced, and the public-client PKCE, refresh-rotation and client-credentials restrictions are unchanged.

The deliberately deferred public-client device flow is now tracked separately in #255, which is the right boundary for this PR.

CI is fully green on this head. I do not see anything else that should block #254.

Approving. Merge timing after the 2.8.0 final remains appropriate.

cdelmonte-zg:
One more, from a self-audit after the round-3 approval (88dd2c0) - flagging it rather than sitting on it.

The UI create-to-none and edit-to-none fixes dropped the dead secret on a public client, but the **MCP** create_client/update_client tools did not - the 'always sync MCP + UI' rule ties them to the same behaviour. Reproduced: create_client with token_endpoint_auth_method 'none' + a client_secret persisted the secret; update_client switching to 'none' kept the old one; a secret supplied to an already-public client got stored. Not a security hole (check_client skips public clients), but it writes a dead, ignored secret into settings.yaml and diverges from the UI.

Fixed: create_client forces client_secret=None for 'none'; update_client keys off the effective method - a public target is always secret-less, a confidential target still sets the secret before flipping the method (validate_assignment ordering). Three MCP tests added; the switch-none-back-to-confidential path stays green.

This surfaced a coupled-field issue in the #231 contract test: FULL_B was a 'none' client carrying secret-b (an incoherent state), so once MCP normalized it away the round-trip diverged from the YAML leg. FULL_B is now a coherent public client (no secret), which also gives the contract suite a 'none' round-trip across all three legs; client_secret is excluded from the non-default fixture check with a comment, since it is coupled to token_endpoint_auth_method (a public client has no secret) and can't be non-default on a fixture that exercises 'none'.

Suite 1755, mypy/ruff/lint-imports clean. Your round-3 approval stands on the auth/security surface; this is the MCP-parity tail of the same public-client work. CI running.

cdelmonte-zg:
Rechecked on the new head 88dd2c0 after the MCP parity self-audit.

The additional change is correct. create_client now normalizes a public client to client_secret=None even when a secret is supplied, and update_client consistently keys the assignment order off the effective target method: public targets become secret-less, while confidential targets still receive the secret before the method switch so validate_assignment cannot reject the intermediate state.

The three new MCP regressions cover the relevant shapes: create none with a supplied secret, switch a confidential client to none, and supply a secret while updating an already-public client.

The contract-fixture adjustment is also sound: FULL_B now represents a coherent public client rather than none plus a dead secret, while the suite still exercises the client_secret distinction across the fixtures.

I rechecked the earlier security/authentication findings after this change and none have been reopened.

No new blocker from this head. The round-3 approval therefore still stands for the current code. Merge remains gated on the final required CI state, and is still intentionally scheduled after the 2.8.0 final. (PR body refreshed to the current state - 1755 tests, #255 for the device flow, the no-REST-surface note.)

cdelmonte-zg:
The two non-blockers are now done (df027fc), since the blocking work settled.

**Schema enum.** ClientEntry.token_endpoint_auth_method is now the closed Literal instead of a plain str, so the generated config.v1.json carries the three-value enum, and an invalid value is rejected at document-model load with a field path (oauth.clients[N].token_endpoint_auth_method) rather than only in to_client(). The type: ignore is gone; a regression test loads an unknown method and asserts the failure names the field.

**Guide.** token-requests.md gains a 'Public clients (no secret)' section: the /token authorization_code exchange for a 'none' client - client_id in the body with the PKCE code_verifier, no -u - plus the PKCE-S256-mandatory / no-client_credentials / forced-rotation notes. The browser /authorize step is described rather than scripted, since curl can't drive the login.

I also swept the rest for stale docs and other access points while here: the token-requests curl examples all use -u (HTTP Basic), still correct for the confidential demo-client; configuration.md is the only page that enumerates client fields and is current; README and mcp.md carry no client-field lists to update; config_validation accepts public clients via the model validator; and no other .client_secret usage assumes non-None (clients.html was the only one). The interactive first-run wizard still creates confidential clients only - left as a possible future enhancement, not a defect.

Suite 1756, mypy/ruff/lint-imports clean. CI running.

Issue #189, opened 2026-08-23
RFC 9207: iss on the authorization response


cdelmonte-zg:
**Why.** The MCP 2026-07-28 revision recommends that the authorization server returns `iss` in the authorization response and that clients validate it to prevent authorization-server mix-up (RFC 9207). nanoidp's redirect carries only `code` and `state` (`routes/oauth.py`, the `state=` append). Small, precise, exactly nanoidp's kind of feature, and it interacts with `issuer_from_request` (#126): the value must be the effective issuer of that request.

**Scope.**
- Append `iss=<effective issuer>` to every authorization response that is actually delivered to the client through a validated `redirect_uri`, success and error alike (RFC 9207 section 2). Errors that are rendered locally because the `redirect_uri` or `client_id` could not be validated stay local: `iss` is never a reason to redirect to an unvalidated URI. Device flow unaffected.
- RFC 9207 requires the issuer to be an `https` URL without query or fragment, while nanoidp usually runs on `http://localhost:8000`. Decision: `authorization_response_iss_parameter_supported: true` is advertised only when the effective issuer meets those requirements; with an `http` issuer the parameter is still appended (harmless, useful for testing) but not advertised, and the docs say why. No silent dev relaxation in metadata.
- Tests: static issuer, reflected issuer behind a proxy, explicit issuer allowlist; the e2e agent validates `iss` against discovery.

cdelmonte-zg:
Implemented and landed on `main` via #268 (the 2.9 stack integration merge, sha 73ad177). Ships in 3.0.0. Closing.

PR #256, opened 2026-08-29
feat: RFC 8707 resource indicators bind the access token audience (#187)

### LINKED ISSUE #187: RFC 8707 Resource Indicators: resource binds the access token audience

cdelmonte-zg:
Second deliverable of milestone 2.9.0 (OAuth/MCP interoperability foundation). Closes #187. **Stacked on #254 (#188)** - the base of this PR is `feat/188-public-clients`, not main, because both touch the token issuance path and #187 builds on the access-token `client_id` claim #254 added. Review/merge order: #254 first, then this rebases onto main. The diff shown against `feat/188-public-clients` is just this PR's ~20 files.

**What.** A client may send one or more `resource` parameters on `/authorize`, `/token` (every grant) and `/device_authorization`; the access token `aud` becomes those resources (a plain string for one, an array for several) instead of the global `oauth.audience`, so a token minted for MCP server A is rejected by server B - and a wrong-audience test can finally be written. `services/resource.py` validates a resource indicator (an absolute URI without a fragment, RFC 8707 §2) and resolves a request against the client's allow-list and any prior binding, mirroring `services/scope.py`; an invalid or disallowed resource is `invalid_target`.

**Contract details:**
- Per-client `allowed_resources` (empty = any valid resource, the dev default, same "empty = unrestricted" convention as `allowed_scopes`) gates which resources a client may target.
- The authorization code and the refresh token remember the bound resources; a `/token` request may narrow them to a subset but never widen (RFC 8707 §2). No `resource` inherits what the prior step bound.
- Sending no `resource` leaves `aud` at `oauth.audience` - **no behaviour change for existing clients**, pinned by a test.
- `/introspect` reports the token's `aud`. To do that for a resource-bound token, `crypto.verify_jwt` gained `audience=None` (verify signature + expiry, not the audience); `/introspect` and `/revoke` use it so a resource-bound token can be introspected and revoked. **`/userinfo` deliberately keeps requiring the OP audience** (`oauth.audience`): it is the OP's own protected resource, and a token audienced to an MCP server belongs at that server (used against `/introspect` by the server), not here - so the existing wrong-audience security test stays valid.
- No `resource_indicators_supported` discovery metadata: RFC 8707 defines none and nanoidp does not invent metadata.

**Profile decision** (the issue left it open): no profile *requires* a `resource` to be sent. RFC 8707 makes the parameter optional; the MCP requirement to send it is an obligation on the client, not the AS. Requiring it would break existing clients and the dev default. When a client *does* send one, `allowed_resources` (if set) is enforced regardless of profile.

**Declarative legs, all of them:** `ClientEntry` + coercion, `client_to_yaml`/`merge_client_entry`, the UI client form (Allowed Resources textarea), MCP `create_client`/`update_client` schemas and handlers, `_client_to_dict`, regenerated `config.v1.json`, configuration reference, the token-requests guide, CHANGELOG.

**Tests:** `tests/test_resource_indicators.py` (22) - indicator syntax, allow-list gate, narrowing; client_credentials aud binding including the multi-resource array and the unchanged default; the authorization-code flow's bind / narrow-to-subset / widen-rejected; refresh keeps and narrows the aud, and cannot widen; `/introspect` reports the resource aud; the device flow binds and rejects an invalid resource. Contract fixtures carry `allowed_resources` (the #231 parametric test flagged the missing legs before I wrote them); the MCP update round-trips it. **e2e:** `test_agent.py` gains `test_resource_indicators` - resource binds the aud, `/introspect` reports it, an invalid resource is `invalid_target` - verified live **59/60** (the one failure is the known environmental Redirect URI Exact Match baseline).

Suite **1780**, mypy, ruff, lint-imports clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   changes only the secret, so every field (present and future) is carried.
 
 ### Added
+- **RFC 8707 Resource Indicators: `resource` binds the access token
+  audience** (#187). A client may send one or more `resource` parameters
+  on `/authorize`, `/token` (every grant) and `/device_authorization`;
+  the access token's `aud` is then those resources (a plain string for
+  one, an array for several) instead of the global `oauth.audience`, so a
+  token minted for one MCP server is rejected by another and a
+  wrong-audience test can finally be written. A `resource` must be an
+  absolute URI without a fragment or the request is `invalid_target`
+  (RFC 8707 §2). New per-client `allowed_resources` gates which resources
+  a client may target (empty = any valid resource, the dev default, same
+  "empty = unrestricted" convention as `allowed_scopes`). The
+  authorization code and refresh token remember the bound resources; a
+  `/token` request may narrow them to a subset but never widen them.
+  Sending no `resource` leaves `aud` at `oauth.audience` - **no change
+  for existing clients**. `/introspect` reports the token's `aud`, and
+  now verifies a token's signature without pinning its audience (so a
+  resource-bound token can be introspected and revoked); `/userinfo`
+  still requires the OP audience. No `resource_indicators_supported`
+  discovery metadata (RFC 8707 defines none). Full support across
+  settings.yaml (`allowed_resources`), the UI client form and the MCP
+  `create_client`/`update_client` tools.
 - **Public clients: `token_endpoint_auth_method: "none"` with mandatory
   PKCE S256** (#188). A client that cannot keep a secret (CLI, desktop
   app, SPA, MCP client) can now be declared with
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   "empty = unrestricted" convention as `allowed_scopes`). The
   authorization code and refresh token remember the bound resources; a
   `/token` request may narrow them to a subset but never widen them.
+  Narrowing the access token does not narrow the refresh token, which keeps
+  the full original grant so a later refresh can still request any resource
+  the authorization covered (RFC 8707 §2.2).
   Sending no `resource` leaves `aud` at `oauth.audience` - **no change
   for existing clients**. `/introspect` reports the token's `aud`, and
   now verifies a token's signature without pinning its audience (so a
--- a/book/src/guides/token-requests.md
+++ b/book/src/guides/token-requests.md
@@ -74,6 +74,27 @@ without it), cannot use the `client_credentials` grant, and always gets a
 rotating refresh token. A `client_secret` sent by a public client is
 ignored.
 
+## Resource indicators (RFC 8707)
+
+To bind the access token to a specific resource server - so a token for
+MCP server A is rejected by server B - send a `resource` parameter (an
+absolute URI without a fragment). An MCP client sends it on both
+`/authorize` and `/token`; the access token's `aud` becomes that
+resource. `resource` is repeatable for more than one target, and a
+`/token` request may narrow the resources an authorization code bound but
+never widen them. A resource outside the client's `allowed_resources` (or
+not a valid URI) is rejected with `invalid_target`.
+
+```bash
+curl -X POST 'http://localhost:8000/token' \
+  -u 'demo-client:demo-secret' \
+  -d 'grant_type=client_credentials' \
+  -d 'resource=https://mcp.example/server'
+# the returned access token's aud is https://mcp.example/server
+```
+
+Sending no `resource` leaves the audience at `oauth.audience`, unchanged.
+
 ## Device authorization flow
 
 ```bash
--- a/book/src/reference/configuration.md
+++ b/book/src/reference/configuration.md
@@ -290,6 +290,11 @@ oauth:
     - client_id: "cli-client"
       token_endpoint_auth_method: "none"  # public client (#188): no secret
       description: "CLI / SPA / MCP client that cannot keep a secret"
+    - client_id: "mcp-client"
+      client_secret: "secret"
+      description: "Client bound to specific MCP servers (RFC 8707, #187)"
+      allowed_resources:        # optional; see "Resource indicators" below
+        - "https://mcp.example/server"
   # logos_dir: "./static/logos"    # optional; defaults to src/nanoidp/static/logos
   # scopes_supported:               # optional; the global scope vocabulary
   #   - openid                      # (default: openid, profile, email, offline_access)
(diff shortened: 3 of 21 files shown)


cdelmonte-zg:
A /code-review pass on the stack turned up five findings; four fixed in cb4c815, the fifth needs no code.

**Finding 1 (contract, the real one).** A grant with a prior authorization step whose code / refresh token / device code bound NO resource could still have one introduced at /token, binding the aud to a resource /authorize never authorized - widening from nothing, against the 'narrow, never widen' contract. Reproduced: a no-resource authorization code + a /token resource bound the token. Fixed: _resolve_token_resource now distinguishes None (no prior step - client_credentials/password, gated only by allowed_resources) from [] (a prior step bound nothing, so none may be introduced); the authorization_code/refresh/device grants pass the stored resources or []. Pinned.

**Finding 2.** /device_authorization stored the raw request resources, so resource=a&resource=a survived into the token aud as ['a','a']. Now stores resolve_resources' de-duplicated granted list; a repeated resource collapses to a single string aud. Pinned.

**Finding 4 (RFC 7662 §2.2).** /introspect reported the caller's client_id, not the client the token was issued to. Now returns the token's own client_id claim (the one #188 added), falling back to the caller only for a legacy token without it - a genuine correctness improvement the claim newly enables. Pinned with a resource server introspecting another client's token.

**Finding 3.** /revoke logged 'revoked: true' even when the verified token had no jti (nothing revoked). Audit now fires only inside the jti branch; a verified nanoidp token always carries one, so this just tightens the edge case.

**Finding 5** (confidential authorization_code now requires client auth) is the intended #188 change (RFC 6749 §3.2.1) and is already in the CHANGELOG migration note - no code change.

Note these findings span the whole stack (the review ran main...HEAD, so it saw #254 too): 3 and 4 touch the revoke/introspect blocks #254 introduced, fixed here since #187 builds on them. Suite 1783, mypy/ruff/lint-imports clean. CI now running (the PR base is stack/188-public-clients, which the workflows' branch filter includes).

cdelmonte-zg:
Review round 2: all three findings fixed in bdb1657, each reproduced first.

**Blocking (RFC 8707 §2.2 - refresh token narrowed with the access token).** You were exactly right. authorize [A,B] + token resource=A left the refresh token remembering only [A], so a later refresh for B (part of the original grant) was invalid_target. Fixed by separating the two concepts you named: _GrantOutcome now carries resource (the access token aud, the narrowed subset) and refresh_resource (the full original grant persisted on the refresh token); create_token uses each for its own token. authorization_code/refresh/device pass the code's / refresh token's / device code's full set as refresh_resource; password (no prior step) falls back to the access-token resources. Pinned exactly as you specified: authorize [A,B], narrow to A (AT aud=A), refresh to B succeeds with AT aud=B.

**is_valid_resource_indicator (500 + empty fragment + permissive parse).** Now charset-checks against the RFC 3986 set first (rejects the raw space in https://exa mple/resource), rejects any '#' (so the empty-fragment https://example.com/# is out, not just a non-empty fragment), and wraps urlparse in try/except (https://[bad is invalid_target, never a 500). Pinned for all three shapes.

**Duplicate resource deduped everywhere except /authorize.** Correct - I fixed the device entry point last round and missed this one. _validate_authorize_resources discarded resolve_resources' granted list; it now writes the de-duplicated list back onto the params, so the authorization code stores ['A'] not ['A','A']. Pinned (authorize resource=A&resource=A, exchange without resource, aud == 'A').

Suite 1788, mypy/ruff/lint-imports clean. CI running.

cdelmonte-zg:
Two things on this head (f91404e).

**Review round 3 (RFC 3986 validator).** Fixed in f91404e. The charset whitelist admitted a bare '%' and never validated the port, so is_valid_resource_indicator accepted '%ZZ', a trailing '%', and ':abc'. It now rejects a '%' not followed by two hex digits (RFC 3986 §2.1 pct-encoded) and reads urlparse's .port inside the try/except so a non-numeric port is invalid_target, not a deferred crash. Tests: %ZZ, trailing %, :abc rejected; %2F and :8443 accepted. You're right that a global charset whitelist isn't the full per-component grammar - I kept the targeted fixes rather than pull in a URI validator, since the remaining gap is standards-completeness, not an audience bypass; happy to go further if you'd prefer a real validator.

**Your earlier question - other indirect consumers of resource-bound tokens** (2dd2279). I swept every verifier and the MCP token tools. One was broken, one was a gap:
- **MCP verify_token** verified against oauth.audience, so it reported a resource-bound token (aud = a resource) as invalid. It now takes an optional 'audience': omitted -> verify signature+expiry only and return claims (a resource token isn't falsely invalid); provided -> also require aud to match, which is how an MCP client simulates a resource server accepting a token for itself and rejecting one for another - the #187 test scenario, now doable entirely through MCP.
- **MCP generate_token** gained a 'resource' argument that binds the aud the same way /token does.
- Confirmed correct as-is: /userinfo and the refresh grant keep pinning oauth.audience (resource tokens belong at their resource server; refresh tokens keep that aud); /introspect and /revoke verify without pinning; decode_token never verified. So no other endpoint or tool misjudges a resource-bound token.

Suite 1794, mypy/ruff/lint-imports clean. CI running.

cdelmonte-zg:
Round 3 verified: the RFC 3986 validator fix does the two things asked - rejects a '%' not followed by two hex digits and reads parsed.port so ':abc' is invalid_target. Tests cover %ZZ, trailing %, non-numeric port and the valid %2F / :8443 counterparts.

Maintainer review: **LGTM**, merge after CI is fully green. All the substantive findings from the earlier passes - the refresh-grant narrowing (RFC 8707 §2.2), the /authorize duplicate resource, the urlparse crash, the empty fragment, the percent-encoding, and the MCP verify_token/generate_token consumers - are fixed with regression tests.

The one remaining nit - _RFC3986_CHARS is a global whitelist, not the full per-component ABNF (e.g. it would accept https://example.com/a[b], where '[' ']' aren't valid path pchar) - is standards-completeness, not an audience bypass, and is tracked as a separate refactor in #257 rather than growing the regex further in this PR.

Merge timing unchanged: this is 2.9.0 material and stacked on #254 (#188), so it merges after #254 lands and 2.8.0 final is cut, rebased onto main.

cdelmonte-zg:
Post-LGTM update: a /code-review pass on the full #187-#189 stack found two real bugs in this PR's code, now fixed in 36f4f47 (added on top of the f91404e you LGTM'd, since #256 hasn't merged).

1. **Charset regex anchored with '$'** - in Python '$' also matches just before a trailing newline, so a resource ending in an (encoded) newline slipped past the control-character guard into the token aud. Anchored with '\\Z' now; two newline cases pinned.
2. **Refresh didn't re-validate inherited resources** - a request that omitted 'resource' returned the prior step's bound set verbatim, without re-checking the client's CURRENT allowed_resources, so a resource removed from the allow-list after the grant was still minted on a later refresh. It now re-validates through resolve_resources (the scope path already does this with validate_only, #186 B1); a now-disallowed inherited resource is invalid_target. Pinned with a refresh after the admin narrows the allow-list.

Suite 1797, mypy/ruff clean. The stack tip (#258) is rebased onto this head. CI re-running on #256.

cdelmonte-zg:

Issue #40, opened 2026-06-11
MCP get_oidc_discovery is out of sync with the HTTP discovery document (missing claims_supported/azp and more)


cdelmonte-zg:
## Problem

The MCP tool `get_oidc_discovery` (`src/nanoidp/mcp_server.py:817-837`) returns a hand-built subset of the discovery metadata. Compared to the HTTP endpoint `/.well-known/openid-configuration` (`src/nanoidp/routes/oauth.py:28-66`) it is missing:

- `claims_supported` — including `azp`, which #37 explicitly added to the HTTP document
- `response_types_supported`
- `id_token_signing_alg_values_supported`
- `code_challenge_methods_supported`
- `token_endpoint_auth_methods_supported`
- `introspection_endpoint_auth_methods_supported`
- `revocation_endpoint_auth_methods_supported`

Agents using the MCP server therefore cannot discover the `azp` claim or PKCE support, and the two documents will keep drifting because the metadata is duplicated.

## Suggested fix

Build the discovery document in one shared helper (e.g. in a service module) and have both the HTTP route and the MCP tool call it, so they can never diverge again.

## Acceptance criteria

- [ ] MCP `get_oidc_discovery` output matches the HTTP document field-for-field
- [ ] Single source of truth for the metadata (no duplicated dict)
- [ ] Test asserting MCP output == HTTP output
- [ ] `examples/test_agent.py` updated in the same PR

PR #258, opened 2026-08-29
feat: RFC 9207 iss on the authorization response (#189)

### LINKED ISSUE #189: RFC 9207: iss on the authorization response

cdelmonte-zg:
Third deliverable of milestone 2.9.0 (OAuth/MCP interoperability foundation). Closes #189. **Stacked on #187 (#256) -> #188 (#254).** Base is `stack/187-resource-indicators` so CI runs and the diff stays this PR's own. Review/merge order: #254, then #256, then this, each rebased onto main in turn.

**What.** `/authorize` returns `iss=<effective issuer>` on every response delivered through a validated `redirect_uri` - success **and** error - so a client can detect an authorization-server mix-up (RFC 9207, recommended by MCP 2026-07-28). The value is `effective_issuer(settings)`, so it follows `issuer_from_request` (#126): a request reflected behind a proxy gets the matching `iss`, one outside the allowlist falls back to the fixed issuer.

**One condition drives metadata and behaviour** (a key review correction). `iss` is emitted exactly when discovery advertises `authorization_response_iss_parameter_supported`, both governed by `discovery.issuer_qualifies_for_iss_parameter` - a query/fragment-free `https` URL with a host, per RFC 9207 §2 / RFC 8414. So the default `http://localhost:8000` dev issuer sends **no** `iss` and advertises **false** (they can never disagree); point the issuer at `https` - directly or reflected via `issuer_from_request` behind a TLS proxy - to turn RFC 9207 on. The MCP `get_oidc_discovery` tool builds the same document (#40), so it carries the same value.

**Error responses.** Completing RFC 9207's "error responses too": once the `redirect_uri` is validated, `unsupported_response_type`, `invalid_scope`, PKCE errors and `invalid_target` are OAuth error redirects to the client carrying `error`, `error_description`, `state` and `iss` (RFC 6749 §4.1.2.1), not a local JSON 400. The validator order is now `client lookup -> redirect_uri -> response_type -> scope -> PKCE -> resources`, so every error past the `redirect_uri` step can redirect; errors before it (unknown client, missing/malformed/unregistered `redirect_uri`) stay local JSON - `iss` is never a reason to redirect to an unvalidated URI. **This is a behaviour change** (CHANGELOGed): the PKCE/scope/response_type tests across several files now assert the redirect + `error` param, via a new `conftest.authorize_error` helper. The device flow is unaffected.

**One authorization-response-URI builder.** `redirect_uri.append_authorization_params` is the single place that builds a success or error response URI. It retains a `redirect_uri`'s own query byte-for-byte (RFC 6749 §3.1.2: the query MUST be kept) - a plain append, no parse/re-encode that would turn `%20` into `+` - so `.../cb?tenant=foo` + response params becomes `.../cb?tenant=foo&code=...`, never a second `?` folding into the last value. The success path shared the same latent bug and now shares the fix.

**Tests.** `tests/test_iss_authorization_response.py`: `iss` present-and-matching on an https issuer, absent on http, with metadata agreeing; reflects `issuer_from_request`; the qualification helper's edges (bare `?`/`#`, no host, malformed IPv6, non-numeric port); error redirects for `unsupported_response_type` (valid redirect_uri) and `invalid_scope`, staying local without a valid redirect_uri; success and error preserving an existing query, plus a byte-for-byte unit test on the builder; MCP discovery carries the metadata. **e2e:** `test_agent` captures the discovery issuer BEFORE the flow (the value a client compares against for the anti-mix-up property) and asserts `iss` present-and-matching iff advertised, absent otherwise; the oauth21 suite's strictness check now expects the PKCE error redirect on a registered client.

Suite **1812**, mypy, ruff, lint-imports clean; live 59/60 (Redirect URI Exact Match baseline) + oauth21 3/3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,19 +46,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - **RFC 9207: `iss` on the authorization response** (#189). `/authorize`
-  now appends `iss=<effective issuer>` to the success redirect, so a
-  client can detect an authorization-server mix-up (MCP 2026-07-28
-  recommends this). The value is the per-request effective issuer, so it
-  stays correct under `issuer_from_request` (#126). RFC 9207 requires an
-  `https` issuer with no query or fragment, so
-  `authorization_response_iss_parameter_supported` is advertised in
-  discovery only when the effective issuer meets that - an `http://localhost`
-  dev issuer still gets the parameter appended (useful for testing) but is
-  not advertised as compliant, no silent dev relaxation in the metadata.
-  nanoidp renders authorization errors locally as JSON rather than
-  redirecting them, so `iss` rides only on the success redirect and is
-  never a reason to redirect to an unvalidated URI. The device flow is
-  unaffected.
+  returns `iss=<effective issuer>` on every response delivered through a
+  validated `redirect_uri` - success and error alike - so a client can
+  detect an authorization-server mix-up (MCP 2026-07-28 recommends this).
+  The value is the per-request effective issuer, so it stays correct under
+  `issuer_from_request` (#126). `iss` is delivered exactly when discovery
+  advertises `authorization_response_iss_parameter_supported`: one
+  condition drives both, so metadata and behaviour never disagree. RFC
+  9207 requires an `https` issuer with a host and no query or fragment, so
+  the default `http://localhost:8000` sends no `iss` and advertises
+  `false`; point the issuer at `https` (directly or reflected via
+  `issuer_from_request` behind a TLS proxy) to turn RFC 9207 on. **Related
+  behaviour change**: authorization errors that occur after the
+  `redirect_uri` is validated (invalid_scope, PKCE errors, invalid_target)
+  are now OAuth error redirects to the client (`error`,
+  `error_description`, `state`, `iss`) instead of a local JSON 400, per
+  RFC 6749 §4.1.2.1 - completing the RFC 9207 "error responses too"
+  requirement. Errors before the `redirect_uri` is trusted (unknown
+  client, malformed or unregistered `redirect_uri`) stay local JSON. The
+  device flow is unaffected.
 - **RFC 8707 Resource Indicators: `resource` binds the access token
   audience** (#187). A client may send one or more `resource` parameters
   on `/authorize`, `/token` (every grant) and `/device_authorization`;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   changes only the secret, so every field (present and future) is carried.
 
 ### Added
+- **RFC 9207: `iss` on the authorization response** (#189). `/authorize`
+  now appends `iss=<effective issuer>` to the success redirect, so a
+  client can detect an authorization-server mix-up (MCP 2026-07-28
+  recommends this). The value is the per-request effective issuer, so it
+  stays correct under `issuer_from_request` (#126). RFC 9207 requires an
+  `https` issuer with no query or fragment, so
+  `authorization_response_iss_parameter_supported` is advertised in
+  discovery only when the effective issuer meets that - an `http://localhost`
+  dev issuer still gets the parameter appended (useful for testing) but is
+  not advertised as compliant, no silent dev relaxation in the metadata.
+  nanoidp renders authorization errors locally as JSON rather than
+  redirecting them, so `iss` rides only on the success redirect and is
+  never a reason to redirect to an unvalidated URI. The device flow is
+  unaffected.
 - **RFC 8707 Resource Indicators: `resource` binds the access token
   audience** (#187). A client may send one or more `resource` parameters
   on `/authorize`, `/token` (every grant) and `/device_authorization`;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,9 +62,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   are now OAuth error redirects to the client (`error`,
   `error_description`, `state`, `iss`) instead of a local JSON 400, per
   RFC 6749 §4.1.2.1 - completing the RFC 9207 "error responses too"
-  requirement. Errors before the `redirect_uri` is trusted (unknown
-  client, malformed or unregistered `redirect_uri`) stay local JSON. The
-  device flow is unaffected.
+  requirement. `unsupported_response_type` (a non-`code` `response_type`)
+  is validated after the `redirect_uri` too, so it redirects as well.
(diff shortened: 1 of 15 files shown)


cdelmonte-zg:
A /code-review pass on the full stack (#187 + #188 + #189) surfaced five findings. Disposition:

- **#187 code (real bugs)** - charset regex trailing-newline (\Z not $), and refresh not re-validating inherited resources against the current allowed_resources - fixed on the #187 branch in 36f4f47, so #256 stays self-consistent; this PR is rebased on top of it. See #256 for detail.
- **#188 code** - the confidential client-auth resolution is copy-pasted across /introspect, /revoke, /device_authorization, and (unlike /token since #254) those three let Basic win when a body secret is also present instead of rejecting the two-method request (RFC 6749 §2.3). Neither is a vulnerability (check_client fails closed); tracked as **#259** rather than reopening the already-approved #254 for a refactor + consistency fix.
- **The confidential authorization_code auth requirement** (#188) is the intended change, already in the CHANGELOG migration note - no action.

Nothing in this PR's own #189 code (RFC 9207 iss) was flagged. This head is the rebased iss commit on top of the #187 fix; suite 1804, ruff clean. CI running.

cdelmonte-zg:
Review round 1: all three findings fixed in 6da0074, each reproduced first.

**1. Discovery false while /authorize still sent iss.** You're right - and it contradicts MCP SEP-2468 (send iss => advertise true, iss must be https). A single predicate (issuer_qualifies_for_iss_parameter) now drives BOTH the metadata and the emission, so they can never disagree. An http dev issuer sends no iss and advertises false; point the issuer at https (directly or reflected via issuer_from_request behind TLS) to turn RFC 9207 on. This supersedes the issue's earlier 'append always on http' note, which - as you showed - was itself inconsistent with the contract.

**2. Error responses not implemented ('success and error alike').** Implemented. redirect_uri is now validated BEFORE scope/PKCE/resource, and those errors are OAuth error redirects to the client carrying error, error_description, state and iss (RFC 6749 4.1.2.1 / RFC 9207); errors before redirect_uri is trusted (unknown client, malformed/unregistered redirect_uri) stay local JSON, never a redirect to an unvalidated URI. New _authorize_error_redirect helper. This is a real behaviour change (CHANGELOGed) - the PKCE/scope tests across five files now assert the redirect + error param via a new conftest.authorize_error helper.

**3. issuer_qualifies_for_iss_parameter edges.** Now rejects an empty query/fragment component (a bare ? or #), a schemeless-host value (https:foo), and a malformed issuer (bad IPv6, non-numeric port) without a discovery 500 - the same RFC 3986 edges as #256, applied to the issuer.

**e2e nit** also taken: test_agent captures the discovery issuer BEFORE the flow (the value a client compares against for the anti-mix-up property) and asserts iss present-and-matching iff advertised, absent otherwise.

Suite 1807, mypy/ruff/lint-imports clean; live 59/60 (Redirect URI Exact Match baseline). CI running.

Separately: your architecture-review direction is recorded as a project principle. Finding 2 here is exactly the symptom you named - the /authorize flow lacks an explicit state model (request parsed / client known / redirect_uri trusted / valid / authenticated), which is why 'local vs redirect error' wasn't centrally decidable. That, and #259's duplicated client-auth, are on the post-integration architecture-review list.

cdelmonte-zg:
Review round 2: both blockers fixed in 1716a42, plus the nit, each reproduced first.

**unsupported_response_type bypassed the error-redirect path.** You're right - response_type was validated in _validate_authorize_client, before the redirect_uri, so a non-code response_type on a valid client + registered redirect_uri returned a local JSON 400. It's now a validator of its own, run AFTER _validate_authorize_redirect_uri via _authorize_error_redirect: response_type=token + valid redirect_uri -> 302 with error=unsupported_response_type, state, iss; + invalid redirect_uri -> local 400. Pinned both directions. This is exactly the 'redirect_uri trusted' state made explicit for one more error, as you noted.

**redirect_uri with an existing query.** Reproduced: https://client.example/cb?tenant=foo + error params produced ...?tenant=foo?error=... folding error into tenant's value. New redirect_uri.append_authorization_params is now the single builder for BOTH the success and error authorization-response URIs - it preserves the existing query (RFC 6749 §3.1.2) and appends with '&'. The success path had the same latent bug you spotted and now shares the one fix. Pinned for both success and error redirects.

**Nit:** the stale discovery.py comment (still said iss 'is still sent but not advertised' on http) is corrected to the single-predicate reality.

Suite 1811, mypy/ruff/lint-imports clean; live 59/60 main + oauth21 3/3.

And point taken on scope: I fixed only these two contracts and did NOT start the structural refactor. append_authorization_params is the smallest possible 'one place builds authorization response URIs', not a redesign - the broader 'authorize flow needs an explicit state model' work stays for the post-integration architecture review, where #259 and the error-model shape belong. CI running.

cdelmonte-zg:
Round 3: fixed in 951884a, plus the PR body is now rewritten to the final state.

**append_authorization_params byte-for-byte.** You're right - parse_qsl/urlencode round-tripped the existing query and normalized it (%20 -> +, %2f -> %2F, bare flag -> flag=). RFC 6749 §3.1.2 says retain it, and the URI is already validated/matched, so there's nothing to interpret. It's now a plain string append (separator '&' if the URI already has a '?', else '?'), which also reads as what it is - append, not parse/normalize/rebuild. A direct unit test pins the byte preservation (x=a%20b&y=%2f&flag survives verbatim), which the parse_qs-based test could not see.

**PR body** rewritten to match the code: the single-condition metadata/behaviour (no more 'iss appended on http'), the error-redirect path (no more 'errors stay local'), and the one URI builder. Thanks for flagging it - exactly the 'a human should understand this later' concern, and a stale PR body becomes the squash message.

Suite 1812, mypy/ruff/lint-imports clean. CI running on this head.

cdelmonte-zg:

PR #260, opened 2026-08-29
feat: mock protected MCP server as an e2e fixture (RFC 9728) (#191)

cdelmonte-zg:
Fourth and final feature deliverable of milestone 2.9.0 - the one that gives the release its identity: **nanoidp can act as the authorization server for a real OAuth-protected MCP resource**, with public clients (#188), resource-bound tokens (#187) and RFC 9728 discovery, proven end to end. Closes #191. **Stacked on #189 (#258) -> #187 (#256) -> #188 (#254)**; base is `stack/189-iss-authorization-response` so CI runs and the diff is this PR's own (a new fixture + e2e + CI + docs + one example client, no `src/` changes).

**The fixture.** `examples/mock_mcp_server.py` is a minimal MCP Streamable HTTP resource server using the `mcp` SDK 2.0 resource-server mode (`MCPServer` + `TokenVerifier` + `AuthSettings`), with three scope-gated tools: `read_document` (`documents:read`), `delete_document` (`documents:write`), `admin_operation` (`admin`). `MockTokenVerifier` validates a bearer JWT **JWKS-only** against nanoidp: RS256 signature via `PyJWKClient` (fetch + kid + rotation), `iss`, `aud` == its own resource URL (string or array), `exp`, scopes from the `scope` claim. Any failure returns `None`, which the SDK turns into a `401` with `WWW-Authenticate: Bearer ... resource_metadata=...`; the SDK also serves `/.well-known/oauth-protected-resource` naming nanoidp in `authorization_servers`.

**Two authorization layers, kept distinct** (a review correction on this PR). A resource-level scope floor `documents:read` is declared through the SDK's `AuthSettings.required_scopes`, so the bearer middleware returns the **conformant MCP/RFC 9728 challenge** - HTTP `403` with `WWW-Authenticate: Bearer error="insufficient_scope", ..., resource_metadata=...` - **before any tool runs**. The finer `documents:write` / `admin` operations are checked **inside** each tool via `get_access_token()` and surface as an in-band MCP tool error (`isError`): a second, application-defined authorization decision, not an HTTP challenge. The e2e tests and labels both, so neither is mistaken for the other.

[...]

Issue #191, opened 2026-08-23
Mock protected MCP server as an e2e fixture (RFC 9728)


cdelmonte-zg:
**Why.** Without a resource server nothing in this milestone can be demonstrated end to end: "wrong audience", "insufficient scope", the 401 -> discovery -> authorize -> token -> `tools/call` round trip. It must stay a fixture, not a second product.

**Scope.**
- `examples/mock_mcp_server.py`: an MCP Streamable HTTP server (mcp SDK) exposing three tools: `read_document` (`documents:read`), `delete_document` (`documents:write`), `admin_operation` (`admin`); validates bearer JWTs against nanoidp's JWKS, checks `iss`, `aud` == its own resource URL, `exp` and scopes; unauthenticated -> 401 with `WWW-Authenticate: Bearer resource_metadata=...`; serves `/.well-known/oauth-protected-resource` (RFC 9728) naming nanoidp in `authorization_servers`; insufficient scope -> 403 with `error="insufficient_scope"`, `scope=` and `resource_metadata=` (RFC 9728 section 5.1 / MCP 2026-07-28).
- **Validation mode is JWKS only** (self-contained JWT). Consequence, stated in the fixture's docs: a token revoked at nanoidp stays valid at the resource server until `exp`; revocation is demonstrated through `/introspect` in the e2e, not at the mock server. One mode only; an introspection-based variant is out of scope unless the point becomes teaching the difference between self-contained and introspected tokens.
- **Deterministic conformance path, no LLM anywhere in this issue**: `examples/test_agent.py` acts as the MCP client through the `mcp` SDK (`tools/list` / `tools/call` over Streamable HTTP), driving the 401 -> RFC 9728 discovery -> `/authorize` (PKCE, `resource=`) -> `/token` -> `tools/call` sequence itself. A failure here must mean PKCE, discovery, `resource`, audience, scope or MCP framing is wrong, never that a model chose not to call a tool.
[...]

PR #261, opened 2026-08-29
refactor: move the e2e harness from examples/ to e2e/ (2.9 arch review)

cdelmonte-zg:
First refactor from the integrated 2.9.0 architecture review (on `review/2.9.0-oauth-integration`). Behaviour-preserving, per the review rule "no new behaviour except an obvious bug". **Stacked on #260 (#191);** base is `stack/191-mock-mcp-server` so CI runs and the diff is this PR's own.

**Why.** The end-to-end suite lived under `examples/` next to the real usage presets, but it was never a usage example - it is the CI e2e suite that drives a running server over the network. Mixing the harness with the presets made the tree harder to read, which is exactly the human-readable-architecture goal the review is about.

**What moves.** Only the harness -> `e2e/`:

| File | Role |
| --- | --- |
| `test_agent.py` | the main agent (default / `--oauth21` / `--saml-signed` / `--mcp`) |
| `mock_mcp_server.py` | the OAuth-protected MCP resource-server fixture (#191) |
| `mcp_smoke_test.py` | the MCP stdio transport smoke test |
| `gen_sp_keypair.py` | the signed-SAML test SP keypair generator |

`examples/` keeps only the real integration presets (spring-boot-saml, react-spa-pkce, microservices-client-credentials, cli-device-flow, persona-login) and the plugins - so `examples/README.md` (a preset index) is now accurate as-is.

**Reference updates** (every invocation path): `.github/workflows/e2e.yml`, `CONTRIBUTING.md`, `VISION.md`, the mdBook guides (`testing-an-mcp-client`, `architecture`, `saml`), `pyproject.toml`, and two source/test comments (`routes/api.py`, `tests/`), from `python examples/...` to `python e2e/...`. New `e2e/README.md`. Historical CHANGELOG entries keep their original paths (accurate for their release); the 2.9 entries and a new **Changed** note point at `e2e/`.

**No behaviour, no packaged code change.** `git` tracks all four as renames. mypy `files=["src"]` and the hatch sdist never included `examples/`, so neither scope changes; `ruff check .` still covers the harness under its new path.

**Verification** (from the new `e2e/` location): MCP interop suite **9/9**, MCP stdio smoke test passing, unit suite **1812**, `ruff` and `mypy` clean.

Not merging yet: like the rest of the 2.9 stack, this stays off `main` during the 2.8.0-rc2 freeze and lands with the stack after 2.8.0 final.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -56,7 +56,7 @@ jobs:
           NANOIDP_E2E_HOOK_LOG: ${{ github.workspace }}/hooks.log
           NANOIDP_E2E_PLUGIN: echo
           NANOIDP_E2E_PLUGIN_RECORD: ${{ github.workspace }}/echo-record.jsonl
-        run: python examples/test_agent.py
+        run: python e2e/test_agent.py
 
       - name: Start oauth21 server
         run: |
@@ -70,11 +70,11 @@ jobs:
           exit 1
 
       - name: Run oauth21 profile suite
-        run: python examples/test_agent.py --oauth21 --url http://localhost:8001
+        run: python e2e/test_agent.py --oauth21 --url http://localhost:8001
 
       - name: Prepare signed-SAML config
         run: |
-          python examples/gen_sp_keypair.py --out e2e-saml
+          python e2e/gen_sp_keypair.py --out e2e-saml
           cp -r config e2e-saml/config
           python - <<'EOF'
           import yaml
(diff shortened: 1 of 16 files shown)


cdelmonte-zg:
P3 fixed (4b501ea). The e2e-move entry had created a second `### Changed` in `[Unreleased]` (the first Changed + Security sections were already there, lower in the file than my initial check looked). Moved the entry under the existing `### Changed`, so the section appears once - `[Unreleased]` is now Fixed / Added / Changed / Security with no duplicate heading. Nothing else touched; the rename diff and all invocation-path updates are unchanged.

cdelmonte-zg:

Issue #262, opened 2026-08-29
Align /introspect, /revoke, /device_authorization client auth with the token endpoint (RFC 7009/7662)


cdelmonte-zg:
## Summary

`/introspect`, `/revoke` and `/device_authorization` authenticate the client with a lenient "HTTP Basic **or** `client_secret_post` body" check, accepting either channel and not enforcing the client's registered `token_endpoint_auth_method`. `/token` is stricter: `_enforce_token_endpoint_auth` (#188) enforces the registered method and rejects presenting two auth methods in one request (RFC 6749 §2.3).

RFC 7009 §2.1 (revocation) says the client authenticates per RFC 6749 §2.3, and RFC 8628 (device) ties client auth to the token endpoint's rules - so for those two, aligning with the token endpoint is RFC-grounded. RFC 7662 (introspection) requires *some* client authentication but does NOT mandate reusing the token-endpoint method (its security considerations leave method choice to the deployment). So for `/introspect`, reusing the registered method is a deliberate **nanoidp consistency policy**, not a spec requirement. Either way the lenient path is worth closing: a `client_secret_basic`-registered client can currently authenticate to `/introspect` via a body secret, and a client presenting both Basic and a body secret is not rejected.

## Proposal

Make the three endpoints authenticate through the same boundary as `/token` (`_enforce_token_endpoint_auth` or a shared derivative), so the registered method is enforced and double-auth is rejected everywhere. Preserve the existing, correct policy differences:
- `/introspect` never relaxes for public clients (RFC 7662 §2.1: protect against token scanning; `none` stays out of `introspection_endpoint_auth_methods_supported`).
- `/revoke` keeps its public-client relaxation (RFC 7009 §2.1: `none` client identified by `client_id`, the ownership check stands in for authentication).
- `/device_authorization` keeps requiring confidential auth (public device flow is a separate follow-up).

## Why separate

This is a **behaviour change** (some requests that authenticate today would start being rejected), so it is deliberately NOT folded into the behaviour-preserving extraction that introduced `_authenticate_confidential_client` (that PR only removed the triplicated inline block). Needs its own tests and a CHANGELOG note.

Context: came out of the integrated 2.9.0 architecture review (the "one protocol rule, one obvious location" pass). Related to the original #259 concern about centralising confidential client auth.

cdelmonte-zg:
[...]

PR #263, opened 2026-08-29
refactor: one home for confidential client auth (/introspect, /revoke, /device) (2.9 arch review)

cdelmonte-zg:
Second refactor from the integrated 2.9.0 architecture review. Behaviour-preserving, per the review rule "no new behaviour except an obvious bug". **Stacked on #261 (e2e move) -> #260 (#191);** base is `stack/e2e-directory` so CI runs and the diff is this PR's own.

**Finding (from the review).** Client authentication had two homes. `/token` uses the strict `_enforce_token_endpoint_auth` (#188): it enforces the client's registered `token_endpoint_auth_method` and rejects presenting two auth methods at once (RFC 6749 §2.3). But `/introspect`, `/revoke` and `/device_authorization` each carried the *same* inline lenient block - "HTTP Basic **or** `client_secret_post` body" - triplicated. That is the "one protocol rule, one obvious location" violation the review targets.

**What this PR does.** Extracts the shared lenient check into a single `_authenticate_confidential_client(config) -> (authenticated, client_id)` helper, and calls it from the three endpoints. **Purely behaviour-preserving** - each endpoint keeps its own policy *around* the shared check:
- `/introspect` never relaxes for public clients (RFC 7662 §2.1: guard against token scanning; `none` stays out of `introspection_endpoint_auth_methods_supported`).
- `/revoke` calls the helper only when the client is not public; its public-client relaxation (RFC 7009 §2.1 ownership check) is unchanged.
- `/device_authorization` requires confidential auth outright.

**Deliberately not in this PR.** The helper is kept *separate* from `_enforce_token_endpoint_auth`. RFC 7009 §2.1 / RFC 7662 §2.1 say these endpoints authenticate "as at the token endpoint", so the truly-one-boundary end state would route them through the strict check too - but that **rejects some requests that authenticate today** (a body secret from a Basic-registered client; Basic+body together), i.e. a behaviour change. Tracked in **#262**, with its own tests and CHANGELOG note.

**No CHANGELOG entry here**: no behaviour and no public surface changed.

**Verification.**
- Unit suite **1812**, `ruff` and `mypy` clean.
- Live (negative + positive), matching pre-refactor behaviour exactly:
  - `/introspect`: no creds -> 401; wrong secret -> 401; **public client (no secret) -> 401** (no relaxation); valid Basic -> 200; `client_secret_post` body -> 200.
  - `/revoke`: valid Basic -> 200; no creds -> 401; **public client (no secret) -> 200** (relaxation preserved); confidential client without secret -> 401.
  - `/device_authorization`: valid Basic -> 200; no creds -> 401; `client_secret_post` body -> 200.

Not merging yet: stays off `main` during the 2.8.0-rc2 freeze, lands with the 2.9 stack after 2.8.0 final.

### DIFF
--- a/src/nanoidp/routes/oauth.py
+++ b/src/nanoidp/routes/oauth.py
@@ -1481,6 +1481,34 @@ def _enforce_token_endpoint_auth(
     return None
 
 
+def _authenticate_confidential_client(
+    config: ConfigManager,
+) -> tuple[bool, Optional[str]]:
+    """Confidential client authentication for the RFC 7009/7662/8628 endpoints.
+
+    The shared "credentials via HTTP Basic or client_secret_post body" check
+    that /introspect, /revoke and /device_authorization each carried inline.
+    Returns ``(authenticated, resolved_client_id)``; the caller applies its own
+    policy around it (introspect never relaxes for public clients; revoke calls
+    this only when the client is not public; device requires it outright).
+
+    This is deliberately SEPARATE from ``_enforce_token_endpoint_auth`` (the
+    /token boundary, #188), which additionally enforces the client's registered
+    token_endpoint_auth_method and rejects presenting two auth methods at once.
+    Aligning these endpoints with that stricter boundary (RFC 7009 §2.1 /
+    RFC 7662 §2.1 say they authenticate "as at the token endpoint") is a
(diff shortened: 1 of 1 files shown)


cdelmonte-zg:

PR #264, opened 2026-08-29
docs: document the MCP generate_token resource boundary (2.9 arch review)

cdelmonte-zg:
Third refactor from the integrated 2.9.0 architecture review. **Doc-only, no behaviour change.** **Stacked on #263 (client-auth) -> #261 (e2e move) -> #260 (#191);** base is `stack/centralize-client-auth`.

**Finding (from the review).** The MCP `generate_token` tool sets the access token `aud` from its `resource` argument but never calls `resolve_resources`, so the per-client RFC 8707 `allowed_resources` ceiling that `/token`, `/authorize` and `/device_authorization` enforce (`invalid_target`) is not applied via MCP. A reader could assume parity with `/token` and treat this as a gap.

**It is not a gap - it is inherent.** `generate_token` has **no client in the request**: it mints a token directly for a *user*, a testing/simulation affordance, not an OAuth grant. There is no per-client `allowed_resources` to apply, so the `aud` is whatever the caller passes - by design. The confusion was purely that the boundary was undocumented.

**What this PR does.** Documents the boundary in two places so issue and code tell the same story:
- the inline comment at the `resource=` call in `_tool_generate_token`, contrasting it explicitly with the grant endpoints;
- the tool's `resource` schema `description` (what an MCP client sees), noting that the per-client ceiling does not apply here.

No code path changes. Verified: `ruff` and `mypy` clean; the MCP tool tests (210) pass.

Not merging yet: stays off `main` during the 2.8.0-rc2 freeze, lands with the 2.9 stack after 2.8.0 final. With this, the three actionable findings from the architecture review are addressed; the optional `oauth.py` package split (finding 3) is left as a separate decision.

### DIFF
--- a/src/nanoidp/mcp_server.py
+++ b/src/nanoidp/mcp_server.py
@@ -551,7 +551,11 @@ _TOOLS: list[Tool] = [
                         "array for several) instead of oauth.audience, so a "
                         "token minted for one MCP server is rejected by "
                         "another. Each must be an absolute URI without a "
-                        "fragment (optional)"
+                        "fragment. This tool mints for a user directly, with no "
+                        "client in the request, so the per-client "
+                        "allowed_resources ceiling that /token enforces does "
+                        "not apply here - the aud is whatever you pass "
+                        "(optional)"
                     ),
                 },
             },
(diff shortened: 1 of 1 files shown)


cdelmonte-zg:
Change request addressed (adca8c1). You were right: the description promised "each must be an absolute URI without a fragment" while the code applies no such validation - `_tool_generate_token` passes `resource` straight to `create_token`, which copies it verbatim into `aud`, so `resource=["foobar"]` yields `aud="foobar"`. Reworded to describe the argument for what it actually is - an **audience override** with no client context, hence no RFC 8707 syntax validation and no per-client `allowed_resources`; the supplied values are used directly. Framing it as an override (not "the same resource as the grant endpoints") matches your suggestion. No behaviour change; 249 MCP/generate_token/resource tests green.

cdelmonte-zg:

PR #265, opened 2026-08-29
refactor: extract the /token grant handlers into oauth_grants.py (2.9 arch review)

cdelmonte-zg:
Fourth refactor from the integrated 2.9.0 architecture review, and the **middle-path** version of finding 3 (oauth.py is large) - chosen deliberately over a full package split. **Behaviour-preserving.** **Stacked on #264 -> #263 -> #261 -> #260 (#191);** base is `stack/generate-token-boundary-doc`.

**Why the middle path.** The evaluation showed a full 9-module `routes/oauth/` split is feasible and low-behaviour-risk, but it is a ~2300-line reshuffle (git can't see it as clean renames), it adds cross-module indirection (`_blueprint.py`, `_shared.py`), and it contradicts the "small, mechanical" bar this review pass set. The real weight of the file is the `/token` grant area (~750 lines); extracting just that gives most of the readability win for a fraction of the risk.

**What moves.** The cohesive grant unit -> `routes/oauth_grants.py`:
- the five `_grant_*` handlers (`refresh_token`, `password`, `authorization_code`, `client_credentials`, `device_code`),
- `_GrantOutcome`, `_GrantContext`, `GrantResult`, `_resolve_token_resource`, `_GRANT_HANDLERS`.

`token()` in `oauth.py` imports `_GRANT_HANDLERS`, `_GrantContext`, `_GrantOutcome` and dispatches exactly as before. **oauth.py: 2296 -> 1550 lines.**

**No cycle, no new concepts.** The handlers depend only on `services/` and `models`, never back on `routes.oauth`, so the import is one-way - verified by `lint-imports` (the `routes -> services -> config` layering contract still holds with the new module). The module's external contract is unchanged (`oauth_bp`), and `_GrantOutcome` stays importable from `routes.oauth` for the one test that imports it, so no test churn.

**Verification.**
- Unit suite **1812**; `lint-imports` 2 contracts kept; `ruff` and `mypy` clean (37 files - `oauth_grants.py` is type-checked).
- Default e2e suite **60/60**, with **OAuth2/OIDC 26/26** exercising every moved grant handler live (authorization_code, client_credentials, refresh incl. resource, password, PKCE, scope enforcement).

The full package split (finding 3) is left as a separate future decision; the file is already traceable, so this is a readability improvement, not a correctness fix.

Not merging yet: stays off `main` during the 2.8.0-rc2 freeze, lands with the 2.9 stack after 2.8.0 final.


cdelmonte-zg:
P3 nit fixed (49dc480). The docstring said "depend only on services/models" while the module imports `._audit.audit_event`, a route-layer helper. Reworded to name it explicitly, so the doc is precise; the architectural property it asserts is unchanged and still true (no import back on `routes.oauth`, so the dependency stays one-way with no cycle - `lint-imports` confirms). Rebased onto the fixed #264 head; still MERGEABLE.

cdelmonte-zg:

Issue #259, opened 2026-08-29
Factor the confidential client-auth resolution shared by /introspect, /revoke, /device_authorization; reject two auth methods there too


cdelmonte-zg:
Follow-up from #188 / PR #254 (a /code-review pass on the #187-#189 stack; deferred here rather than reopening the approved #254).

Two related findings on the confidential client authentication at the non-token endpoints:

**1. Duplicated credential resolution.** The 'Basic or client_secret_post' block that reads `request.authorization` vs `client_id`/`client_secret` from the body and validates via `check_client` is copy-pasted, near-identically, across `/introspect`, `/revoke` and `/device_authorization` (routes/oauth.py). A future fix to that resolution has to be applied in three places and will drift. Factor a single `_resolve_confidential_auth(config)` helper carrying the rule once, alongside the `_enforce_token_endpoint_auth` boundary #188 added for `/token`.

**2. Inconsistent RFC 6749 §2.3 handling.** `/token` (since #254) rejects a request that presents HTTP Basic AND a body `client_secret` (two authentication methods in one request, RFC 6749 §2.3). The other three endpoints check `if auth:` first and silently let Basic win, ignoring a body secret when both are present - so the same double-credential request is treated differently depending on the endpoint. The shared helper from (1) should apply the same 'reject two methods' rule everywhere.

Neither is a vulnerability: `check_client` fails closed, and these endpoints correctly authenticate a well-formed request. This is consistency and maintainability. Not a blocker for #254; the #187 review found it while tracing the wider auth surface. Targeted at 2.9.0 alongside the rest of the milestone.

The `/token` boundary (`_enforce_token_endpoint_auth`) additionally enforces the client's REGISTERED method (basic vs post) - the non-token endpoints intentionally accept either channel, so the shared helper should keep that distinction: enforce the method only where token_endpoint_auth_method governs (the token endpoint), reject double-auth everywhere.

cdelmonte-zg:
Closing as consolidated - both halves are now handled:

- **Finding 1 (duplicated confidential credential resolution)** shipped in #263, which extracted the shared `_authenticate_confidential_client` helper so the rule had one home (behaviour-preserving).
[...]

Issue #255, opened 2026-08-29
Public clients on the device flow (RFC 8628)


cdelmonte-zg:
Follow-up from #188 / PR #254 (deliberately out of scope there).

#188 shipped public clients (`token_endpoint_auth_method: none`) for the authorization_code + PKCE flow. The device authorization grant was left confidential-only on purpose: `/device_authorization` still requires client authentication (HTTP Basic or, since #188, client_secret_post), and the `urn:ietf:params:oauth:grant-type:device_code` token grant goes through the same confidential path.

RFC 8628 explicitly contemplates public clients: a public client presents `client_id` at the device authorization endpoint (§3.1) and again when polling the token endpoint (§3.4), with no client secret. So a public CLI/TV/IoT-style client currently cannot use nanoidp's device flow at all.

**Scope to decide when picked up:**
- `/device_authorization`: a public `client_id` (method `none`) is accepted by client_id alone, no secret; a confidential client still authenticates.
- The `device_code` token grant: a public client polls with `client_id` alone; the existing `_enforce_token_endpoint_auth` boundary already refuses `client_credentials` for public clients and identifies device-flow requests by client_id, so the main work is the device authorization endpoint's own auth gate plus binding the issued device_code to the (public) client.
- Rate limiting / abuse: an unauthenticated device authorization request is cheaper to spam; confirm the existing per-process limiter covers it.
- e2e: a public client completes the device flow end to end (test_agent already exercises the confidential device flow).

No client-secret handling changes; PKCE is not part of the device grant. Targeted at 2.9.0 alongside the rest of the OAuth/MCP interoperability foundation, but lower priority than #187/#189/#191.

PR #266, opened 2026-08-29
feat: enforce client auth at /introspect, /revoke, /device as at /token (#262, BREAKING)

### LINKED ISSUE #262: Align /introspect, /revoke, /device_authorization client auth with the token endpoint (RFC 7009/7662)

cdelmonte-zg:
**BREAKING** (client-authentication contract) - the change that turns this line into a **major release (3.0.0)** alongside #188. **Stacked on #265 -> #264 -> #263 -> #261 -> #260 (#191);** base is `stack/oauth-grants-module`. Closes #262 and #259.

**What.** `/introspect`, `/revoke` and `/device_authorization` now authenticate a confidential client the same way `/token` does (#188): the registered `token_endpoint_auth_method` is enforced and two auth methods in one request are rejected.

- A `client_secret_basic` client must present its secret over **HTTP Basic**; a body `client_secret` is rejected as the wrong channel.
- A `client_secret_post` client must present its secret in the **request body**; HTTP Basic is rejected.
- **Basic + a body secret together** is rejected (RFC 6749 §2.3 - "MUST NOT use more than one"); Basic no longer silently wins.

Applying the one registered method across all four endpoints is nanoidp's **consistency policy**, not an obligation of every RFC: RFC 7009 §2.1 and RFC 8628 do tie `/revoke` and `/device_authorization` to the token-endpoint method, while RFC 7662 requires *some* client authentication at `/introspect` but does **not** mandate reusing that method - nanoidp reuses it there too rather than add a second field. Previously these endpoints accepted either channel and did not reject double-auth. A client that legitimately used `client_secret_post` at introspection while registered `client_secret_basic` was not misbehaving; it stops working here by nanoidp's stricter policy, and the CHANGELOG says so.

**Why now.** We were already cutting a major for #188 (confidential clients must authenticate on `authorization_code`; registered method enforced at `/token`). #262 is the same-family breaking change on the other three endpoints - landing it in the **same major** means the whole client-auth contract changes once, with one migration note, instead of forcing a second major later.

**How (one home for the rule).** The lenient `_authenticate_confidential_client` helper (added behaviour-preservingly in #263) is replaced by a single `_enforce_registered_client_auth(config, client_id, auth, body_secret) -> Optional[reason]` boundary, now shared by all four endpoints - `_enforce_token_endpoint_auth` calls it too. This finishes the #259/#188 arch-review thread (one obvious location for confidential client auth), so **#259 is closed** (its refactor half shipped in #263, its "reject two auth methods" half is here).

**Policy preserved, made explicit per endpoint.** `/introspect` and `/device_authorization` refuse public clients (RFC 7662; device-flow public support stays deferred to #255); `/revoke` keeps its RFC 7009 §2.1 ownership relaxation (a public client revokes with `client_id` alone).

**Migration.** Register each client with the `token_endpoint_auth_method` it actually uses; send credentials over that one channel; declare secret-less clients as `token_endpoint_auth_method: "none"`. A client already using its registered method consistently needs no change.

**Verification.**
- Unit suite **1826** (+14): a new `TestNonTokenEndpointAuthEnforcement` covering, per endpoint, body-secret-from-a-basic-client rejected, Basic-from-a-post-client rejected, Basic+body rejected, the public-client policy, and the valid paths still working.
- `ruff` and `mypy` clean.
- **Live over HTTP**: `/introspect`, `/revoke`, `/device_authorization` now return `401` for a body secret and `401` for double-auth, while a valid HTTP Basic still `200`s.

CHANGELOG gains a **Breaking Changes** section covering #188, #262 and #189 with per-change migration notes. Not merging yet: held off `main` during the 2.8.0-rc2 freeze; lands with the stack as **3.0.0** after 2.8.0 final.


cdelmonte-zg:
All three findings addressed (39c4ca1).

**[P1] Stale config surfaces - fixed.** Updated everywhere the field is described: the `OAuthClient.token_endpoint_auth_method` Field description (`models.py`), the MCP `create_client` tool description (`mcp_server.py`), the configuration reference (`configuration.md`) - all now say the registered method is enforced at all four client-authenticated endpoints, not just `/token`. On the schema: `docs/schema/config.v1.json` is drift-free because the config schema is built from `config_documents.ClientEntry`, whose `token_endpoint_auth_method` carries **no** `Field` description - the model prose never reached the schema (confirmed: `config-schema --write` produces zero diff, and the drift test is green). So there was nothing to regenerate there.

**[P2] RFC 7662 rationale - fixed, reframed as nanoidp policy.** You are right: RFC 7662 requires *some* client authentication at introspection but does not mandate reusing the token-endpoint method (RFC 7009/RFC 8628 do tie `/revoke` and `/device_authorization` to it). Corrected the `_enforce_registered_client_auth` docstring, the `/introspect` comment, the CHANGELOG, the PR body and issue #262 to present the introspection alignment as nanoidp's deliberate consistency policy. Also dropped "well-behaved clients are unaffected" - a client using `client_secret_post` at introspection while registered `client_secret_basic` was not misbehaving; it breaks by our stricter policy, and the CHANGELOG now says exactly that.

**[P2] Positive `client_secret_post` paths - added.** `TestNonTokenEndpointAuthEnforcement` now has `test_{introspect,revoke,device}_post_client_via_body_works`: `post-cli` authenticates over the body and gets `200` at each endpoint, so the contract shows the correct channel works, not only that Basic is rejected.

Unit suite 1829 (+3), config-schema drift test green, ruff and mypy clean.

cdelmonte-zg:
Both P3 wording nits fixed (abe821c + a PR-body edit), so no old interpretation is left in the test or PR history:

- **Test class docstring** reworded to the policy framing: "nanoidp applies the registered `token_endpoint_auth_method` consistently to /introspect, /revoke and /device_authorization ... RFC 7009 and RFC 8628 tie /revoke and /device_authorization to token-endpoint authentication; RFC 7662 permits authentication at introspection but leaves the method open." No more "the same way /token does (RFC 7662 §2.1)".
- **PR migration line** changed from "A well-behaved confidential or public client needs no change" to "A client already using its registered method consistently needs no change" - which no longer implies an RFC-7662-compliant introspection client was misbehaving.

Docstring/text only; nothing functional changed. Thanks for the thorough passes on this one - the client-auth contract is a lot cleaner and more honestly described for them.

cdelmonte-zg:

PR #269, opened 2026-08-29
feat: reject refresh tokens without a client_id binding (#73, BREAKING)

### LINKED ISSUE #73: 3.0: stop accepting refresh tokens without a client_id binding claim
Tracking issue for the transitional compatibility introduced in 2.2.0 (#56 / PR #57): refresh tokens minted before the `client_id` binding claim existed are still spendable by **any** authenticated client until they expire. That was the accepted residual at the time — cutting it was deferred to the next major.

At 3.0: a refresh token without a `client_id` claim is rejected with `invalid_grant` (RFC 6749 §5.2). One-line check in the refresh path plus a changelog entry telling testers to mint fresh tokens.

Blocked on: a 3.0 release actually being planned. This milestone is where other breaking cleanups accumulate in the meantime.

cdelmonte-zg:
First of the three follow-ups going into 3.0.0 alongside the landed 2.9 stack. **Breaking** (the milestone's whole purpose - accumulate the deferred breaking cleanups in the major). Closes #73.

**What.** A refresh token with **no `client_id` binding claim** is now refused with `invalid_grant` (RFC 6749 §5.2). The binding was added in 2.2.0 (#56, RFC 9700 §4.14: a refresh token may only be spent by its issuing client); tokens minted before it were still spendable by any authenticated client until they expired - the residual compat #73 tracked, "blocked on a 3.0 being planned". It is now.

- One check in `_grant_refresh_token`: `payload.get("client_id")` absent -> `invalid_grant`.
- *Migration:* discard refresh tokens minted before 2.2.0; obtain new ones (every grant since 2.2.0 already binds them).

**The generate_token interaction (surfaced by the suite, addressed here).** The MCP `generate_token` tool mints directly for a user with no client, so its refresh tokens were unbound - and now non-refreshable. Rather than leave the primary test-token minter unable to produce a refreshable token, `generate_token` gains an **optional `client_id`**:
- given -> **must name a real client**; binds the token and issues a refresh token spendable by that client (a validated, working credential);
- omitted -> unbound access token with **no refresh token at all** - handing back one with no client_id binding would be a dead credential since 3.0.

This is additive; the resource/`allowed_resources` boundary documented in #264 is unchanged (still never applied here, even with `client_id`). MCP schema + `book/src/reference/mcp.md` updated.

**Tests.** The #56 legacy-compat test (`..._still_works`) is flipped to `..._is_rejected` (400 `invalid_grant`, the real #73 requirement, built directly from a hand-crafted unbound token). For generate_token: `test_unbound_token_has_no_refresh_token` (unbound -> no refresh token issued), `test_bound_token_refresh_token_is_spendable` (a valid client_id -> the refresh token actually works on `/token`, 200), and `test_unknown_client_id_is_rejected`. The two claims-across-refresh tests now mint **bound** tokens so they still exercise claims persistence with a valid 3.0 token.

**Verified**: unit suite **1832**, `ruff` and `mypy` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,9 +56,10 @@ previous leniency allowed - needs a one-time adjustment.
   until they expired - a transitional compat deferred to the next major, now
   closed. *Migration:* discard refresh tokens minted before 2.2.0 and obtain
   new ones (every grant since 2.2.0 already binds them). The MCP `generate_token`
-  tool gains an optional `client_id` that binds the minted token so its refresh
-  token is spendable; without it the tool mints an unbound token, fine for a
-  one-shot access-token test but not refreshable.
+  tool gains an optional `client_id` (which must name a real client) that binds
+  the minted token and issues a refresh token spendable by it; without it the
+  tool mints an unbound access token with no refresh token at all, rather than
+  hand back one that could not be spent.
 
 ### Added
 - **Mock protected MCP server as an e2e fixture** (#191). `e2e/mock_mcp_server.py`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,15 @@ previous leniency allowed - needs a one-time adjustment.
   missing/malformed/unregistered `redirect_uri`) still return JSON locally.
   *Migration:* a client that parsed the `400` body should read the error from
   the redirect query instead - which is what a spec-compliant client already does.
+- **A refresh token without a `client_id` binding claim is rejected** (#73,
+  `invalid_grant`, RFC 6749 §5.2). The binding was added in 2.2.0 (#56);
+  tokens minted before it were still spendable by any authenticated client
+  until they expired - a transitional compat deferred to the next major, now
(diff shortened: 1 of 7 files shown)


cdelmonte-zg:
Both points fixed (73f8fa8). And agreed on not splitting it - it is direct integration fallout of #73.

1. **client_id is validated.** A given `client_id` must name a real client, else a clean MCP error (`Client 'X' not found`) up front - no more stamping a binding no authenticable client could match. `allowed_resources` is deliberately still not consulted (the #264 boundary stands; generate_token stays an admin/test minter, not an OAuth grant).

2. **No refresh token when unbound.** `issue_refresh_token=client_id is not None`, so an omitted `client_id` yields an access token with **no** `refresh_token` key at all, rather than a dead credential. `create_token` already supported this.

Tests updated to match: `test_unbound_token_has_no_refresh_token` (no refresh token issued), `test_bound_token_refresh_token_is_spendable` (a valid client_id -> refresh works), `test_unknown_client_id_is_rejected`. Schema description, MCP reference and CHANGELOG reworded. Suite 1832, ruff/mypy clean.

PR #270, opened 2026-08-29
feat: per-component RFC 3986 validation for resource indicators (#257)

### LINKED ISSUE #257: Full RFC 3986 per-component URI validation for resource indicators
Follow-up from #187 / PR #256 review (deliberately deferred, not a blocker).

`services/resource.py`'s `is_valid_resource_indicator` validates a resource indicator (RFC 8707 §2: an absolute URI without a fragment) with a global RFC 3986 character whitelist (`_RFC3986_CHARS`) plus targeted checks (no `#`, valid percent-encoding, numeric port via `urlparse().port`). This catches the concrete holes found in review (raw space, bad IPv6 literal, `%ZZ`, trailing `%`, `:abc`, empty fragment), but a global whitelist is not the full per-component ABNF:

- `[` and `]` are gen-delims allowed somewhere in a URI, but not in a path's `pchar` (`pchar = unreserved / pct-encoded / sub-delims / ":" / "@"`), so `https://example.com/a[b]` is accepted where the grammar would reject it.
- `urlparse().port` is slightly stricter than the RFC in the other direction (it enforces the 0-65535 range, where RFC 3986 §3.2.3 defines `port = *DIGIT` syntactically).

**Not a security issue**: no audience bypass, escalation or OAuth regression - RFC 8707 requires an absolute URI without a fragment and otherwise lets the AS apply its own policy on acceptable resource identifiers. This is standards-completeness.

If we want truly exhaustive RFC 3986 validation, do it as a separate refactor (a real URI validator or per-component ABNF), rather than growing the regex further. Low priority; the current check is sufficient for the resource-indicator contract.

cdelmonte-zg:
Second of the three follow-ups for 3.0.0. **Tightening only** (standards-completeness, not a security issue - the issue #257 says so). Closes #257.

**What.** `is_valid_resource_indicator` validated a resource indicator (RFC 8707 §2: an absolute URI without a fragment) against one permissive RFC 3986 character whitelist plus targeted probes. That whitelist accepted `[` and `]` **anywhere** - so `https://host/a[b]` passed where the path grammar rejects them - and leaned on `urlsplit`'s numeric-range port check (RFC 3986 §3.2.3 defines `port = *DIGIT`).

Replaced with **per-component ABNF**: `urlsplit` into scheme / authority / path / query, each matched against its own production:
- **path / query**: `pchar` (`[`/`]` excluded - they are host-only gen-delims);
- **authority**: `userinfo`, host, `*DIGIT` port; **IP-literal** hosts (`[::1]`) validated via `ipaddress.IPv6Address` (or an `IPvFuture` regex);
- **reg-name** host rejects brackets (they are legal only in an IP-literal).

**The pre-parse guard stays, and its role is now explicit.** `urlsplit` doesn't just permit a raw space - it silently **strips** ASCII tab/newline (WHATWG) before parsing, so a control character would otherwise reach the token `aud` with the character removed. The `_RFC3986_CHARS` check rejects the space/tab/newline/control/non-ASCII bytes the per-component grammar never sees.

**Impact.** Mostly a tightening: malformed input to this opt-in feature that used to slip through is now `invalid_target` (and a scoped IPv6 `[fe80::1%eth0]` is rejected, per RFC 9844). It also stops rejecting a valid path-empty absolute URI (`about:`). No audience bypass, escalation or OAuth regression. Categorised under **Changed**, not Breaking.

**Tests.** 12 new syntax cases (`[`/`]` in path and query rejected; `[::1]` host with/without port accepted; malformed `[gg::1]` rejected; `userinfo`; a syntactically-valid out-of-range port accepted per `*DIGIT`; `[` in a reg-name host rejected); all 16 existing cases preserved. Suite **1838**, `ruff` and `mypy` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -175,6 +175,14 @@ previous leniency allowed - needs a one-time adjustment.
   repository harder to read. Invocations change from `python examples/...` to
   `python e2e/...` (CI, CONTRIBUTING and the docs are updated); no behaviour
   and no packaged code changed.
+- **Resource indicators are validated per RFC 3986 component, not by a single
+  character whitelist** (#257). A `resource` is still an absolute URI without a
+  fragment (RFC 8707 §2), but each component is now checked against its own
+  ABNF: `[`/`]` are accepted only inside an IP-literal host (so
+  `https://host/a[b]` is rejected where it used to pass), a port is `*DIGIT`
+  (no numeric-range limit, matching RFC 3986 §3.2.3), and IPv6 host literals
+  are validated. Tightening only, on malformed input to an opt-in feature; no
+  audience bypass, escalation or change for a well-formed resource.
 
 ## [2.8.0] - 2026-08-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -181,8 +181,10 @@ previous leniency allowed - needs a one-time adjustment.
   ABNF: `[`/`]` are accepted only inside an IP-literal host (so
   `https://host/a[b]` is rejected where it used to pass), a port is `*DIGIT`
   (no numeric-range limit, matching RFC 3986 §3.2.3), and IPv6 host literals
-  are validated. Tightening only, on malformed input to an opt-in feature; no
-  audience bypass, escalation or change for a well-formed resource.
+  are validated (a scoped `[fe80::1%eth0]` is rejected: RFC 3986 IPv6address
+  has no ZoneID, per RFC 9844). Mostly a tightening on malformed input to an
+  opt-in feature; it also stops rejecting a valid path-empty absolute URI
+  (RFC 3986 §3, e.g. `about:`). No audience bypass or escalation.
 
 ## [2.8.0] - 2026-08-29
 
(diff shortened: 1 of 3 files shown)


cdelmonte-zg:
Both blocking points fixed (9c956c9).

1. **path-empty absolute URIs are now accepted.** The final `bool(netloc or path)` artificially rejected a scheme-only absolute URI, which contradicts closing #257 as "full RFC 3986": absolute-URI = `scheme ":" hier-part [ "?" query ]` and hier-part may be **path-empty** (RFC 3986 §3/§4.3, e.g. `about:`), and RFC 8707 asks only for an absolute URI without a fragment. Dropped the requirement - the scheme presence is already enforced by `_SCHEME_RE` (so a scheme-less relative reference is still rejected). New cases: `about:` and `foo:?q=1` -> valid.

2. **Scoped IPv6 is now rejected.** You are right, and thanks for the RFC 9844 pointer: RFC 3986 `IP-literal = "[" ( IPv6address / IPvFuture ) "]"` and IPv6address has no ZoneID; `ipaddress.IPv6Address` accepts the RFC 6874 scoped form `fe80::1%eth0`, which RFC 9844 (2025) obsoleted for URIs. The validator now rejects `%` in the IP-literal inner before handing it to `IPv6Address`. New case: `https://[fe80::1%eth0]/` -> invalid.

CHANGELOG adjusted (it also notes the path-empty acceptance now, so the entry no longer says "tightening only"). Suite 1841, ruff and mypy clean.

PR #271, opened 2026-08-29
feat: public clients on the device flow (#255, RFC 8628)

### LINKED ISSUE #255: Public clients on the device flow (RFC 8628)

cdelmonte-zg:
Third and last follow-up for 3.0.0. **Additive** (enables a capability). Closes #255.

**What.** A public client (`token_endpoint_auth_method: "none"`) can now use the device authorization grant. RFC 8628 §3.1/§3.4 contemplate exactly this: a public CLI/TV/IoT-style client presents its `client_id` alone at `/device_authorization` and again when polling `/token`, with no secret. #188 shipped public clients for authorization_code + PKCE; this completes the pair. (#262 had made the confidential-only refusal explicit, with a pointer here.)

**How it works.** The core change is the `/device_authorization` auth gate: a public client is accepted by `client_id` alone (and RFC 8628 §3.1-precisely: presenting HTTP Basic or a `client_secret` is rejected), the same shape as `/revoke`'s relaxation. Everything else on the flow already worked:
- the issued `device_code` is bound to the `client_id`, and the store's `poll()` enforces it (`WRONG_CLIENT`);
- the `/token` client-auth boundary (#188) already lets a public client through the non-`client_credentials` device grant.

Confidential clients still authenticate with their registered method (via `_enforce_registered_client_auth`). No PKCE (not part of the device grant).

**Bounded state (the abuse item #255 scoped).** Because a public client can now create device authorizations with its `client_id` alone, `DeviceCodeStore` is capacity-bounded (`MAX_PENDING_DEVICE_CODES = 10000`): at the cap `create()` raises `DeviceCodeStoreFull` and the endpoint returns a plain `503` with `Retry-After` (not an OAuth error code - RFC 6749 §5.2 has no registered code for saturation, and `slow_down` is the token endpoint's polling signal), refusing new authorizations rather than growing without bound or evicting a live one.

**Tests.**
- The #262 public-device rejection test is flipped to accept (`200` + `device_code`).
- A full public device flow: `device_authorization` -> approve at `/device` -> poll `/token`, `client_id` only, no secret -> `access_token`.
- A binding test: a `device_code` issued to the public client cannot be redeemed by another client (`invalid_grant`); Basic and a body secret on a public client are each rejected (`401`); the store raises at capacity and the endpoint returns a plain `503` (no fabricated OAuth error).
- **e2e**: `e2e/test_agent.py` gains `test_public_client_device_flow` against the bundled `mcp-public-client` - verified live, **OAuth2/OIDC 27/27, TOTAL 61/61**.
- Unit suite **1850**, `ruff` and `mypy` clean. The #262 CHANGELOG note is corrected (`/device_authorization` now accepts public clients).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,8 +74,8 @@ previous leniency allowed - needs a one-time adjustment.
   authorization_code + PKCE, and this completes the pair for CLI/TV/IoT-style
   clients. Because an unauthenticated device authorization request is cheaper to
   spam, the in-memory device-code store is now capacity-bounded: at the cap it
-  returns `503` `slow_down` rather than growing without bound or evicting a live
-  authorization.
+  returns `503` `server_error` rather than growing without bound or evicting a
+  live authorization.
 - **Mock protected MCP server as an e2e fixture** (#191). `e2e/mock_mcp_server.py`
   is a minimal MCP Streamable HTTP resource server (the `mcp` SDK's
   resource-server mode) with three scope-gated tools (`read_document` /
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,8 +74,9 @@ previous leniency allowed - needs a one-time adjustment.
   authorization_code + PKCE, and this completes the pair for CLI/TV/IoT-style
   clients. Because an unauthenticated device authorization request is cheaper to
   spam, the in-memory device-code store is now capacity-bounded: at the cap it
-  returns `503` `server_error` rather than growing without bound or evicting a
-  live authorization.
+  returns a plain `503` (with `Retry-After`) rather than growing without bound or
+  evicting a live authorization - not an OAuth error code, since RFC 6749 §5.2
+  has no registered code for server saturation.
 - **Mock protected MCP server as an e2e fixture** (#191). `e2e/mock_mcp_server.py`
   is a minimal MCP Streamable HTTP resource server (the `mcp` SDK's
   resource-server mode) with three scope-gated tools (`read_document` /
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,9 +39,10 @@ previous leniency allowed - needs a one-time adjustment.
   - **Two authentication methods in one request are rejected** (HTTP Basic and a
     body `client_secret` together, RFC 6749 §2.3) - Basic no longer silently
     wins. *Migration:* send credentials one way only.
-  - Public-client policy is unchanged: `/introspect` and `/device_authorization`
-    still refuse public clients; `/revoke` keeps its RFC 7009 §2.1 ownership
-    relaxation.
+  - Public-client policy: `/introspect` refuses public clients (RFC 7662),
+    `/revoke` keeps its RFC 7009 §2.1 ownership relaxation, and
+    `/device_authorization` now accepts a public client by client_id alone
+    (RFC 8628, see the device-flow entry below).
 - **`/authorize` reports errors after `redirect_uri` validation by redirecting
   to the client** (#189, RFC 6749 §4.1.2.1 / RFC 9207): `unsupported_response_type`,
   `invalid_scope`, PKCE errors and `invalid_target` are now `302` redirects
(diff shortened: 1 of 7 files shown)


cdelmonte-zg:
Both points fixed (612fe0b).

**Blocker - bounded state / abuse surface** (the item #255 explicitly scoped). You are right on both counts: the global limiter has `default_limits=[]` and no per-route limit, and `DeviceCodeStore.create()` only pruned expired entries - unbounded. Since there is no established `@limiter.limit` pattern here (the limiter is lazy/`Optional`, routes are on blueprints imported before it exists), I took the capacity-bound option you suggested, which is self-contained and independent of IP/proxy: `DeviceCodeStore` now has a hard cap (`MAX_PENDING_DEVICE_CODES = 10000`). At the cap, `create()` raises `DeviceCodeStoreFull` and `/device_authorization` returns `503 slow_down` (RFC 8628 §3.5 - the honest signal), **refusing new authorizations rather than growing without bound or evicting a live one** (so a legitimate in-flight code is never dropped).

**Minor - Basic on a public client.** Fixed: the public branch now rejects a request that carries HTTP Basic or a body `client_secret` - RFC 8628 §3.1 says a public client provides its `client_id` as a parameter. The `client_id` is public, so this is channel correctness (matching how `/token` enforces the registered method), not secrecy. Confidential clients still go through `_enforce_registered_client_auth`.

Tests added: `DeviceCodeStore` raises at capacity + the endpoint returns `503 slow_down`; Basic and a body secret on a public client are each rejected (`401`). CHANGELOG notes the bounded store and the §3.1 rejection. Suite 1850, ruff and mypy clean.

cdelmonte-zg:
Fixed (b230c5f). You are right: `slow_down` is RFC 8628 §3.5's **token endpoint polling** response (increase the poll interval), not a device-authorization capacity error - the device authorization response's errors follow RFC 6749 §5.2. The capacity 503 now carries `server_error` with a capacity message, and the misleading code comment is corrected (it explicitly says why not `slow_down`). Status stays 503; test renamed and asserts `server_error`.

Also refreshed the stale PR body (it still said "the only code change is the auth gate", suite 1846) - it now describes the capacity bound and the §3.1 Basic/secret rejection, suite 1850.

cdelmonte-zg:
Fixed (1131db9), and thanks for catching your own earlier steer - you are right. RFC 8628 §3.2 takes the RFC 6749 §5.2 (token endpoint) error form, and §5.2 has no registered code for saturation: `server_error` is authorization-endpoint-only in IANA, and `slow_down` is the token endpoint polling signal (§3.5). So there is no "almost right" third OAuth code to reach for.

The capacity response is now a **plain HTTP 503** - the correct semantics for a temporary capacity exhaustion - with a `message` and a `Retry-After` hint, and **no `error` field** (so it never claims to be a valid OAuth token error). The test asserts `error` is absent, `message` is present, and `Retry-After` is set; CHANGELOG and the code comment say why. The capacity bound itself (prune-then-check, no eviction of live grants, bounded memory) is unchanged.

cdelmonte-zg:
Added a doc commit (3b5a866): a **public-client policy table** in the configuration reference. It came out of the post-follow-ups architecture check - the code answers "how does nanoidp treat public clients?" across five endpoints, each following its own RFC (so the policy genuinely differs and should stay per-endpoint in code), and a reader deserves to see the whole picture in one place. The table covers `/authorize`, the `/token` grants, `/device_authorization`, `/revoke`, `/introspect`; the prose is also updated to include `/device_authorization` now accepting a public client (which this PR introduces), since that paragraph would otherwise be stale on merge. Doc-only, no code change - flagging it since the PR was already LGTM.

PR #272, opened 2026-08-29
fix: /api/users/<u>/token honours #73 client binding (access-point parity)

cdelmonte-zg:
Access-point parity gap found while auditing whether #73 (`refresh token needs a client_id binding`) reached every token minter.

**The gap.** `#73` / PR #269 fixed the MCP `generate_token` tool (optional `client_id`, no refresh token when unbound) but its **HTTP twin** - `POST /api/users/<username>/token`, which the dashboard Token Tester (`test.html`) and the per-user "generate token" button (`users.html`) call - still minted via `create_token` with no `client_id`. So it kept returning a `refresh_token` that #73 now refuses on use: a dead credential.

**Fix.** Mirror the MCP tool exactly:
- optional `client_id` in the JSON body, which **must name a real client** (else `400`);
- `issue_refresh_token = client_id is not None`, so an unbound token gets an access token with **no** refresh token, and a bound one gets a spendable refresh token.

**No caller breaks.** The e2e agent (`test_api_direct_token`) and both UI callers only read `access_token`; none reference `refresh_token`.

**Tests.** `TestApiGenerateTokenBinding`: unbound -> no `refresh_token`; bound (`client_id`) -> the refresh token actually works on `/token` (200); unknown `client_id` -> `400`. Existing `/api/users/<u>/token` issuer tests unaffected. `ruff`/`mypy` clean.

Small, independent of the 3.0.0 device-flow PR; goes into 3.0.0 with the rest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,11 @@ previous leniency allowed - needs a one-time adjustment.
   tool gains an optional `client_id` (which must name a real client) that binds
   the minted token and issues a refresh token spendable by it; without it the
   tool mints an unbound access token with no refresh token at all, rather than
-  hand back one that could not be spent.
+  hand back one that could not be spent. The HTTP testing endpoint
+  `POST /api/users/<username>/token` gets the same treatment - a new optional
+  `client_id` binds the token, and it no longer returns a `refresh_token` when
+  unbound - so all three token minters (a grant, the MCP tool, this endpoint)
+  agree.
 
 ### Added
 - **Mock protected MCP server as an e2e fixture** (#191). `e2e/mock_mcp_server.py`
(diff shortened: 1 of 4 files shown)


cdelmonte-zg:
Docs added (40fe317). You are right - this is the third token minter and its default response contract changed, so it should not be left to reconstruct from code. Two lines added:

- **CHANGELOG**: the #73 breaking bullet now names `POST /api/users/<username>/token` alongside the MCP tool (new optional `client_id`; no `refresh_token` when unbound), closing with "all three token minters - a grant, the MCP tool, this endpoint - agree."
- **REST reference** (`book/src/reference/endpoints.md`): the row now spells out the `exp_minutes` / `client_id` body and the refresh-token-only-when-bound contract.

No code change from the review.

PR #274, opened 2026-08-30
feat: per-client horizontal /authorize login layout (#249)

dshvedchenko:
Adds OAuthClient.layout ("vertical" default, "horizontal"), following
the codebase's field-flow: model, config_documents, client_to_yaml/
merge_client_entry (omit-at-default, #127), the clients form, UI
create/edit routes, and the MCP create_client/update_client tools
(pre-validated before assignment, same rule as token_endpoint_auth_method).
authorize.html splits the client-info and login-form blocks into local
macros so horizontal can lay them out side by side in a Bootstrap
row/col-md-6 (collapses to vertical below md) while the header/footer
stay full width, with no new static asset.

Covers the imperative legs test_client_field_parity.py doesn't reach:
extends the parametric test_client_contract.py fixtures, adds
TestClientLayout to test_mcp_tool_branches.py, and extends
e2e/test_agent.py's Client Branding check to assert the two-column
markup for horizontal and its absence for vertical. Regenerates
docs/schema/config.v1.json and documents the field in
configuration.md and the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,15 @@ previous leniency allowed - needs a one-time adjustment.
   hand back one that could not be spent.
 
 ### Added
+- **Horizontal `/authorize` login card composition** (#249). New per-client
+  `layout` field, `"vertical"` (default, unchanged) or `"horizontal"`: the
+  latter places the client info block and the login form side by side in a
+  Bootstrap two-column split, with the header and footer still full width,
+  collapsing back to the single-column stack on narrow viewports. One of
+  exactly two nanoidp-owned layouts - no per-client CSS or column widths.
+  Full support across settings.yaml, the UI client form, and the MCP
+  `create_client`/`update_client` tools; omitted (or `"vertical"`) writes
+  nothing to YAML, matching every other default-valued client field.
 - **Mock protected MCP server as an e2e fixture** (#191). `e2e/mock_mcp_server.py`
   is a minimal MCP Streamable HTTP resource server (the `mcp` SDK's
   resource-server mode) with three scope-gated tools (`read_document` /
--- a/book/src/reference/configuration.md
+++ b/book/src/reference/configuration.md
@@ -281,6 +281,9 @@ oauth:
       footer_color: "#e8f4f8"      # optional; hex only, the card's footer band
       show_client_id: true         # optional; default true
       show_description: true       # optional; default false
+      layout: "horizontal"         # optional; "vertical" (default) or "horizontal" (#249):
+                                    # horizontal places the client info and login form side by
+                                    # side, collapsing back to a single column on narrow viewports
     - client_id: "scoped-client"
       client_secret: "secret"
       description: "Client restricted to a scope subset"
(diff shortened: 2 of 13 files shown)


cdelmonte-zg (approved):
Approved. This follows the #249 contract point by point, and I verified it empirically at bfb4f57 rather than from the diff alone:

- Full suite 1858 green, mypy/ruff/import-linter clean, CI 9/9 including both e2e jobs.
- `docs/schema/config.v1.json` is byte-identical to a fresh `nanoidp config-schema` run.
- Live probe on a fresh init: a `layout: horizontal` client renders `class="authorize-card authorize-card-horizontal"` with the two `col-md-6` columns; a default client renders the plain card. Your e2e's choice to assert on the rendered class attribute instead of a substring is exactly right, since the CSS selectors put the string in every page's style block (3 matches on the vertical page).
- The #127 round-trip holds live: editing a client that never declared `layout` (the form always submits `layout=vertical` from the select) does not materialize the key, and switching horizontal back to vertical pops it from YAML.
- Negative: a crafted POST with `layout=diagonal` lands in the existing ValueError flash path (302 back to the form, nothing written), same as `token_endpoint_auth_method`; the MCP pre-validation test pins the no-partial-update guarantee.
- One leg no test exercises, so I probed it live: persona login mode plus horizontal. The persona picker macro resolves its context correctly inside the column and both persona buttons render.

Nice touches beyond the letter of the contract: the compact one-row header for the horizontal card, and excluding `layout` from the differ-in-every-field contract check with the same rationale as the booleans.

One scheduling note: #249 sits on the 3.1 milestone and main is in the 3.0.0 rc soak (freeze: regression/security/packaging/doc-contract only), so this merges after 3.0.0 final, same as the 2.9 stack did. Nothing for you to do; it will be merged from here once the final is tagged.

cdelmonte-zg:
Correction to the scheduling note in my review: this goes into **3.0.0** after all, not 3.1 - the change is additive and opt-in (default layout unchanged), so it does not invalidate what the rc soak is actually testing, and a 3.1 release just for this would be disproportionate. Merged to main; the next rc absorbs it. #249 and this PR moved to the 3.0.0 milestone.

cdelmonte-zg:
@dshvedchenko **3.0.0-rc2 is out** (`pip install --pre nanoidp==3.0.0rc2` / `ghcr.io/cdelmonte-zg/nanoidp:v3.0.0-rc2`) - and your #249 horizontal layout ships in it. Since rc1 the release also absorbed a full pre-release architecture audit, with more breaking tightening that is exactly worth exercising against a shared instance: every `/token` error is now RFC 6749 §5.2 JSON (grant failures that used to 401 are `invalid_grant` 400; `invalid_client` carries `WWW-Authenticate` only when Basic was attempted), the `exp` claim is required on every accepted token, and `stricter-dev` now really rate-limits `/token`. The release notes' breaking section is written as a migration guide. Anything surprising, especially from clients that parse token errors - a report here or a new issue is gold before the final.

cdelmonte-zg:
Heads-up: **v3.0.0-rc3** supersedes rc2 as the soak candidate (`pip install --pre nanoidp==3.0.0rc3`) - the earlier rc2-pinned commands in this thread are stale. Delta over rc2 is one addition (auto-login personas, #318); every rc2 breaking note still applies unchanged, so anything you already validated against rc2 carries over.

cdelmonte-zg:
@dshvedchenko **v3.0.0 is out** as a final: `pip install nanoidp` now resolves 3.0.0 and `:latest` moved to it, so the rc-pinned commands in this thread are stale. Your horizontal `/authorize` layout (#249) ships in it, credited in the [release notes](https://github.com/cdelmonte-zg/nanoidp/releases/tag/v3.0.0) and the CHANGELOG. Every rc2/rc3 breaking note applies unchanged to the final - the CHANGELOG's Breaking Changes section is the migration guide. Thanks for the contribution and for exercising the release candidates. Your #323 arrived after the soak window, so it is reviewed against the next minor.

PR #289, opened 2026-08-30
fix: cross-entry-point consistency for the 3.0 clean base (#275-#279)

### LINKED ISSUE #275: fix: /saml/attribute-query docstring claims JWT auth; fabricated attributes for unknown users
`/saml/attribute-query` (`routes/saml.py:834`) has two contract lies, found by the pre-3.0 architecture audit:

1. **The docstring claims "backend-to-backend attribute fetching after JWT authentication", but no JWT (nor anything else) is verified.** The endpoint is deliberately unauthenticated, consistent with the #163 decision that read surfaces are never gated - but the prose must say that, not the opposite. A reader deploying on a shared instance deserves the truth.
2. **Unknown users get fabricated attributes** (`saml.py:917-925`): a query for a nonexistent NameID returns `email: <user>@example.com, identity_class: INTERNAL, entitlements: DOCUMENT_READ` in a signed assertion. A signed assertion asserting invented facts about a principal that does not exist is a trap for the integration being tested - the SP under test passes with data nanoidp made up.

Fix (3.0.0, behavior change on point 2):
- Docstring and docs state the endpoint is unauthenticated by design (same read model as the REST API) and what that implies on shared instances.
- Unknown NameID returns a SAML error status (`urn:oasis:names:tc:SAML:2.0:status:UnknownPrincipal`) instead of a fabricated assertion; CHANGELOG entry under Breaking Changes with the old behavior described.

Question to settle in review: whether the read-surfaces-ungated model itself (decided in #163) should be revisited for 3.0 is a separate decision - this issue only makes the code honest within the current model.

### LINKED ISSUE #276: fix: device_code grant redeems scope without re-validation at /token (unlike authorization_code and refresh)
Found by the pre-3.0 audit. The three grants that redeem a previously authorized artifact treat a stale scope differently:

- `authorization_code` redemption re-validates the stored scope with `resolve_scope(..., validate_only=True)` (`routes/oauth_grants.py:573`) so a scope removed from the client's `allowed_scopes` between authorization and redemption is rejected;
- `refresh_token` does the same (`oauth_grants.py:296`);
- **`device_code` does not**: `_grant_device_code` passes `grant.scope` verbatim into the outcome (`oauth_grants.py:786`). A scope revoked from `allowed_scopes` (or removed from `scopes_supported`) between `/device_authorization` and the poll is still minted.

Fix (3.0.0, behavior change): re-validate the stored device grant scope with `validate_only=True` at redemption, mirroring the other two paths, with the same `invalid_scope` JSON error; regression test with the revoke-between-authorize-and-poll sequence; CHANGELOG entry.

### LINKED ISSUE #277: fix: client_id precedence and mismatch handling diverge between /token and /introspect, /revoke, /device_authorization
Found by the pre-3.0 audit. `/token` resolves the client identity as `body_client_id or auth_client_id` and **rejects a mismatch** between the Basic username and a body `client_id` (`routes/oauth.py:777`, `:790-800`). The other three client-authenticated endpoints do the opposite and check nothing:

- `/introspect` (`oauth.py:1086`), `/revoke` (`oauth.py:1197`), `/device_authorization` (`oauth.py:1375`) all use `(auth.username if auth else None) or body_client_id` - header silently wins, and a request carrying Basic auth for client A plus `client_id=B` in the body is processed as A with no error.

Not a vulnerability (authentication is still enforced against the authenticated identity), but the same request means different things at different endpoints, and the divergence lives in four copies of the same preamble (each one re-reads the form, re-derives the id, re-fetches the client).

Fix (3.0.0, behavior change): one shared identity-resolution helper used by all four endpoints with `/token`'s semantics (mismatch = 401 invalid_client); negative tests per endpoint (Basic A + body B); CHANGELOG entry under Breaking Changes.

### LINKED ISSUE #278: refactor: enforce the #73 mint-side guard in TokenService.create_token instead of duplicating it in callers
Found by the pre-3.0 audit: the #73 invariant ("no refresh token without a client_id binding") is enforced on the spend side in one place (`routes/oauth_grants.py:190-213`) but on the **mint side it lives in the callers, in duplicate**:

- `routes/api.py:104` - `issue_refresh_token=client_id is not None`, with its own existence check;
- `mcp_server.py` `_tool_generate_token` - the same line and the same check, with a near-identical prose comment.

`TokenService.create_token` itself accepts `issue_refresh_token=True, client_id=None` and would happily mint a refresh token that `/token` then unconditionally rejects - a dead credential. Today no caller does this; nothing but review discipline keeps it that way. This is exactly the #269/#272 shape (rule enforced at N-1 of N surfaces).

Fix (3.0.0, behavior-preserving): `create_token` raises `ValueError` on `issue_refresh_token=True` without `client_id` - the service owns the invariant, callers keep their friendly errors; unit test for the backstop; the two caller comments shrink to a reference.

### LINKED ISSUE #279: docs: document the simulation boundary of MCP generate_token (scope) and verify_token (revocation)
Two MCP token tools intentionally diverge from their HTTP counterparts, but only one divergence is documented. From the pre-3.0 audit:

1. **`generate_token` bypasses the scope machinery entirely**: `scope` is passed raw to `create_token`, with no `scopes_supported` vocabulary check and no `allowed_scopes` ceiling even when `client_id` is supplied. For `resource` the same bypass carries a deliberate BOUNDARY comment ("a testing / simulation affordance, not an OAuth grant" - `mcp_server.py:1364-1374`); for `scope` there is nothing. The rationale is identical (minting an out-of-ceiling token is precisely how you test a resource server's rejection path), so the stance should be identical and explicit: same boundary comment, a sentence in the tool description, and a test pinning the exemption.
2. **`verify_token` checks signature/expiry/audience but not revocation and not `token_use`** (`mcp_server.py:1401-1418`): a revoked access token or an ID Token presented as an access token reports `valid: true` where `/introspect` says `active: false`. This matches the #191 design (a JWKS-validating resource server does not see revocation either - revocation is `/introspect`'s answer), but the tool description does not say so. Document it: `verify_token` simulates a stateless resource server; ask `/introspect` (or a future introspect tool) for revocation.

Docs + tests only, no behavior change. 3.0.0 so the contract ships with the major.

cdelmonte-zg:
Implements five findings from the pre-3.0 architecture audit, all on one theme: **the same rule must mean the same thing at every entry point**. All are 3.0.0 material - three are behavior changes that belong inside the major.

Closes #275, closes #276, closes #277, closes #278, closes #279.

## Behavior changes (CHANGELOG under Breaking Changes)

- **#277 - one request, one client identity.** `/introspect`, `/revoke` and `/device_authorization` now resolve the requesting client through the same `_request_client_identity()` helper as `/token`: HTTP Basic naming client A plus `client_id=B` in the body is rejected with `invalid_client` everywhere. Previously those three endpoints resolved header-first and silently ignored the contradictory body value. The rejected `/revoke` request provably revokes nothing (pinned).
- **#276 - device_code grant re-validates scope at redemption.** A scope removed from the client's `allowed_scopes` (or the vocabulary) between `/device_authorization` and the poll now fails with `invalid_scope`, mirroring the `authorization_code` and `refresh_token` re-checks (`validate_only=True`, absent-stays-absent per the #186 B1 rule).
- **#275 - `/saml/attribute-query` stops inventing principals.** Unknown NameID now gets a SAML error status (`Requester`/`UnknownPrincipal`, no assertion) instead of a **signed assertion with fabricated attributes**. The docstring claimed "after JWT authentication" while verifying nothing; docs and docstring now state plainly that the endpoint is unauthenticated by design, per the #163 read model. Whether that read model itself should change is deliberately NOT touched here - flagging it as a separate question.

## Behavior-preserving

- **#278 - the #73 mint-side rule lives in the service.** `TokenService.create_token`'s `issue_refresh_token` now defaults to "only when the token is bound" and raises on explicit `True` without `client_id`; the duplicated caller guards in `routes/api.py` and `mcp_server.py` are gone. The three minting surfaces behave exactly as before.
- **#279 - the simulation boundary is documented and pinned.** `generate_token`'s tool description now states it applies no scope/resource ceiling even with `client_id` (same deliberate stance the #187 review put on `resource`, now extended to `scope` with the same BOUNDARY comment); `verify_token`'s states it verifies like a stateless resource server (no revocation, no `token_use` - `/introspect` is the revocation answer, per the #191 model). Both exemptions have pin tests so a future "fix" that starts enforcing shows up as a deliberate contract change.

## Verification

- 1872 unit tests green (14 new; 3 pre-existing tests updated because they encoded the pre-#73 unbound-refresh-token pattern), mypy/ruff/import-linter clean.
- Live probes on the shipped config: mismatch 401 at all three endpoints with the token provably NOT revoked after a rejected `/revoke`; matching body `client_id` still accepted; unknown-NameID attribute query returns `UnknownPrincipal` with no assertion while a known user still gets all attributes; `/api/users/<u>/token` unbound has no `refresh_token`, bound has one.
- e2e `test_agent.py` extended (same PR, per house rule): contradictory `client_id` negative in Token Introspection, UnknownPrincipal negative in SAML Attribute Query - 61/61 live.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,43 @@ previous leniency allowed - needs a one-time adjustment.
   `client_id` binds the token, and it no longer returns a `refresh_token` when
   unbound - so all three token minters (a grant, the MCP tool, this endpoint)
   agree.
+- **One request, one client identity** (#277). `/introspect`, `/revoke` and
+  `/device_authorization` now resolve the requesting client exactly as
+  `/token` always has: HTTP Basic naming client A plus `client_id=B` in the
+  body is one request claiming two identities and is rejected with
+  `invalid_client` (previously the Basic username silently won and the body
+  value was ignored). *Migration:* drop the contradictory body `client_id`,
+  or make it match the authenticated client.
+- **The `device_code` grant re-validates the stored scope at redemption**
+  (#276): a scope removed from the client's `allowed_scopes` (or from
+  `scopes_supported`) between `/device_authorization` and the poll now fails
+  the poll with `invalid_scope`, exactly as the `authorization_code` and
+  `refresh_token` grants already re-check theirs. Previously the stale scope
+  was still minted.
+- **`/saml/attribute-query` answers an unknown NameID with a SAML error
+  status** (#275): top-level `Requester` with subordinate `UnknownPrincipal`,
+  no assertion. Previously an unknown user got a **signed assertion with
+  fabricated attributes** (`<user>@example.com`, `identity_class: INTERNAL`,
+  `entitlements: DOCUMENT_READ`) - an SP under test would pass with data
+  nanoidp invented. The endpoint's docs now also state plainly that it is
+  unauthenticated by design (the same read model as the REST API, #163):
+  it previously claimed "after JWT authentication", which the code never did.
+
+### Changed
+- `TokenService.create_token` now owns the #73 mint-side rule (#278):
+  `issue_refresh_token` defaults to "only when the token is bound"
+  (`client_id` given), and asking for a refresh token without a binding
+  raises instead of minting a credential `/token` would reject. The three
+  minting surfaces (grants, MCP `generate_token`,
+  `POST /api/users/<username>/token`) behave as before; the rule just lives
+  in one place.
+- The MCP `generate_token` and `verify_token` tool descriptions now document
+  their simulation boundary (#279): `generate_token` stamps `scope` and
+  `resource` as given with no vocabulary check and no per-client ceiling
+  (minting an out-of-ceiling token is how a resource server's rejection path
+  gets tested), and `verify_token` checks signature/expiry like a stateless
+  resource server - revocation is `/introspect`'s answer. Behavior is
+  unchanged; both exemptions are now pinned by tests.
 
 ### Added
 - **Horizontal `/authorize` login card composition** (#249). New per-client
--- a/book/src/reference/endpoints.md
+++ b/book/src/reference/endpoints.md
@@ -30,10 +30,17 @@ in the tokens - see [Tokens and claims](tokens.md#where-do-the-email--profile-cl
 | `GET /saml/metadata` | IdP Metadata |
 | `GET /saml/cert.pem` | IdP signing certificate (PEM) |
 | `GET/POST /saml/sso` | Single Sign-On (supports both HTTP-POST and HTTP-Redirect bindings) |
-| `POST /saml/attribute-query` | Attribute Query |
+| `POST /saml/attribute-query` | Attribute Query (SOAP, backend-to-backend) |
 
 Bindings, strict-binding mode, response signing, and canonicalization are
 covered in [SAML options](saml.md).
+
+`/saml/attribute-query` is **unauthenticated by design** - the same model as
+the REST read surfaces (reads are never gated): nanoidp is a testing IdP and
+its user directory is test data. On a shared instance, anyone who can reach
+the endpoint can read any configured user's attributes; deploy accordingly.
+An unknown NameID gets a SAML error status (`Requester`/`UnknownPrincipal`),
+never a fabricated assertion.
 
 ## REST API
 
--- a/e2e/test_agent.py
+++ b/e2e/test_agent.py
@@ -1938,11 +1938,22 @@ class NanoIDPTestAgent:
                 active = data.get("active", False)
                 username = data.get("username", "?")
                 scope = data.get("scope", "?")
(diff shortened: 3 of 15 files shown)


cdelmonte-zg:
Round 1 finding fixed in 309dca6. Confirmed real and reproduced red first: with `saml_sign_responses: true` the success response was signed and the new UnknownPrincipal error response skipped `_sign_attribute_query_response` entirely. The error path now goes through the same signing gate as the success path. Regression tests both ways (signing on: `ds:Signature` present on the error response; signing off: absent), and verified live against the shipped config (SignatureValue present). Suite 1874, statics clean.

Issue #284, opened 2026-08-30
test: user-field parity across the nine user shapes


cdelmonte-zg:
The client record has parity coverage (`test_client_field_parity.py` + `test_client_contract.py`). The **user record has none**, and it is the widest family in the codebase - the pre-3.0 audit counted nine shapes that must stay in sync by hand:

1. `User` (`models.py:73`), 2. `User.to_dict()` (`:110`), 3. `UserEntry`/`_USER_KNOWN_FIELDS`/`to_user()` (`config_documents.py:438`), 4. `user_to_yaml()` (`serialization.py:393`), 5. `_user_to_dict()` (MCP, `mcp_server.py:195`), 6. `_build_user_from_arguments()` + create schema (`:211`, `:291`), 7. `update_user` schema (`:440`), 8. UI create/edit form parsing (`ui.py:220`, `:329`), 9. `GET /api/users/<u>` (`api.py:59`).

The drift is not hypothetical: shape 7 already lost `attributes` silently (#280).

Proposal (3.0.0, tests only): a user-field parity test mirroring the client one - field-name-set equality across the declarative shapes, plus a round-trip contract (fixtures with every field non-default pushed through YAML, UI POST, and MCP legs and read back), with explicitly documented exclusions (e.g. `password` elision on read surfaces, `extra="allow"` attributes passthrough). Acceptance: adding a field to `User` fails until every shape carries it or excludes it on purpose.

Issue #286, opened 2026-08-30
refactor: decompose mcp_server.py before the Streamable HTTP transport (#193)


cdelmonte-zg:
`mcp_server.py` is 2,032 lines in one file with nine distinct responsibilities (process/transport bootstrap; its own ConfigManager reference; readonly+admin-secret authz; audit emission; its own client/user serializers; its own normalization layer; 767 lines of hand-written JSON Schema; dispatch/validation/error shaping; 25 tool handlers holding domain logic). The pre-3.0 audit's churn analysis puts it in 41 of the last 100 src-touching commits and in all five of the top co-changing file pairs: it is a second, parallel front end that every domain change must edit twice (handler + schema).

#193 (Streamable HTTP) means adding a second transport to this file. The stdio-shaped responsibilities (bootstrap, config wiring, authz gate, audit) are interleaved with the transport-neutral ones, so the transport work would either fork the file or finally split it. Split it first:

- transport/bootstrap vs tool table vs handlers vs shared normalization as separate modules;
- derive the `_TOOLS` JSON Schema from the Pydantic models' own `Field(description=...)` metadata where possible instead of re-typing it (the audit found the descriptions literally duplicated, e.g. hex colors `mcp_server.py:750` vs `models.py:178`);
- the `routes._auth.verify_secret` import (which drags Flask into the stdio process, `mcp_server.py:67`) moves to a neutral home.

No milestone: this is the prerequisite refactor for #193 and should land immediately before it, behavior-preserving, guarded by the existing MCP test surface (`test_mcp.py`, `test_mcp_tool_branches.py`, mcp_smoke_test).

cdelmonte-zg:
One more piece for this decomposition, from the #290 review: `__main__.py` currently imports `wizard._validate_and_write` (validate a rendered template through the document models + atomic text write) - an entry point reusing another entry point's private helper. Works, but the behavior deserves a neutral home when this issue restructures the module layout (a small `config_bootstrap` seam, or wherever the shared write-path helpers land).

PR #290, opened 2026-08-30
fix: three audit bugs for the 3.0 clean base (#280 #281 #282)

### LINKED ISSUE #280: fix: MCP update_user cannot update attributes (schema and handler lack the field)
Found by the pre-3.0 audit, an already-happened silent drift in the user field flow: the MCP `update_user` tool schema has **no `attributes` property** and `_tool_update_user` has no `attributes` branch, while `create_user` accepts it (`mcp_server.py:329`, applied at `:231`) and `_user_to_dict` returns it (`:207`). An agent can create a user with custom attributes but can never update them - the field silently never lands.

Fix (3.0.0, bug): add `attributes` to the `update_user` schema and handler (scratch-copy validation ordering as the handler already does for other fields), test the round trip, and note it in the CHANGELOG under Fixed. The systematic guard for this class is the user-field parity test tracked separately.

### LINKED ISSUE #281: fix: get_crypto_service silently ignores keys_dir once the singleton exists
`get_crypto_service(keys_dir)` ignores its argument once the singleton exists (`services/crypto.py:475-483`): first call wins, every later call gets the existing instance no matter what `keys_dir` it asked for. All 19 call sites pass `config.settings.keys_dir`, which suggests they believe the argument matters. After a config reload that changes `keys_dir`, every consumer silently keeps signing and serving JWKS from the old directory.

Found by the pre-3.0 audit (also noted: `TokenService.__init__` caches `get_config()` at construction, `services/token.py:105`, with the same staleness shape - documented in `conftest.py:90` as a known hazard).

Fix (3.0.0, bug): make `get_crypto_service` compare the requested `keys_dir` with the live instance's and recreate under the existing lock when it differs (external-key initialization via `init_crypto_service` stays authoritative: a keys_dir-only recreation must not discard configured external keys - decide precedence in the PR); regression test: reload with changed `keys_dir`, JWKS must serve the new keys. The broader singleton-lifecycle question stays with #230; this fixes the one accessor that takes an argument and lies about it.

### LINKED ISSUE #282: fix: wizard writes YAML with raw open(), bypassing the atomic writer and validation
The wizard writes both YAML files with raw `open(..., "w")` and f-string blobs (`wizard.py:219`, `:263`), bypassing everything the rest of the write path guarantees: the atomic replace + backup of `config_writer`, the process-wide write lock, and any validation. A wizard run concurrent with a running instance (or a crash mid-write) can leave a torn file that the atomic writer exists to prevent, and a typo in the template reaches disk unvalidated.

Found by the pre-3.0 audit as part of the "six places define the default configuration" family.

Fix (3.0.0): route the wizard's writes through the shared atomic write primitive and run the resulting files through the document models before declaring success (`validate-config` already exists as the check); test that a wizard-produced directory loads cleanly with strict validation.

cdelmonte-zg:
Three bugs found by the pre-3.0 architecture audit, each verified live before and after. Closes #280, closes #281, closes #282.

- **#280 - MCP `update_user` can now update `attributes`.** The field was accepted by `create_user` and returned by `_user_to_dict`, but the `update_user` schema and handler silently lacked it: an agent could set custom attributes at creation and never change them again. The handler applies it on the same validate-on-scratch-copy path as every other field; the new mapping replaces the whole object, consistent with the tool's replace semantics. This is the concrete drift that motivates the user-field parity test (#284).
- **#281 - `get_crypto_service` honours `keys_dir` on every call.** Once the singleton existed, the argument was silently dropped: after a config reload changing `keys_dir`, all 19 call sites kept signing and serving JWKS from the old directory. The service is now recreated (under the existing lock) when the requested directory differs. Operator-provided external keys stay authoritative: a service created by `init_crypto_service` with external PEMs is never discarded over a `keys_dir` divergence (new `uses_external_keys` property, pinned by test). Verified live: `jwt.keys_dir` edit + `POST /api/config/reload` now rotates the served JWKS kid; before the fix the kid stayed stale.
- **#282 - the wizard and `init` write configuration atomically and validated.** Both used raw `open(..., "w")` with f-string templates: no atomicity (a crash could leave a torn file) and no validation (a template error reached disk unseen). They now validate the rendered text through the document models BEFORE anything touches disk and write via a new `serialization.atomic_write_text` (same temp-then-replace shape as `atomic_write_yaml`, byte-exact so the commented templates survive verbatim). Negative pinned: an invalid template raises and writes nothing.

Verification: 1866 unit tests green (8 new), mypy/ruff/import-linter clean; live probes: fresh `init` loads and validates clean, keys_dir reload probe above.

Note: independent of #289 (different hunks of `mcp_server.py`); either merges first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -227,6 +227,22 @@ previous leniency allowed - needs a one-time adjustment.
 ## [2.8.0] - 2026-08-29
 
 ### Fixed
+- **MCP `update_user` can now update custom `attributes`** (#280): the field
+  was accepted by `create_user` and returned by every read surface, but the
+  `update_user` schema and handler silently lacked it - an agent could set
+  attributes at creation and never change them again. The new mapping
+  replaces the whole `attributes` object, like every other field there.
+- **`get_crypto_service` honours `keys_dir` on every call** (#281): once the
+  singleton existed the argument was silently ignored, so a config reload
+  that changed `keys_dir` kept signing tokens and serving JWKS from the old
+  directory. The service is now recreated when the requested directory
+  differs; operator-provided external keys (`init_crypto_service`) stay
+  authoritative and are never discarded over a `keys_dir` change.
+- **The wizard and `init` write configuration atomically and validated**
+  (#282): both used to write raw template text with `open()`, bypassing the
+  temp-then-replace primitive every other config writer uses - a crash could
+  leave a torn file, and a template error reached disk unvalidated. Both now
+  validate through the document models before anything touches disk.
 - **`client_credentials` no longer returns a refresh token** (#239). RFC
   6749 §4.4.3: "A refresh token SHOULD NOT be included" - the client
   authenticates itself on every request, and the token handed out was a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -224,8 +224,6 @@ previous leniency allowed - needs a one-time adjustment.
   opt-in feature; it also stops rejecting a valid path-empty absolute URI
   (RFC 3986 §3, e.g. `about:`). No audience bypass or escalation.
 
-## [2.8.0] - 2026-08-29
-
 ### Fixed
 - **MCP `update_user` can now update custom `attributes`** (#280): the field
   was accepted by `create_user` and returned by every read surface, but the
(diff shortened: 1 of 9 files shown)


cdelmonte-zg:
Round 1 finding fixed in 0791d99: the three entries sat under the already-released [2.8.0] Fixed heading - the changelog claimed they shipped in 2.8.0. [Unreleased] now carries its own Fixed section with them and [2.8.0] is back to exactly what was released. (Same stacked-edit family as #132/#152/#169; deserved a sharper eye from me given the history.)

On the non-blocker: agreed on the ownership smell - `__main__.py` reaching into `wizard._validate_and_write` works but the 'validate template + atomic write' behavior belongs in a neutral home. It fits naturally in the #286 decomposition wave (or a tiny `config_bootstrap` module alongside it); noted there rather than grown in this PR. #292 has been rebased onto this head, stack branch updated.

Issue #291, opened 2026-08-30
fix: UI users form corrupts non-string attributes on edit; parity test blind to the attr_key[]/attr_value[] widget


cdelmonte-zg:
Found while writing the user-field parity test (#284): custom `attributes` can be set via YAML and the MCP `create_user`/`update_user` tools (#280), and every read surface returns them, but the UI users form (`users_form.html` + the create/edit parsers in `routes/ui.py`) has no input for them - the one write surface that cannot touch the field. The parity test records this as a documented exclusion pointing here.

Proposal: a textarea accepting one `key: value` per line (the same shape `_parse_textarea_list` established for lists), parsed to a string mapping; blank clears. Ships whole per the client-field flow rules: both UI parsers, template, feature tests, e2e branding-style check.

cdelmonte-zg:
Premise correction, found while implementing: **the input exists**. The users form has had a dynamic attributes widget all along - `attr_key[]`/`attr_value[]` rows, prefilled on edit, parsed by both the create and edit routes. The audit (and the #284 parity test's exclusion) missed it because the template check's regex `name="([a-z_]+)"` cannot match the brackets, and neither the audit agent nor I read the template before asserting the gap. Lesson recorded: verify a 'missing surface' claim against the artifact itself, not a scanner's output.

What IS broken, reproduced empirically: **editing a user that has a non-string attribute silently corrupts it**. The template renders `value="{{ value }}"` - a list-valued attribute (settable via YAML and MCP) renders as its Python repr - so an untouched edit round-trip stores `{'teams': "['alpha', 'beta']"}` in place of the list. Retitled accordingly; the fix is JSON-convention rendering/parsing for container values plus a parity-test update that recognizes the widget instead of excluding the field.

PR #292, opened 2026-08-30
test+docs: parity contracts and the one-home rule (#283 #284 #285)

### LINKED ISSUE #283: test: access-point parity contract for token-issuing surfaces (closes the #269/#272 class)
The pre-3.0 audit found that no test pins cross-entry-point policy application: every policy (client binding #73, scope ceiling, resource ceiling, client auth) is verified by hand-enumerated per-endpoint tests whose completeness is a review property, not a suite property. That is the class #269/#272 came from: a capability exposed by N entry points, a rule updated at N-1 of them.

The repo already has the right pattern twice: `test_settings_plumbing_parity.py` and `test_discovery_parity.py` derive their expectations from a machine-readable registry so a new surface fails until accounted for. Token issuance has no such registry.

Proposal (3.0.0, tests only):

1. **Caller-set guard**: a test asserting that the set of `TokenService.create_token` call sites equals a declared registry (today: `routes/oauth.py`, `routes/api.py`, `mcp_server.py`). Mechanical (grep over `src/`), zero framework: adding a fourth minting surface fails the test until the surface is added to the registry and takes a stance on every policy below.
2. **Policy stance matrix**: a parametric test over (surface x policy) where each registered surface declares, per policy (mint-side #73 guard, scope ceiling, resource ceiling), either `enforced` (with the behavioral assertion run against that surface) or `exempt` (with a pointer to the documented boundary - e.g. MCP generate_token's simulation affordance). An undeclared combination fails.

Acceptance: removing any existing enforcement (e.g. the api.py #73 guard) turns the suite red; adding a new create_token caller turns it red until declared.

### LINKED ISSUE #284: test: user-field parity across the nine user shapes

### LINKED ISSUE #285: docs: domain-invariant review rule in CONTRIBUTING; correct the false circular-dependency comments
Two documentation debts from the pre-3.0 audit, both cheap and both about teaching the right model:

1. **A domain-invariant review rule in CONTRIBUTING**: "OAuth policy belongs in services/domain logic; routes and MCP are adapters and must not independently reinterpret the same rule. A PR implementing the same policy separately at two surfaces is an architectural smell to flag in review." One paragraph, plus a pointer from VISION. This is the review criterion for the 3.x series, not a refactoring mandate.
2. **Five deferred imports claim to avoid circular dependencies that do not exist** (`services/auth_code.py:110` says "late import to avoid circular dependency" - `config.py` never imports `services`; `services/device_code.py:25`; `routes/ui.py:780/:817/:836` and `routes/api.py:250/:279` defer an import their own module already does at top level). One deferred import IS justified (`services/audit.py:96`) but for re-entrancy of `_config_lock`, and its comment also mislabels the reason as circularity. False comments teach the next contributor a wrong module graph and make the one real constraint (`serialization` must not import the package at runtime, enforced by import-linter, #149) indistinguishable from cargo cult. Fix: hoist the fake ones to module top, correct the audit.py comment to name re-entrancy.

3.0.0, docs and comment-only changes plus import hoists with no behavior change.

cdelmonte-zg:
The discipline half of the pre-3.0 audit: two contract tests and the review rule. Closes #283, closes #284, closes #285.

## #283 - access-point parity for token issuance

`tests/test_token_issuance_parity.py`, following the table-driven precedent of `test_settings_plumbing_parity` and `test_discovery_parity`:

1. **Call-site registry**: the set of CALL SITES of `TokenService.create_token` - `file::enclosing_function`, discovered by AST walk (comments cannot fool it) - must equal `ISSUANCE_SURFACES`: `routes/oauth.py::token`, `routes/api.py::generate_token`, `mcp_server.py::_tool_generate_token`. A new minting call site - including a second function inside an already-registered module (review round 1) - fails the suite until registered; a collector self-test pins that exact scenario.
2. **Policy stance matrix**: every (call site x policy) pair - mint-side client binding, scope ceiling, resource ceiling - must declare `enforced` or `exempt: <reason>`. Enforced pairs have a behavioral assertion against that surface; exempt pairs pin the exemption itself (e.g. MCP `generate_token`'s #279 simulation boundary), so silently starting OR stopping enforcement both surface as deliberate contract changes. The stance strings and the assertions are linked by convention, deliberately - nine combinations do not justify a mini-framework (documented in place).

This turns the #269/#272 bug class ("rule updated at N-1 of N surfaces") from a review property into a suite property. Known theoretical limit, accepted in review: the collector matches any callee named `create_token`, not a proven `TokenService` receiver - a future same-named method would be a false POSITIVE with an immediate, legible failure, which is the right failure mode for a tripwire.

## #284 - user-field parity

`tests/test_user_field_parity.py`, mirroring the client one: field-set equality across the nine user shapes (model, YAML `UserEntry`, MCP `_user_to_dict`/create/update schemas + `_USER_COMMON_PROPERTIES`, UI form, REST read surface), with every exclusion documented AND asserted - `username` as YAML key, `password` on read surfaces, derived `authorities`, and the one real gap found while writing it: the UI users form has no `attributes` input, recorded as #291 with an assertion that fails if the input appears (so the exclusion cannot outlive the gap).

## #285 - one-home rule + honest imports

- CONTRIBUTING gains "Domain invariants have one home": routes and MCP are adapters; the same policy implemented twice is a review smell. Echoed from VISION principle 4.
- Five deferred imports claiming circular dependencies that do not exist hoisted to module top (`services/auth_code.py` on the same `get_config_if_loaded` never-construct rule as audit, `services/device_code.py` typing import from `models`, `routes/ui.py` x3 plus stdlib-in-function, `routes/api.py` x2); `services/audit.py`'s comment now names its real constraint instead of mislabeling it circularity.

## Verification

1900 unit tests green on the rebased head (base: current `main`), mypy/ruff/import-linter clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,24 @@ previous leniency allowed - needs a one-time adjustment.
   unchanged; both exemptions are now pinned by tests.
 
 ### Added
+- **Access-point parity contract for token issuance** (#283,
+  `tests/test_token_issuance_parity.py`): the set of modules calling
+  `TokenService.create_token` must equal a declared registry (AST-checked),
+  and every (surface, policy) pair - client binding, scope ceiling, resource
+  ceiling - must declare `enforced` (behaviorally asserted) or `exempt`
+  (the exemption itself pinned, with its documented reason). A new minting
+  surface fails the suite until it takes a stance; this mechanizes the
+  #269/#272 bug class out of existence.
+- **User-field parity test** (#284, `tests/test_user_field_parity.py`):
+  the nine user shapes (model, YAML entry, MCP read/create/update surfaces,
+  UI form, REST read) are held to field-set equality with documented,
+  asserted exclusions - the guard whose absence let #280 drift silently.
+  Found and recorded #291 (the UI users form has no `attributes` input).
+- **"Domain invariants have one home" review rule** (#285) in CONTRIBUTING,
+  echoed from VISION principle 4; five deferred imports claiming
+  nonexistent circular dependencies hoisted to module top, and the one
+  legitimately special case (`services/audit.py`) now documents its real
+  reason (never construct config from a log path).
 - **Horizontal `/authorize` login card composition** (#249). New per-client
   `layout` field, `"vertical"` (default, unchanged) or `"horizontal"`: the
   latter places the client info block and the login form side by side in a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,13 +105,14 @@ previous leniency allowed - needs a one-time adjustment.
 
 ### Added
 - **Access-point parity contract for token issuance** (#283,
-  `tests/test_token_issuance_parity.py`): the set of modules calling
-  `TokenService.create_token` must equal a declared registry (AST-checked),
-  and every (surface, policy) pair - client binding, scope ceiling, resource
-  ceiling - must declare `enforced` (behaviorally asserted) or `exempt`
-  (the exemption itself pinned, with its documented reason). A new minting
-  surface fails the suite until it takes a stance; this mechanizes the
-  #269/#272 bug class out of existence.
+  `tests/test_token_issuance_parity.py`): the set of CALL SITES of
+  `TokenService.create_token` (`file::function`, AST-checked) must equal a
+  declared registry, and every (surface, policy) pair - client binding,
+  scope ceiling, resource ceiling - must declare `enforced` (behaviorally
+  asserted) or `exempt` (the exemption itself pinned, with its documented
+  reason). A new minting call site - even a second function in an
+  already-registered module - fails the suite until it takes a stance;
+  this mechanizes the #269/#272 bug class out of existence.
 - **User-field parity test** (#284, `tests/test_user_field_parity.py`):
   the nine user shapes (model, YAML entry, MCP read/create/update surfaces,
   UI form, REST read) are held to field-set equality with documented,
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -117,6 +117,26 @@ what lets `config.py` import it without a cycle). The contracts live in
 count. If a change needs a new edge between layers, adjust the contract in the
 same PR and say why.
 
+### Domain invariants have one home (#285)
+
+OAuth/OIDC/SAML policy belongs in services and domain logic; routes and the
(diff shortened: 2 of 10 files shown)


cdelmonte-zg:
Round 1 finding fixed in b2b7cae. You were right that the file-level registry was a holed tripwire, and the hole was exactly the class this test exists for: a second minting function inside an already-registered module left the AST result unchanged and the test green. The registry now keys `file::enclosing_function` (dotted for nested defs, `<module>` at module level) - today's three entries are `routes/oauth.py::token`, `routes/api.py::generate_token`, `mcp_server.py::_tool_generate_token` - and POLICY_STANCE follows the same keys. A collector self-test pins the mechanism with your exact scenario: a synthetic `routes/api.py` carrying `generate_token` plus a `sneaky_new_endpoint` yields two registry entries, not one.

On the second observation: agreed, and recorded in place rather than 'fixed' - a comment above POLICY_STANCE now states that the stance strings and the behavioral assertions are linked by convention deliberately, that nine combinations do not justify a mini-framework, and that the trigger to revisit is the matrix outgrowing what a reviewer can eyeball.

Suite 1884, mypy/ruff/import-linter clean. CHANGELOG entry updated to say call sites.

PR #293, opened 2026-08-30
fix: RevocationStore entries expire and get swept (#288)

### LINKED ISSUE #288: fix: RevocationStore grows without bound (revoked JTIs and families never swept)
`RevocationStore` keeps revoked JTIs and rotated-family markers in two module-level sets that grow without bound and are never swept (`services/revocation.py:26-27`) - unlike auth codes (TTL sweep) and device codes (TTL + `MAX_PENDING_DEVICE_CODES`). Every revocation and every refresh rotation on a long-lived instance is a permanent memory increment: a leak by construction, invisible in short-lived test runs.

A revoked JTI only needs remembering until the token it names expires; after `exp` the signature check rejects it anyway. Fix: store expiry alongside the JTI (callers know the token's `exp` at revocation time) and sweep opportunistically like `AuthCodeStore` does. Family markers for refresh rotation can carry the family's last `exp` the same way.

3.1: small, self-contained, and the future runtime store (#192/#235) inherits whatever contract this establishes, so it is worth fixing in-memory first.

cdelmonte-zg:
Closes #288. The audit's "leak by construction": revoked jtis and rotation-family markers were kept in two never-swept sets, so every revocation and every refresh rotation on a long-lived instance was a permanent memory increment.

## The three-state trust contract (final shape, review rounds 1+2)

Entries are `id -> expires_at`, resolved by a contract expressed in the store's own API (not implicitly in the callers):

- **`expires_at` omitted** (`_UNSET` sentinel, the default): the caller has no trusted expiry - an unverified payload (`/logout`'s `id_token_hint`, which is decoded without verification and passes NO claim to the store) or none at all. Bounded default retention: 8 days, covering the 7-day refresh JWT plus clock skew.
- **`expires_at=None` from a verified payload**: the token carries no `exp` claim - which `verify_jwt` accepts - so it never expires, and the entry gets **indefinite retention** (never swept). The pre-#288 behavior for exactly the tokens where forgetting would be wrong.
- **`expires_at=<number>` from a verified payload** (`/revoke` after signature verification; the refresh grant's claim): kept **exactly, however long** - `/api/users/<u>/token` and MCP `generate_token` mint arbitrary lifetimes, so no fixed cap is safe. Normalized to `float` at the boundary (PyJWT can hand back an exp it merely coerced for its own check); a non-coercible trusted exp fails SAFE toward indefinite retention, never toward forgetting a revocation.

Writes are **monotonic** (`max(existing, new)`): re-revoking a jti can only extend its retention - restoring the idempotence the old `set.add()` gave for free. The family marker set on reuse detection follows the presented ancestor: bounded when it carries an `exp` (every nanoidp-minted descendant lives at most 7 more days), indefinite when it does not. `check_and_claim_refresh` keeps its exact pre-#288 contract (pinned): first claim passes, reuse revokes the whole family, last validation in the grant.

Sweep runs opportunistically on the mutating paths, which already serialize under the store lock; `is_revoked` stays lock-free (a stale entry past `exp` can only concern a token every verification path already rejects on `exp`).

Out of scope, agreed in review: requiring `exp` globally in `verify_jwt` (a wider contract change touching every token consumer). Inheritance note for #192/#235: a future persistent store should represent "no expiry" explicitly (NULL) rather than serializing infinity - this PR's in-memory `float("inf")` is pragmatic, not a serialization format.

## Verification

12 tests in `tests/test_revocation_expiry.py` (time monkeypatched, no sleeps), including the review regressions: a verified 14-day exp survives the day-8 sweep and is swept after its own exp; a re-revoke with a shorter expiry never shortens retention; a verified no-exp revocation and a consumed no-exp refresh claim both survive a 10-year jump; a forged 10-year exp in `/logout`'s hint provably never reaches the store; string-exp normalization. Suite 1912 green, mypy/ruff/import-linter clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -288,10 +288,12 @@ previous leniency allowed - needs a one-time adjustment.
   opportunistically on the mutating paths. A VERIFIED token's own `exp` is
   kept exactly - tokens minted via `/api/users/<u>/token` or MCP
   `generate_token` take arbitrary lifetimes, so no fixed cap is safe for
-  them; callers holding only unverified claims (the `/logout`
-  id_token_hint) pass no exp at all and get a bounded 8-day default; and
-  writes are monotonic, so re-revoking a jti can only extend its
-  retention, never shorten it.
+  them; a verified payload WITHOUT an `exp` claim (which `verify_jwt`
+  accepts) gets indefinite retention, since a token that never expires can
+  never have its revocation forgotten; callers holding only unverified
+  claims (the `/logout` id_token_hint) pass nothing and get a bounded
+  8-day default; and writes are monotonic, so re-revoking a jti can only
+  extend its retention, never shorten it.
 - **MCP `update_user` can now update custom `attributes`** (#280): the field
   was accepted by `create_user` and returned by every read surface, but the
   `update_user` schema and handler silently lacked it - an agent could set
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -281,6 +281,14 @@ previous leniency allowed - needs a one-time adjustment.
   (RFC 3986 §3, e.g. `about:`). No audience bypass or escalation.
 
 ### Fixed
+- **`RevocationStore` entries now expire** (#288): revoked jtis and rotation
+  family markers lived in two sets that were never swept - every revocation
+  and every refresh rotation on a long-lived instance was a permanent memory
+  increment. Entries now carry an expiry (the verified token's own `exp`
+  where the caller has it; a bounded 8-day default otherwise - nothing
+  outlives the 7-day refresh JWT) and the store sweeps opportunistically on
+  the mutating paths. An unverified `exp` (the `/logout` id_token_hint) can
+  only narrow retention, never extend it past the bound.
 - **MCP `update_user` can now update custom `attributes`** (#280): the field
   was accepted by `create_user` and returned by every read surface, but the
   `update_user` schema and handler silently lacked it - an agent could set
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -284,11 +284,14 @@ previous leniency allowed - needs a one-time adjustment.
 - **`RevocationStore` entries now expire** (#288): revoked jtis and rotation
   family markers lived in two sets that were never swept - every revocation
   and every refresh rotation on a long-lived instance was a permanent memory
-  increment. Entries now carry an expiry (the verified token's own `exp`
-  where the caller has it; a bounded 8-day default otherwise - nothing
-  outlives the 7-day refresh JWT) and the store sweeps opportunistically on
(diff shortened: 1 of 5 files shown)


cdelmonte-zg:
Round 1 blockers fixed in 6370213, both reproduced red before the fix.

**Blocker 1** - you were right and the premise sentence was mine and wrong: `/api/users/<u>/token` and MCP `generate_token` accept arbitrary `exp_minutes`, so 'nothing outlives the 7-day refresh JWT' does not hold for access tokens, and the reproduction (revoke a 14-day token, advance 9 days, sweep, still valid for 5 more) confirmed the flicker-back. Fixed exactly along the line you suggested: the trust decision moved to the call sites - `/revoke` and the refresh grant pass their VERIFIED payload's `exp`, which the store keeps exactly however long; `/logout` no longer passes any claim from the unverified hint and gets the bounded default. `_effective_expiry` has a single semantics again (verified-exact or default), no cap contaminating the trusted path; the constant is renamed `_DEFAULT_RETENTION_SECONDS` to say what it now is.

**Blocker 2** - writes are monotonic (`max(existing, new)`), restoring the idempotence `set.add()` gave for free; the shorter-re-revoke reproduction is the pin.

Both regression tests you asked for are in, plus a third at the route level: a forged 10-year `exp` in `id_token_hint` provably never reaches the store (the entry gets the default). Store docstrings now state the trust contract explicitly. Suite 1909, statics clean.

cdelmonte-zg:
Round 2 blocker fixed in 098b063, reproduced red first (both legs: a revoked verified no-exp token came back after the day-8 sweep, and a consumed no-exp refresh claim became replayable).

Fixed with the sentinel you pointed at - the pre-#288 file's `_UNSET` is back, as a typed class, and the store contract is now three-state, spelled out on `RevocationStore`:

- **omitted** (`_UNSET`): no trusted expiry available - bounded default (`/logout`'s unverified hint);
- **`None` from a verified payload**: the token never expires, so the entry gets **indefinite retention** (`float('inf')`, never swept) - the pre-#288 behavior for exactly the tokens where forgetting would be wrong;
- **a number from a verified payload**: kept exactly.

The call sites did not need to change: `/revoke` and the refresh grant already pass `payload.get("exp")`, which now lands in the right state automatically; `/logout` already passes nothing. The family marker follows the presented ancestor (bounded with exp, indefinite without). Your minor note is in too: a trusted exp is normalized to `float` at the store boundary, and a non-coercible one fails SAFE toward indefinite retention, never toward forgetting.

Agreed on keeping `require: ["exp"]` out of `verify_jwt` in this PR - that is a wider contract change; worth its own issue if you want it for 3.0 (it would tighten every consumer of every token at once).

Tests: the indefinite-revocation and indefinite-refresh-consumption regressions you asked for, plus the boundary-normalization pin. Suite 1912, statics clean.

PR #294, opened 2026-08-30
fix: UI edit no longer corrupts non-string user attributes (#291)

### LINKED ISSUE #291: fix: UI users form corrupts non-string attributes on edit; parity test blind to the attr_key[]/attr_value[] widget

cdelmonte-zg:
Closes #291 - with a premise correction that reshaped the issue (see the retitle and comment there): the users form has had a dynamic `attr_key[]`/`attr_value[]` attributes widget all along; the audit's "no input" claim came from a single-name regex that cannot see brackets, asserted without reading the template.

The **real defect**, reproduced empirically before fixing: the edit form rendered `value="{{ value }}"`, so a list- or mapping-valued attribute (settable via YAML and MCP) rendered as its Python repr - an **untouched** edit round-trip silently replaced `{"teams": ["alpha", "beta"]}` with the string `"['alpha', 'beta']"`.

## The fix (final shape, review round 1)

Each attribute row carries an explicit hidden **`attr_encoding[]`**, because a text box cannot distinguish the STRING `'["a"]'` from the LIST `["a"]` once both are rendered as text:

- **`string`** (stamped by the template for string values): kept verbatim on submit, even when the value looks like JSON - the round-1 blocker was exactly this string being re-parsed into a real list by the `[`/`{` heuristic;
- **`json`** (stamped for container values, rendered via `|tojson|forceescape` - `tojson` marks its output safe and does NOT escape double quotes, which would break the double-quoted attribute): parsed back on submit; edited into non-JSON, it degrades to the literal string, matching the form's permissive style;
- **`auto`** (rows typed fresh in the browser, and the per-row fallback for scripted POSTs without the hidden field): the `[`/`{` convenience heuristic - a typed JSON container stores as a container, plain scalars stay strings.

The attribute-row parser, previously duplicated verbatim in the create and edit routes, is one shared `_parse_attribute_rows()`. `test_user_field_parity.py` recognizes the widget explicitly (asserting `attr_key[]`/`attr_value[]` exist) instead of excluding the field on a false premise, and the misleading CHANGELOG line from #292 is corrected.

## Verification

6 round-trip tests in `tests/test_ui_user_attributes.py`, browser-realistic (HTML entities decoded, encodings carried like a browser would): list and mapping surviving an untouched edit; **the JSON-looking string surviving next to the real list** (the round-1 regression, reproduced red first); typed JSON container parsed; malformed JSON kept literal; plain scalar verbatim. Verified live both rounds: seeded via explicit encodings, untouched round-trip preserves both types. Suite 1892 on the rebased head (current `main`), mypy/ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,7 +129,10 @@ previous leniency allowed - needs a one-time adjustment.
   the nine user shapes (model, YAML entry, MCP read/create/update surfaces,
   UI form, REST read) are held to field-set equality with documented,
   asserted exclusions - the guard whose absence let #280 drift silently.
-  Found and recorded #291 (the UI users form has no `attributes` input).
+  Found #291 - originally misfiled as "no `attributes` input in the UI
+  form" (the dynamic `attr_key[]`/`attr_value[]` widget was there all
+  along, invisible to the single-name regex); the real defect behind it is
+  fixed below.
 - **"Domain invariants have one home" review rule** (#285) in CONTRIBUTING,
   echoed from VISION principle 4; five deferred imports claiming
   nonexistent circular dependencies hoisted to module top, and the one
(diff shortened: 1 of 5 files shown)


cdelmonte-zg:
Round 1 blocker fixed in db153bc, reproduced red first with your exact scenario: the string `'["alpha","beta"]'` seeded next to the real list `["alpha","beta"]`, both under an untouched edit - the string came back a list.

Fixed with the explicit row metadata you suggested: a hidden `attr_encoding[]` per row, stamped by the template - `string` (kept verbatim even when it looks like JSON), `json` (container values rendered as JSON, parsed back; degrades to the literal string if edited into non-JSON), and `auto` for rows typed fresh in the browser (the `[`/`{` convenience heuristic, unchanged). A scripted POST without the hidden field falls back to `auto` per row, so pre-existing non-browser callers keep their behavior.

The mandatory regression test is in (both types side by side surviving an untouched edit), the resubmit-helper now carries encodings the way a browser would, and the whole thing is verified live: seed with explicit encodings, untouched round-trip, both types preserved. Suite 1906, statics clean, CHANGELOG updated to describe the encoding contract.

Issue #193, opened 2026-08-23
MCP Streamable HTTP transport for nanoidp's own MCP server


cdelmonte-zg:
**Context (updated 2026-09-19).** nanoidp's direction is written down in [VISION](https://github.com/cdelmonte-zg/nanoidp/blob/main/VISION.md#what-comes-next) (the text lands with #396): it is a test identity provider, a real IdP built for testing, and the next work is the runtime state chain **#363 -> #354 -> #193**, in that order. This issue is step 3 of 3: nanoidp's own MCP server over Streamable HTTP, a separate process and the first consumer of the shared store. It is also where MCP/HTTP parity (VISION principle 4) starts to hold for runtime state and not only for the declared configuration: today an MCP stdio edit is invisible to a running IdP until it is saved and reloaded. It is scheduled, no longer deferred. The design below is the starting point: like every issue here it gets a census against the current code and a rewrite before implementation.

**Why.** nanoidp's MCP server is stdio only (`stdio_server` in `mcp_server.py`); agent hosts such as n8n connect to a remote HTTP MCP endpoint. Without it an n8n AI Agent cannot call `create_user`, `generate_token` or `rotate_keys` directly. SSE is legacy in MCP 2026-07-28 and is not implemented.

**Scope.**
- `nanoidp-mcp --http HOST:PORT` (a separate process: the SDK is asyncio/Starlette and the app is sync Flask; no in-process mounting in v1), Streamable HTTP only, stateless sessions by default.
- Authentication: nanoidp's own access tokens, resource `urn:nanoidp:mcp` (RFC 8707, #187), scopes `nanoidp:read` and `nanoidp:admin` (#186): read tools need `nanoidp:read`, mutating tools `nanoidp:admin`; `--readonly` stays a server-side policy on top. No `management_secret` bearer: by this point in the milestone the proper model exists and a temporary scheme would only be replaced.
- The endpoint is itself a protected MCP server and behaves like one: `/.well-known/oauth-protected-resource` (RFC 9728) naming this nanoidp as its authorization server, `401` with `WWW-Authenticate: Bearer resource_metadata=...` on missing/invalid tokens, `403 insufficient_scope` with `scope=`, so a host such as n8n can run the normal discovery flow (MCP 2026-07-28). The mock server (#191) and this endpoint share the validation code.
[...]

PR #295, opened 2026-08-30
refactor: one error shape per surface class (#287)

### LINKED ISSUE #287: refactor: consolidate error shaping (six mechanisms today) before a second HTTP transport
The pre-3.0 audit counted six distinct error-shaping mechanisms across the surfaces: OAuth JSON (`jsonify error/error_description`, 26 sites in `routes/oauth.py`), OAuth redirect errors (`_authorize_error_redirect`), UI flash+redirect (49 sites, `routes/ui.py`), Werkzeug `abort()` HTML (`routes/saml.py:453` etc.), bare `return "text", 400` (`saml.py:862/:868/:873`), and two different MCP shapes in one surface (`_reject`'s `{error, code, tool}` vs handlers' `{success: False, error, kind}` - `_reject`'s own docstring admits the branches "had drifted"). The `kind` field threads through a third channel (`ConflictError.kind`, `HookError.kind`) surfaced by MCP and `/api/config/reload` but swallowed into flash strings by the UI.

Related: `exceptions.py` defines 20 exception classes of which 18 are referenced nowhere (only `SAMLSignatureError` and `InvalidTokenError` are used); routes throw `ValueError`/bare `Exception` instead. Either the hierarchy becomes the real error channel during this consolidation or it goes.

A second HTTP transport (#193) adds a seventh channel if this is not consolidated first. Scope: pick one error model per surface class (protocol JSON, browser, MCP), route the `kind` taxonomy consistently, fold or delete the dead hierarchy. Behavior-preserving on the wire except where a `text/400` becomes proper JSON (CHANGELOG note). No milestone: sequenced with the #193 prerequisite refactor, after it or with it.

cdelmonte-zg:
Closes #287. Net -351 lines. Three moves, one behavior change:

- **Dead exception hierarchy removed.** `exceptions.py` declared 20 classes; a usage census (part of the audit) found exactly one ever raised or caught anywhere - `SAMLSignatureError` (the other apparent hits were pyjwt's own `InvalidTokenError` and a same-named test class). It stays, as a plain `Exception` subclass; the other 19 and their only consumer (their own 286-line test file) are gone. A taxonomy nothing raises is documentation that lies.
- **`/saml/attribute-query` transport errors are SOAP 1.1 Faults** (behavior change, CHANGELOGed): a malformed query or internal failure now answers `soap:Fault` on HTTP 500 with `faultcode` Client/Server per SOAP 1.1 §6.2, replacing `return "text", 400` - a shape no SOAP stack parses. Protocol-level conditions (UnknownPrincipal, #275) stay inside the SAML Response, signed as of #289. Also: the inline `import traceback` in the handler is now `logger.exception`.
- **The MCP error model is documented as the two-layer contract it already is** - dispatch refusals (`is_error` + `MCP_*` code taxonomy) vs domain results (`success: false` + `kind`), stated on `_reject` where the shape is defined. Deliberately NOT collapsed: "user not found" is an answer, not a failed call, and one wire shape would break every MCP consumer for cosmetic gain.

CONTRIBUTING gains an **"Error surfaces"** section naming the shape for each surface class (OAuth JSON/redirect per RFC 6749/9207, UI flash, SAML Status vs SOAP Fault, MCP layers) - the review criterion that stops a seventh mechanism from appearing, which is the real point before a second HTTP transport (#193, #286).

Suite 1873 green (31 dead-hierarchy tests replaced by 4 real ones), mypy/ruff/import-linter clean; verified live: a malformed query answers `soap:Fault`/`soap:Client` on HTTP 500 with the message in `faultstring`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,18 @@ previous leniency allowed - needs a one-time adjustment.
   it previously claimed "after JWT authentication", which the code never did.
 
 ### Changed
+- **`/saml/attribute-query` transport errors are SOAP 1.1 Faults** (#287):
+  a malformed query (missing AttributeQuery/Subject/NameID) or an internal
+  failure now answers with a proper `soap:Fault` (HTTP 500, `faultcode`
+  Client/Server per SOAP 1.1 §6.2) instead of bare plain text with a 400 -
+  a shape no SOAP stack could parse. Protocol-level conditions (unknown
+  principal) keep answering inside the SAML Response as before.
+- **The dead exception hierarchy is gone** (#287): `exceptions.py` declared
+  20 classes of which exactly one was ever raised; only `SAMLSignatureError`
+  remains (now a plain `Exception` subclass). Errors are shaped per surface
+  - the model is written down in CONTRIBUTING ("Error surfaces"), including
+  the deliberate two-layer MCP contract (dispatch refusals vs domain
+  results), now documented where the shapes are defined.
 - `TokenService.create_token` now owns the #73 mint-side rule (#278):
   `issue_refresh_token` defaults to "only when the token is bound"
   (`client_id` given), and asking for a refresh token without a binding
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -153,11 +153,14 @@ class's shape, never invents a seventh:
   failures on the SOAP endpoint answer with a SOAP 1.1 Fault (HTTP 500,
   `faultcode` Client/Server per SOAP 1.1 §6.2). Browser-facing SSO steps may
   use Werkzeug `abort()` HTML, since the reader is a person mid-redirect.
-- **MCP**: two deliberate layers - dispatch refusals
-  (`{"error", "code", "tool"}`, `is_error=True`, the `MCP_*` code taxonomy)
-  vs domain results (`{"success": False, "error", ...}` plus `kind` for
-  typed conditions), because "not found" is an answer, not a failed call.
-  The layer contract is documented on `_reject` in `mcp_server.py`.
+- **MCP**: two deliberate SHAPES - dispatch refusals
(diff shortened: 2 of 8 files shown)


cdelmonte-zg:
Round 1 findings all fixed in 6a993b0, each verified:

1. **The e2e was worse than out-of-date - it had never tested anything.** The 'SAML Attribute Query' agent test sent a BARE, un-enveloped AttributeQuery; ElementTree's `.//` never matches the root element, so every run since the test existed landed in the `status in [200, 400, 501]` fallback - the success path, and the #275 unknown-user leg later added inside it, never executed. Your 400->500 change did not break a working test; it unmasked a dead one. Rewritten with the SOAP envelope the binding requires and three real legs (valid -> attributes; unknown NameID -> Requester/UnknownPrincipal without assertion; malformed -> SOAP Fault 500 soap:Client). Verified live: 11 consecutive full-agent runs green against a fresh server. (The other two attribute-query e2e tests already used proper envelopes.)
2. **MCP docs corrected to reality, not the other way around** - exactly as you prescribed: `success: false` domain results DO come back `is_error=true` (`call_tool`: `success is False or 'error' in result`); the 'an answer, not an error' cases are the negative queries (`found: false` / `valid: false`). `_reject`'s docstring and CONTRIBUTING now cite the module-docstring isError table my #287 text had contradicted, and the code-taxonomy names are the ones that actually exist (`MCP_READONLY_MODE`, `MCP_INTERNAL_ERROR`, ...).
3. **Broken XML is a Client fault**: the parse is caught specifically and answers `soap:Client`; only genuine nanoidp-side failures reach the catch-all's `soap:Server`. Unit pin added (`soap:Server` asserted absent).

Suite 1874, statics clean. #296 has been rebased onto this head (the docstring correction ported into the package's `__init__`), its stack branch updated, and its body wording softened per your non-blocking note: the compatibility surface is the explicitly re-exported names, not every internal symbol of the old monolith.

Issue #306, opened 2026-08-30
decision: require the exp claim in verify_jwt (major-window call)


cdelmonte-zg:
DECISION for the 3.0 window, raised in the #293 round-2 review: `CryptoService.verify_jwt` does not pass `require: ["exp"]`, so a correctly signed JWT WITHOUT an exp claim verifies - everywhere: /userinfo, /introspect, /revoke, the refresh grant, MCP verify_token. Nothing nanoidp mints omits exp, so this only concerns hand-crafted tokens signed with the IdP key; the RevocationStore already handles them correctly (indefinite retention, #293). But an eternal bearer token accepted by every consumer is a surprising default for a tool whose point is testing realistic behavior.

Requiring exp is a BREAKING change for anyone minting such tokens deliberately (a test fixture with no expiry, say) - which is why it sits in this major's window: after 3.0.0 final it costs another major or stays debt.

Options: (a) `require: ["exp"]` globally in verify_jwt, CHANGELOG breaking note, one escape hatch nowhere (strictest, spec-realistic); (b) require it only where tokens are SPENT (refresh grant) and keep verification-only paths lenient; (c) decide it stays lenient and document that on verify_jwt, closing this issue as by-design. Filed on the 3.0.0 milestone so the decision happens before the final; moving it out IS deciding (c) implicitly - better to write it down either way.

PR #307, opened 2026-08-30
feat: the nanoidp token profile requires exp (#306, BREAKING)

cdelmonte-zg:
Implements decision (a) from #306, in exactly the frame agreed there: **this is nanoidp's token-profile policy, not a JWT-spec rule**. RFC 7519 leaves `exp` optional; OIDC Core requires it on ID Tokens and RFC 9068 on JWT access tokens; `create_jwt` has always stamped one on everything nanoidp mints. The single security property: a token accepted by nanoidp has a finite lifetime - an eternal bearer would let an integration test pass here and fail against any real IdP.

- `verify_jwt` adds `require: ["exp"]` and its docstring now states what it verifies: a JWT acceptable **as a nanoidp token**, not any valid RS256 JWT. No other claim is newly required (`iss`/`iat`/`nbf`/`jti` stay a separate token-profile discussion, per the decision); no escape hatch. Only hand-crafted tokens signed with the nanoidp key are affected - *migration:* add an `exp` to such fixtures.
- The #293 'verified no-exp -> indefinite retention' branch in RevocationStore **stays**, deliberately: the store does not assume every future caller applied this validation policy (the fail-safe property named in the round-3 review).
- **Adjacent fix the new refresh test surfaced**: an unverifiable refresh token answered with a Werkzeug 401 HTML abort. Now RFC 6749 §5.2 JSON (`invalid_grant`, 400); CHANGELOG breaking note, old pin updated. NOT the only such branch (review round 1 corrected my 'the one' claim): the /token shell and the grant handlers still hold ~20 abort() HTML branches contradicting the #287 error-surfaces contract - tracked as #308 rather than widened into this PR.

Tests, per the decision's list: signed no-exp -> `verify_jwt` rejects; signed future-exp -> accepted; no-exp at representative endpoints through their own error contracts (`/introspect` -> `active: false`, `/userinfo` -> 401); and the refresh case that surfaced the issue -> `invalid_grant`. Suite 1898 green, mypy/ruff clean; verified live with a discriminating pair (same claims, only exp differs: `active` False/True at `/introspect`), full e2e 61/61.

(Verification note: an initial live probe showed false negatives that turned out to be a leftover server from another worktree answering on the probe port - re-run against a verified-fresh instance before concluding anything. Process lesson recorded.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,10 +142,11 @@ previous leniency allowed - needs a one-time adjustment.
   affected. *Migration:* add an `exp` to such fixtures. No other claim is
   newly required.
 - **An unverifiable refresh token now answers RFC 6749 §5.2 JSON**
-  (`invalid_grant`, HTTP 400) instead of a Werkzeug 401 HTML page - the
-  one protocol-endpoint error that still violated the "Error surfaces"
-  rule (#287). *Migration:* read `error` from the JSON body; the status
-  moves from 401 to 400.
+  (`invalid_grant`, HTTP 400) instead of a Werkzeug 401 HTML page, per
+  the "Error surfaces" rule (#287). Other /token error branches still
+  answer Werkzeug HTML; converting them is tracked separately.
+  *Migration:* read `error` from the JSON body; the status moves from 401
+  to 400.
 
 ### Added
 - **Access-point parity contract for token issuance** (#283,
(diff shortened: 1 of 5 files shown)


cdelmonte-zg:
The CodeQL finding on this PR is fixed in 6904581. Real and worth taking even though the exposure predates the PR: the old `abort(401, description=f"...{str(e)}")` leaked the same library exception text, and moving it into the JSON body just made it visible to the scanner. Same rule as the #200 review (never raw exception text outward): the `error_description` is fixed text now - "bad signature, expired, or not acceptable under the nanoidp token profile" - and the library detail stays where it already was, in the audit event. No test pinned the old message; suite 1898 green. The alert should auto-resolve on the next CodeQL scan of this branch.

Issue #309, opened 2026-08-30
flaky e2e: SAML Attribute Query intermittently fails; orphan bare-query 500 in the same runs


cdelmonte-zg:
Intermittent failure of the e2e "SAML Attribute Query" test (the #295 rewrite with SOAP envelope + three legs), observed twice in ~15 local full-agent runs; not yet seen in CI. Filing with full forensics so the next occurrence lands on prepared ground.

**Observed** (local, verified-fresh server, sanity-checked with a server-minted token): run reports `[FAIL] SAML Attribute Query`, `TOTAL 60/61`. The server log of a failing run shows two things that do not compose into an explanation:

1. An orphan `POST /saml/attribute-query` -> **500 "Invalid AttributeQuery: AttributeQuery element not found"** right after the SSO login-flow test's requests, ~17s BEFORE the attribute-query test's own cluster - with no unknown-user or missing-Subject legs following it, and no e2e test that still sends a bare (un-enveloped) query (all senders were audited; the last bare one was fixed in #295).
2. The attribute-query test's own three requests, later in the same run, behaving EXACTLY to contract server-side: valid -> 200 with 6 attributes; unknown NameID -> 200 with the failed-audit signature; malformed -> 500 "Subject not found". If those were the requests scored, every status check passes - so if the failure came from them, it was a BODY assertion (UnknownPrincipal present, Assertion absent, soap:Client present), which the server log cannot show.

**Ruled out empirically**:
- Stale-server interference (one earlier red WAS a ghost server from another worktree - process-hygiene issue, not this one; the runs counted here had a sanity probe against the live server first).
- lxml parser thread-unsafety (`routes/saml.py` shares one module-level `etree.XMLParser` across request threads, the same shape as the #229 ruamel bug): probed with 8 threads x 2000 parses on the shared parser - zero failures. The shared parser remains a theoretical hazard worth removing independently (fresh parser per call is cheap, the serialization.py precedent), but it is NOT demonstrated to be this flake.
- Simple reproduction: 12 consecutive fresh-boot first runs green (10 on cycling ports + 2 ad hoc).

**Done in #308's PR to make the next occurrence self-explaining**: on failure the e2e result message now includes the status AND a 160-char body prefix for each of the three responses.

**Next steps when it fires again**: the failure message now says which leg and what body; correlate with the server log's orphan-500 timing. If the orphan 500 recurs, log the raw request body server-side for that error path (temporarily) to identify the sender.

[...]

PR #310, opened 2026-08-30
fix: every /token error branch answers RFC 6749 §5.2 JSON (#308, BREAKING)

### LINKED ISSUE #308: fix: /token grant error branches still return Werkzeug HTML, contradicting the #287 error-surfaces contract
From the #307 review round 1: #307 converted ONE /token error branch (unverifiable refresh token) to RFC 6749 §5.2 JSON and its prose claimed it was "the one" remaining violation - it was not. The `/token` shell and every grant handler still answer roughly twenty error conditions with Werkzeug `abort()` HTML, in direct contradiction of the #287 "Error surfaces" contract ("OAuth/OIDC protocol endpoints: RFC 6749 §5.2 JSON. Never HTML, never bare text"). As it stands, #287 documents an invariant the code does not yet fully honor.

The census (current line numbers): `routes/oauth.py` :822 (client auth required), :833 (client_id mismatch), :864/:874 (`exp` bounds), :889/:899 (`extra` JSON), :912 (unsupported grant_type); `routes/oauth_grants.py` :155 (missing refresh_token), :201 (access token used as refresh), :244 (client mismatch), :249 (missing subject), :261 (user not found), :400 (revoked), :442+, :462 (password grant missing credentials), :474 (invalid credentials), :536/:546 (missing code/redirect_uri), :568 (invalid code), and the remaining sites a fresh `grep -n "abort(" ` finds on these two files (`oauth.py:637` is the `/client-logos` 404, out of scope - a static asset, not a protocol error).

Fix shape: one conversion pass mapping each branch to its RFC 6749 §5.2 error code (`invalid_request` for missing/malformed parameters, `invalid_client` + 401 for authentication failures, `invalid_grant` for bad codes/tokens/credentials-as-grant, `unsupported_grant_type`), with FIXED descriptions (no exception text - the CodeQL rule #307 just hit) and the existing audit events untouched. Behavior change (status codes may move, bodies become JSON): CHANGELOG under Breaking Changes with a migration line, like #307's.

Tests: the existing suite mostly checks `status_code` only (`test_refresh_token_with_access_token_fails` asserts `== 400` and never sees the HTML body - the review's example), so each converted branch needs its body asserted; consider a small parametric test over (request, expected error code) instead of twenty hand-written ones.

Sequencing: 3.0.0 material in spirit (it completes the #287 contract inside the same major), but it is a wide diff late in the rc cycle - milestone decision to the maintainer.

cdelmonte-zg:
Closes #308 (in 3.0.0 per maintainer decision). Completes the #287 'Error surfaces' contract on the token endpoint: ~20 conditions across the `/token` shell and all five grant handlers answered Werkzeug `abort()` HTML.

## Mapping (one shared `_oauth_error` helper, fixed-text descriptions, audit events untouched)

- **`invalid_request` 400**: missing `refresh_token`/`code`/`redirect_uri`/password-grant credentials; `exp` bounds; malformed `extra`.
- **`invalid_grant` 400**: invalid/expired code; access token spent as refresh; foreign or revoked refresh token; missing subject; unknown user; bad resource-owner credentials. **Status change**: these used to 401 - §5.2 reserves 401 for CLIENT authentication, not grant validity. CHANGELOGed with a migration line.
- **`unsupported_grant_type` 400**: unknown grant type (FIXED description - the caller's value goes to the audit event, never reflected into error_description, per §5.2's ASCII constraint), and the oauth21-disabled password grant (previously mislabeled as a plain 400).
- **`invalid_client`**: 401 with `WWW-Authenticate: Basic` when the client ATTEMPTED header authentication (the §5.2 MUST), 400 otherwise - §5.2's default, and RFC 9110 §11.6.1 forbids a challenge-less 401; a `client_secret_post` client with a wrong body secret gets 400 with no Basic challenge (Basic is not its method).

## Tests

`tests/test_token_error_contract.py`: a parametric table of (request -> status, error code) with **body assertions** on every case - the guard the old suite lacked (the #308 review's point: it only checked status codes and never saw the HTML). Plus foreign-client and revoked-refresh scenario tests. 7 existing pins updated to the new contract; suite 1915 green, mypy/ruff/import-linter clean.

## Verification note (and a new tracked flake)

Full e2e ran 22+ times on fresh boots during this PR: all green except ONE occurrence of an intermittent 'SAML Attribute Query' failure that predates nothing in this diff - investigated in depth (ghost-server interference ruled out via sanity probes; the shared lxml parser hypothesis probed negative with 16k concurrent parses; 12+12 fresh-boot reproduction attempts all green) and tracked as **#309** with the forensics. This PR adds the diagnosability piece: on failure the e2e message now carries status + body prefix per leg, so the next occurrence identifies itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,9 +151,11 @@ previous leniency allowed - needs a one-time adjustment.
   revoked or foreign code/refresh-token - and invalid resource-owner
   credentials on the password grant - are `invalid_grant` (400); an
   unknown or profile-disabled grant type is `unsupported_grant_type`
-  (400); client-authentication failures stay `invalid_client` (401) and,
-  when the client attempted HTTP Basic, now carry the `WWW-Authenticate:
-  Basic` challenge §5.2 requires. Descriptions are fixed text (no library
+  (400); client-authentication failures are `invalid_client` - 401 with
+  the `WWW-Authenticate: Basic` challenge when the client attempted HTTP
+  Basic (the §5.2 MUST), 400 otherwise (§5.2's default; RFC 9110 §11.6.1
+  forbids a challenge-less 401, and a Basic challenge would be wrong for
+  a `client_secret_post` client anyway). Descriptions are fixed text (no library
   detail and no reflected caller input - the unsupported grant type's raw
   value lives in the audit event, not the response). *Migration:* branches that used to answer 401 for GRANT problems
   (revoked/foreign refresh token, unknown user, wrong password) now
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,9 +151,11 @@ previous leniency allowed - needs a one-time adjustment.
   revoked or foreign code/refresh-token - and invalid resource-owner
   credentials on the password grant - are `invalid_grant` (400); an
   unknown or profile-disabled grant type is `unsupported_grant_type`
-  (400); client-authentication failures stay `invalid_client` (401).
-  Descriptions are fixed text (no library detail; the audit events keep
-  it). *Migration:* branches that used to answer 401 for GRANT problems
+  (400); client-authentication failures stay `invalid_client` (401) and,
+  when the client attempted HTTP Basic, now carry the `WWW-Authenticate:
+  Basic` challenge §5.2 requires. Descriptions are fixed text (no library
+  detail and no reflected caller input - the unsupported grant type's raw
+  value lives in the audit event, not the response). *Migration:* branches that used to answer 401 for GRANT problems
   (revoked/foreign refresh token, unknown user, wrong password) now
   answer 400 with `error: invalid_grant` - §5.2 reserves 401 for client
   authentication; read `error` from the JSON body instead of matching
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,10 +143,21 @@ previous leniency allowed - needs a one-time adjustment.
   newly required.
 - **An unverifiable refresh token now answers RFC 6749 §5.2 JSON**
   (`invalid_grant`, HTTP 400) instead of a Werkzeug 401 HTML page, per
-  the "Error surfaces" rule (#287). Other /token error branches still
-  answer Werkzeug HTML; converting them is tracked separately.
-  *Migration:* read `error` from the JSON body; the status moves from 401
-  to 400.
+  the "Error surfaces" rule (#287).
+- **Every `/token` error branch now answers RFC 6749 §5.2 JSON** (#308):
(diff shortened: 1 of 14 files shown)


cdelmonte-zg:
Round 1 findings all fixed in 7735f24, each verified live:

1. **Reflected grant_type**: fixed text now (`The requested grant_type is not supported`); the raw value stays in the audit event, which already recorded it. Your distinguishing test is in - `grant_type='späce\"💥\ntype'` gets the fixed description, asserted both in the parametric table and standalone. You were right that the inline comment justifying reflection contradicted the PR's own premise.
2. **WWW-Authenticate on attempted-Basic 401**: `_token_auth_failed` and the two shell `invalid_client` sites (auth-required, mismatch) issue `Basic realm="nanoidp"` when the Authorization header was presented - and nothing when it was not, since the MUST is conditional on the attempt. Both directions pinned and probed live (header present with bad Basic, absent with no header).
3. **Test coverage**: the four missing branches are in the suite (client_id mismatch WITH the challenge assertion; refresh token without `sub`; oauth21-disabled password grant asserting `unsupported_grant_type` JSON; authorization code for a deleted user), and the two weak pre-existing pins you called out are strengthened - the `/token` mismatch test no longer stops at `== 401`, and the oauth21 test asserts the JSON error code instead of grepping bytes.

The non-blocker is taken as well, since blocker 2 forced touching all the shell sites anyway: the §5.2 shape now has one home, `routes/_oauth_error.py` (`oauth_error` + `invalid_client_error`), imported by both files - the shell no longer hand-builds the JSON four times.

Suite 1923 (+8), mypy/ruff/import-linter clean.

cdelmonte-zg:
Round 2 blocker fixed in 1f2bf5f, exactly along your snippet - and it is the better design on all three axes you named:

- `invalid_client_error` now answers **400 (no challenge) without an Authorization attempt, 401 + `Basic realm="nanoidp"` with one** - satisfying RFC 6749 §5.2's default AND RFC 9110 §11.6.1's no-challenge-less-401 rule at once.
- The `client_secret_post` case is pinned as you asked: wrong body secret -> **400 `invalid_client`, no WWW-Authenticate** (a Basic challenge would name a method that is not the client's).
- `test_invalid_client_without_auth_header_*` moved to 400, the parametric no-auth case follows, and seven pre-existing 401 pins for no-Basic failures moved to 400 with the citation in place.

Also fixed the stale PR-body sentence ('echoing the caller's own parameter') along with the invalid_client bullet. Verified live (no header -> 400 JSON, no challenge; wrong Basic -> 401 + challenge); suite 1924, e2e 61/61, statics clean.

PR #313, opened 2026-08-30
fix: detect an attempted Basic from the raw Authorization header (#311)

### LINKED ISSUE #311: fix: detect an attempted-Basic auth from the raw Authorization header, not werkzeug's parsed object
From the #310 round-3 review (non-blocking edge, recorded per the reviewer's note): `basic_attempted` is derived from `request.authorization is not None`, but werkzeug returns `None` for a syntactically broken `Authorization` header (its parsing history confirms invalid Basic headers can be discarded during parsing). A malformed `Authorization: Basic ...` therefore lands in the no-attempt branch - 400 with no challenge - when RFC 6749 §5.2's MUST arguably covers it as an attempt.

Fix shape: establish the ATTEMPT from the raw header (`request.headers.get("Authorization", "").lower().startswith("basic")` or equivalent), and keep the parsed object only as the source of username/password. One helper touch in `routes/_oauth_error.py` callers (or a tiny `_basic_attempted()` next to `_request_client_identity`), plus a pin with a garbage Basic header asserting 401 + challenge.

Low severity: only misformats the failure response for callers already sending broken headers. 3.1.

cdelmonte-zg:
Closes #311 (pulled into 3.0.0 so the §5.2 contract born in #310 ships complete, not amended in a minor). The #310 round-3 review edge: werkzeug parses a syntactically broken `Authorization: Basic ...` to `None`, so a botched attempt landed in the no-attempt branch - 400 with no challenge - where the §5.2 MUST covers it as an attempt.

`_basic_attempted()` now reads the RAW header for the ATTEMPT decision only - the scheme token is split off and compared with `casefold()` EQUALITY (review round 1: `startswith` would have classified `BasicFoo` as Basic); `request.authorization` stays the sole source of username/password. Both /token shell call sites converted (the mismatch site already passes `True` by construction).

Pins in `test_token_error_contract.py`, both directions: garbage Basic -> 401 + `WWW-Authenticate: Basic`; a `Bearer` header is NOT a Basic attempt -> 400, no challenge. Suite 1926 green, statics clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -155,7 +155,10 @@ previous leniency allowed - needs a one-time adjustment.
   the `WWW-Authenticate: Basic` challenge when the client attempted HTTP
   Basic (the §5.2 MUST), 400 otherwise (§5.2's default; RFC 9110 §11.6.1
   forbids a challenge-less 401, and a Basic challenge would be wrong for
-  a `client_secret_post` client anyway). Descriptions are fixed text (no library
+  a `client_secret_post` client anyway). The attempt is detected from the
+  raw `Authorization` header (#311): a syntactically broken Basic header -
+  which werkzeug parses to nothing - still counts as an attempted Basic
+  and gets the 401 + challenge. Descriptions are fixed text (no library
   detail and no reflected caller input - the unsupported grant type's raw
   value lives in the audit event, not the response). *Migration:* branches that used to answer 401 for GRANT problems
   (revoked/foreign refresh token, unknown user, wrong password) now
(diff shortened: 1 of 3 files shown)


cdelmonte-zg:
Micro-blocker fixed in 17543b7 with your exact shape: the scheme token is split off and compared with `casefold()` EQUALITY - `BasicFoo whatever` is a different scheme and gets 400 with no challenge (pinned), while a bare `Authorization: Basic` with no credentials still counts as an attempt since the scheme was named (also pinned, per your note that the split-based code already handles it). The two earlier pins stay. Suite 1928, statics clean.

PR #314, opened 2026-08-30
fix: rate limiting on /token is enforced for real (#304)

### LINKED ISSUE #304: fix: rate limiting is advertised but never enforced (wire it or remove it)
Found by the audit and then found to be WORSE on verification: rate limiting is advertised but never enforced.

- `create_app` builds a `Limiter` with `default_limits=[]` and there is NO `@limiter.limit(...)` decorator or `.limit()` call anywhere in the codebase - so with `rate_limit_enabled: true` the app logs "Rate limiting: enabled (10/minute on /token)" while enforcing nothing.
- `settings.rate_limit_token_endpoint` is a documented, validated setting with zero effect.
- `get_limiter()` has zero callers.

This is a "metadata never lies" violation (VISION principle 2, the #41 precedent: an advertised-but-unimplemented feature was removed rather than kept). Options, in order of preference: (a) actually wire `rate_limit_token_endpoint` onto `/token` (the log line then becomes true; a test hammers /token past the limit and asserts 429), or (b) remove the flag, the setting, the limiter plumbing and the flask-limiter dependency. Either is small; shipping the current state further is the only wrong option. 3.1.

cdelmonte-zg:
Closes #304 (pulled into 3.0.0: the stricter-dev behavior change belongs in the major). Direction chosen: **wire it** - removal would drop a capability a testing IdP legitimately wants, and wiring makes the log line true.

Worse than the audit said, on inspection: besides the limiter never being applied (`default_limits=[]`, zero decorated views), the two Settings fields were in **no document section** - YAML could not set them at all. The only path that ever flipped the flag was the **stricter-dev profile, which has always forced `rate_limit_enabled: true`** - so the profile's hardening promise was a no-op for its entire life.

- The configured `rate_limit_token_endpoint` now decorates `oauth.token` (post blueprint registration); 429 answers **JSON** with `Retry-After`/`RateLimit-*` headers - the only limited view is a protocol endpoint, so no framework HTML.
- `server.rate_limit_enabled` / `server.rate_limit_token_endpoint` are configurable from YAML through the full settings plumbing - and the table-driven legs (#175/#283 machinery) picked the new OWNED_SETTINGS rows up with zero extra code: the parity suite passed on first run after adding the two rows. Schema regenerated, `/api/config` + MCP `get_settings` expose both, configuration.md documents them.
- **Behavior change (CHANGELOGed under Fixed with a bold note): stricter-dev instances now really throttle `/token`** at the configured rate (default 10/minute) - the hardening the profile always claimed.

Tests (5): the 4th hit in a 3/minute window answers 429 JSON + `Retry-After`; other endpoints unlimited; scope is /token only; the default stays unlimited (12 rapid hits all 200); the stricter-dev promise pinned via a profile boot. Verified live from a YAML-only config: `200 200 200 429`. Suite 1929 green, mypy/ruff/import-linter clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -352,8 +352,12 @@ previous leniency allowed - needs a one-time adjustment.
   `rate_limit_enabled: true` logged "Rate limiting: enabled (10/minute on
   /token)" while enforcing nothing - a "metadata never lies" violation.
   The configured `rate_limit_token_endpoint` now actually applies to
-  `POST /token` (429 with a JSON body and `Retry-After`/`RateLimit-*`
-  headers; every other endpoint stays unlimited). The two settings are
+  `POST /token` (429 with a JSON body and `Retry-After`/`X-RateLimit-*`
+  headers; every other endpoint stays unlimited), and the rate string is
+  VALIDATED at the config boundary: flask-limiter silently ignores a
+  malformed one and falls back to the (empty) defaults, so an unparsable
+  `rate_limit_token_endpoint` now refuses to load instead of silently
+  recreating the enabled-but-unenforced lie. No fallback value either. The two settings are
   also configurable from YAML at last (`server.rate_limit_enabled`,
   `server.rate_limit_token_endpoint`) - the fields existed on Settings
   but no document section carried them, so only the profile could flip
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -347,6 +347,20 @@ previous leniency allowed - needs a one-time adjustment.
   (RFC 3986 §3, e.g. `about:`). No audience bypass or escalation.
 
 ### Fixed
+- **Rate limiting on `/token` is enforced for real** (#304): the limiter
+  was constructed with no limits and no view was ever decorated, so
+  `rate_limit_enabled: true` logged "Rate limiting: enabled (10/minute on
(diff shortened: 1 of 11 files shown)


cdelmonte-zg:
Round 1 blocker fixed in 4caa87a, plus both non-blockers:

- **Rate string validated at the config boundary**: `Settings.validate_rate_limit_notation` runs `limits.parse` and refuses to load an unparsable value - deliberately with NO fallback, per your point that a default swapped in behind the operator's back is another configuration that lies. Pinned three ways: at the model, at the YAML load path (`SettingsDocument.to_settings`), and with the valid notations (`10/minute`, `1 per second`, `100/hour`) accepted. Verified live: booting with `rate_limit_token_endpoint: "banana"` fails with a ValidationError naming field and value. The wiring site carries a comment explaining WHY the boundary owns this (flask-limiter's silent-fallback-to-defaults behavior with defaults=[]).
- **`on_breach` on the /token limit itself** replaces the global `@app.errorhandler(429)`: a future limit on another endpoint no longer inherits an OAuth-flavored body. The existing 429 test still asserts `Retry-After`, confirming header injection applies to the on_breach response too.
- Docs/CHANGELOG corrected to `X-RateLimit-*` (the headers flask-limiter 4.1.1 actually sends).

Rebased on main (post-#313). Suite 1936, statics clean.

PR #315, opened 2026-08-30
refactor: one home for SAML attribute resolution (#302)

### LINKED ISSUE #302: refactor: SAML duplication - two assertion builders, two user-to-claims mappings
From the audit's SAML findings, two independent implementations of things that should have one home:

1. **Two assertion builders**: `_build_saml_response` (SSO, full AuthnStatement/AuthnContextClassRef/SubjectConfirmation) and `_build_attribute_query_response` + its signer (no AuthnStatement, no SubjectConfirmation). They agree today by coincidence, not construction; a change to NameID policy or audience handling must be made twice.
2. **Two user->claims mappings**: `_sso_build_attributes` (SAML attributes) vs the `extra` assembly in `TokenService.create_token` (OIDC claims). Both export identity_class/entitlements/email/attributes; SAML adds gated roles/groups, OIDC adds source_acl/groups/authorities unconditionally - divergence is undocumented and untested.

Proposal: extract the shared user->attribute resolution into one service consumed by both protocols (documenting the deliberate per-protocol differences as parameters, not parallel code), and share the assertion skeleton where the two builders genuinely overlap. One-home track; take it when SAML work next happens, this issue is the map until then.

cdelmonte-zg:
Closes #302 - with its residual debts explicitly re-tracked first, per the review round 1: this PR delivers the SAML-internal half (shared resolver + shared AttributeStatement emission); the SAML<->OIDC user-to-facts duplication is now **#316** and the shared assertion skeleton is **#317**, both filed with the analysis, and #302 carries the scope-resolution comment. Nothing closes by implication. Pulled into 3.0.0: every visible change here belongs in the major.

## What is shared now

- `services/saml_attributes.resolve_saml_attributes(settings, user, include_source_acl=)` - the one resolver both surfaces call (`_sso_build_attributes` and the query route's inline block are gone); `_add_export_attr` moved verbatim with its #134 merge-on-collision semantics.
- `append_attribute_statement` - one emission path for both builders.

## Three visible corrections (each finishes a rule the project already decided)

1. **No fabricated email**: the query surface invented `<user>@example.com` for a user without one - the #275 rule (never invent facts about a principal) now applies here too; an absent fact is an absent attribute on both surfaces.
2. **#134 completed**: custom LIST attributes reach the XML as one `AttributeValue` per entry (the query route used to `",".join` them) and a comma-bearing STRING is never split (the query builder used to re-split any string with a comma - directly contradicting the #134 comment sitting a few lines above it).
3. **No empty `Attribute` elements**: an empty collection is omitted (the SSO builder used to emit an empty element for `entitlements: []`).

## What stays deliberately different (documented in a table in saml.md)

`source_acl` only on the query surface (it exists for backend authorization lookups; a login assertion carries no document ACLs); no AuthnStatement/SubjectConfirmation/AudienceRestriction on the query assertion - **an attribute lookup is not an authentication event**, so this corrects the issue's own framing of those as gaps, with a route-level test pinning the absence as a declaration. NameID Format turned out identical on both surfaces (non-divergence).

## Verification

12 new tests (resolver branches incl. the memberOf collision merge; emission rules; cross-surface parity asserting the two surfaces differ ONLY by `source_acl`; route-level pins for all three corrections). Suite 1934 green, mypy/ruff/import-linter clean, full e2e 61/61 live.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -365,6 +365,22 @@ previous leniency allowed - needs a one-time adjustment.
   forced `rate_limit_enabled: true`, so its instances now really throttle
   `/token` at the configured rate (default 10/minute) - the hardening the
   profile always claimed.
+- **One resolver for SAML attributes; the query surface stops fabricating
+  and mangling values** (#302). The SSO assertion and the attribute-query
+  assertion resolved a user's attributes through two independent
+  implementations that had drifted five ways; both now share
+  `services/saml_attributes.py`, and the emission of the
+  `AttributeStatement` is one helper. Three visible corrections, each
+  finishing an existing rule: the query no longer invents
+  `<user>@example.com` for a user without an email (#275 - an absent fact
+  is an absent attribute); a custom LIST attribute reaches the XML as one
+  `AttributeValue` per entry and a comma-bearing STRING is never split
+  (#134 - the query used to `",".join` lists and re-split any string with
+  a comma); an empty collection no longer emits an empty `Attribute`
+  element. The differences that remain between the two surfaces
+  (`source_acl` only on the query; no AuthnStatement/SubjectConfirmation/
+  AudienceRestriction on the query assertion) are deliberate and now
+  documented in a table in the SAML reference.
 - **`RevocationStore` entries now expire** (#288): revoked jtis and rotation
   family markers lived in two sets that were never swept - every revocation
   and every refresh rotation on a long-lived instance was a permanent memory
--- a/book/src/reference/saml.md
+++ b/book/src/reference/saml.md
@@ -180,7 +180,10 @@ they cannot drift silently. Their remaining differences are deliberate:
 | `AudienceRestriction` | pinned to `oauth.audience` | absent | the query requester's audience is unknown (the endpoint is unauthenticated by design) |
 
 Shared rules on both surfaces: an absent or empty fact is an absent
-attribute (no fabricated `email`, no empty `Attribute` elements); a list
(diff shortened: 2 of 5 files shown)


cdelmonte-zg:
Round 1 findings addressed in 8ef9009 (rebased on main post-#313/#314):

1. **#302 does not close by implication**: took your second option - the two residual debts are now explicit issues, filed with the analysis: **#316** (one user->facts resolver across SAML and OIDC; `TokenService.create_token` still builds its own view of the same user facts, with the divergences enumerated) and **#317** (shared assertion skeleton; this PR shared only the AttributeStatement, and both builders still construct Response/Issuer/Status/Subject/Conditions independently - the deliberate SSO-only elements stay per saml.md's table). #302 carries a scope-resolution comment pointing at both; the PR body now says exactly what closes and what moved.
2. **`{}` is absent too**: the predicate is type-based now (`_is_absent`: None, or an empty list/tuple/set/dict/string), so `metadata: {}` is omitted instead of being emitted as the literal AttributeValue `"{}"`. Pins extended with `{}` and `()`, plus a non-empty dict staying kept. You were right that the doc's generic 'empty collection' promise and the literal-list implementation disagreed - the implementation now matches the promise.
3. **Merge-collision wording made precise** (the non-blocker): saml.md now says the guaranteed merge is the roles-vs-groups collision on the SAME configured export name, and that a collision between an export name and a scalar core attribute is not merged - the export list replaces it.

Suite 1947, statics clean.

PR #318, opened 2026-08-31
feat: auto login personas

### LINKED ISSUE #250: feat: Auto-Login personas for automated integration testing
For automated integration tests, it would be nice to perform the authentication flow without user-input. In short: if persona mode, and auto-login is enabled, and client sets the `login_hint` to `persona-auto-login:USERNAME`, nanoidp doesn't show any login form / persona selection, but directly authenticates the configured user.

* By using a common prefix like `persona-auto-login:` we can validate the intent to auto login. So when something went wrong, like the username is unknown, we can fail fast and return e.g. `400 Bad Request`, because can we shouldn't display HTML and potentially block the automated test.
* `login_hint` is widely supported in client libraries, and to my opinion it is not misused here, as it is there to provide the IDP with an hint of the user to be logged in.
* It should be supported by all mayor flows 
* I did some basic research for SAML compatibility, which doesn't have the `login_hint` directly. SAML is not relevant for my specific app, but maybe it would be nice to support this here as well? I'm not a SAML specialist, but maybe it could be placed into `saml:NameID` with `Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"` like shown below:

```xml
<samlp:AuthnRequest ...>
  <saml:Subject>
    <saml:NameID Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified">
      persona-auto-login:USERNAME
    </saml:NameID>
  </saml:Subject>
</samlp:AuthnRequest>
```

gprossliner:
Implements the auto-login contract from #250: with `login.mode: persona` and the new opt-in `login.auto_login: true`, an OIDC `/authorize` request whose `login_hint` is `persona-auto-login:USERNAME` authenticates that persona directly and issues the authorization code - no picker, no HTML - for driving a real OIDC client library in automated integration tests. An unknown persona reports through the standard OAuth error redirect (`error=invalid_request`, `state` preserved, `iss` attached), never a bare 400. Any other `login_hint` stays an ordinary ignored hint, and with the flag off a prefixed hint is inert too. First surface is OIDC `/authorize` only (contract point 5); device flow and SAML have no defined transport for the hint today.

Ships whole per contract point 6: config schema + persistence (`login.auto_login`, omitted at its `false` default), settings UI checkbox, `/api/config` (a new `login` block that also closes the pre-existing gap of `login_mode` never being exposed there), MCP `get_settings`/`update_settings`, e2e agent coverage (flag off -> picker; known persona -> code -> token; unknown persona -> error redirect with `state`), and docs (`book` configuration reference, SECURITY.md, `examples/persona-login`).

## Implementation assumptions (`#250-assumption`, resolved during review)

The design contract left these as judgment calls; each was resolved with a concrete default and confirmed by the maintainer in review round 1:

1. **`auto_login: true` without `login.mode: persona` is inert, not rejected.** `Settings` accepts the combination; the derived `auto_login_enabled` property (`persona_mode_enabled and auto_login`) is simply `False`, so a stray prefixed hint falls through to the ordinary flow. Matches how profile and login-mode settings already compose orthogonally.
2. **No new `auth_method` value, `routes/saml.py` untouched.** The contract's SAML concern presumed `/authorize` would establish a session; implementation found its inline login (password, picker, or auto-login) never writes `session["user"]`/`session["auth_method"]` - it is a one-shot code issuance. Contract point 4's distinguishability is carried by the audit record instead: `details.auto_login: true` on the success event, and the attempted username as a structured field on the failure event.
3. **Errors reuse `_authorize_error_redirect`** - the same post-`redirect_uri` channel every other `/authorize` validation failure uses, `state` and `iss` included.
4. **`/api/config` exposes `login_mode` alongside `auto_login`.** Shipping `auto_login` alone would have repeated the #165 "blank = clear" wipe on the next settings round-trip, since the settings form and e2e agent rebuild from this endpoint. Deliberate small scope expansion, guarded by `tests/test_api_config_parity.py`.

## Review record

- Round 1: two blockers (a stale session `login_hint` could trigger auto-login on requests that never sent one, and a test-class header lost to an edit anchor), plus non-blocking polish - see the review thread.
- Round 2: both blockers fixed (`login_hint` is read from the current request only and never stored in the session; regression tests reproduce both measured cases) and re-verified live by the maintainer.
- Round 3: failed auto-login audits the attempted `username` as a structured field, working-plan doc removed (its assumption call-outs folded into this description), optional cleanups applied (shared `_login_settings_defaults()`, shared test helpers, stale docstring count dropped).

CHANGELOG entry is left to the maintainer at merge time, per the persona-login-mode precedent.

Closes #250.


cdelmonte-zg (changes requested):
# Review round 1 (head c4c71ff)

Thanks @gprossliner - this is a well-built PR. The contract from #250 is implemented faithfully on the points it names: the auto-login branch runs after every other `/authorize` validator, so an unknown persona reports through `_authorize_error_redirect` with `state` preserved and `iss` attached (verified live, never a bare 400); the flag composes with persona mode exactly as agreed; a non-prefixed hint stays an ordinary ignored hint; the audit record distinguishes the two persona variants; and the feature ships whole (settings UI, `/api/config`, MCP, e2e, docs, schema). All four `#250-assumption` call-outs in the plan are accepted:

1. **Inert, not rejected** without `login.mode: persona` - agreed, matches the composition precedent.
2. **No session/`auth_method` change needed** - verified against the code: `/authorize`'s inline login never writes `session["user"]`/`session["auth_method"]` (nothing in `routes/oauth.py` touches them), so `routes/saml.py`'s literal `== "persona"` check has nothing to regress against. Good catch revising the plan mid-implementation instead of adding a guard for a session that does not exist.
3. **Reusing `_authorize_error_redirect`** - correct, that is exactly what it is for.
4. **`/api/config` gets `login_mode` alongside `auto_login`** - verified: `main`'s `get_configuration()` indeed never exposed `login_mode`, and shipping `auto_login` alone would have recreated the #165 wipe on the next settings round-trip. The parity test is exactly the right guard. Deliberate small scope expansion accepted.

What I verified empirically at `c4c71ff`: full suite 1974 passed, ruff/mypy/import-linter clean, CI 9/9 including e2e 62/62 with the new check; live against a booted instance: known persona goes straight to a code that exchanges for a token, unknown persona and empty username both produce the `invalid_request` redirect with `state`, flag off leaves a prefixed hint inert (picker shows), an ordinary `login_hint` is untouched.

Two blocking findings, both reproduced live.

## Blocking

### 1. A stale session `login_hint` triggers auto-login on requests that never sent one

`_read_authorize_params()` gives `login_hint` the same session-fallback-and-store treatment as the other parameters. That parity was a reasonable instinct, but `login_hint` is the first parameter for which the fallback changes *who gets authenticated*, and the stored hint survives every leg that does not end in a successful code issuance: `_issue_authorization_code` clears the `oauth_*` keys, `_authorize_error_redirect` and an abandoned picker clear nothing. Two measured consequences on a booted instance:

- **The feature's own inertness promise breaks.** Flag on; `GET /authorize` with `login_hint=persona-auto-login:ghost` returns the error redirect (correct) and stores the hint; a follow-up `GET /authorize` from the same browser session with **no** `login_hint` at all returns the error redirect again instead of the picker, and keeps doing so until cookies are cleared. The docstring's contract ("`login_hint` is absent/unprefixed - the caller then falls through to the ordinary login page/picker") does not hold for any session that ever carried a prefixed hint.
- **A picker selection can be silently overridden.** Flag off; `GET /authorize` with `login_hint=persona-auto-login:alice` shows the picker (correct) and stores the hint; the flag is then enabled; the user submits the picker form with `username=bob` - the POST leg resurrects the session hint before `_handle_authorize_login` ever reads the form, and the response is a code redirect whose token has `sub = alice`. The explicitly selected user loses to a session leftover from before the flag was even on.

The requirement (implementation is yours): auto-login must act only on a `login_hint` present in the current request, never on one replayed from the session, and a submitted picker selection must never be overridden by a hint. Note the auto-login branch never renders a form, so unlike the other parameters it has no POST leg that needs the session fallback at all. Please add tests for both measured cases above.

### 2. `tests/test_config.py` eats the `class TestUsersYamlPasswordOptional:` header

The new `TestAutoLogin` class replaced the `class TestUsersYamlPasswordOptional:` line instead of being inserted before it. Its docstring ("Tests for loading users without a password...") is now a stray string literal inside `TestAutoLogin`, and the password-optional tests silently became `TestAutoLogin` methods. Nothing fails - collection goes 68 to 73, no test is lost - which is exactly why it has to be caught in review rather than by CI: `pytest -k TestUsersYamlPasswordOptional` now selects nothing, and the passwordless-users regression tests report under an unrelated class. Same shape as the first blocker of #247 round 1, so worth a look at whatever produces these edits. Restore the class header above the stranded docstring.

## Before merge (not blocking, and explicitly not a third blocker)

- **Pass the attempted username into the failed auto-login audit event.** A failed picker/password attempt audits `username=<attempted name>` (`_handle_authorize_login`), while a failed auto-login goes through `_authorize_error_redirect` and carries the name only inside the free-text `reason`. An audit consumer filtering failed `authorization_request` events by the `username` field sees every `persona-auto-login:<name>` probe as anonymous - an odd gap next to this PR's own audit-distinguishability point.
- **Drop `docs/plans/auto-login.md`**, per the #247 precedent and its own header ("remove or fold ... once the feature ships"). The four `#250-assumption` call-outs are the part with lasting review value - fold those into the PR description (the plan says they would be there, and they currently are not), then delete the file.
- **One sentence for the new SECURITY.md section**: with `auto_login` on and a client that has no registered `redirect_uris` (the permissive dev default), a plain GET link mints a code for any persona to any `redirect_uri` with zero interaction - measured live. That permissiveness is pre-existing and out of scope here, but auto-login removes the one human click persona mode still required, so the section should recommend registering `redirect_uris` for any client used with auto-login.

## Optional cleanups (take or leave, none gates the merge)

- `yaml_writer.py`: `update_login_settings` and `update_settings_form` now share a verbatim `if mode / if auto_login` mutate pair and each calls `document_defaults()` twice (each call rebuilds a full `SettingsDocument`); one shared helper and a single `defaults` local would do, and `update_login_settings`'s early no-op branch is redundant now (the mutate closure is already a no-op when both are `None`).
- Test reuse: `AUTO_LOGIN_QS` re-inlines `AUTHORIZE_QS` byte for byte plus one param (`AUTHORIZE_QS + "&login_hint=..."` keeps them in sync), `_enable_auto_login` re-inlines `_enable_persona_mode`, and `TestSettingsUiAutoLogin` carries the file's third copies of `_login_as_admin`/`_base_form` - the #165 form-drift scenario is exactly what stale copies of `_base_form` would reproduce.
- `_AuthorizeParams` docstring: "the ten /authorize request parameters" - the dataclass has eleven fields (the old "nine" was already stale since #187; incrementing inherited the miscount). Drop the number rather than maintain it.

CHANGELOG: leaving it to the maintainer per the #247 precedent is fine - the draft is good and will be used at merge, with the milestone decision (3.0.0-in-rc vs 3.1) resolved then per the policy on #250.

cdelmonte-zg (commented):
# Review round 2 (head 2c30de1)

Both blockers are fixed, in exactly the shape asked for, and I re-verified everything empirically at this head. Fast turnaround, and thanks for the honest root-cause note on the eaten class header - "the edit's find-text included the header, the replacement didn't" is precisely the failure shape worth remembering.

**Blocking 1 (stale session hint) - CLOSED, re-measured live.** `login_hint` now comes from the current request only and is never stored; both round-1 repros re-run against a booted instance from this head:

- Unknown-persona probe, then a hint-less `GET /authorize` from the same browser session: **200, picker shown** (was: error redirect forever).
- Hint `persona-auto-login:alice` sent while the flag was off, flag then enabled, picker POST with `username=bob`: **code minted for `bob`** (was: `sub = alice`).

Both regression tests match the measured cases, and the field comment on `_AuthorizeParams` documents why this parameter is the exception to the session-fallback pattern - future readers will need exactly that sentence.

One adjacent behavior I probed and am fine with, noting it for the record: a hand-crafted POST carrying *both* `username=bob` and `login_hint=persona-auto-login:alice` in its body mints the code for `alice` - the request-borne hint wins over the posted username. That is consistent with the contract (a request that sends the prefixed hint is asking for auto-login), and the real picker form never sends a hint, so the round-1 harm (a *session leftover* overriding an explicit choice) is gone. No change requested.

**Blocking 2 (class header) - CLOSED.** `TestUsersYamlPasswordOptional` is back above its docstring; `pytest -k TestUsersYamlPasswordOptional` selects its test again (1, same as on `main`), collection parity holds (68 on `main`, 73 here, the +5 all `TestAutoLogin`).

Also verified at `2c30de1`: full suite 1976 passed (the +2 are the new regression tests), ruff/mypy clean, CI 9/9, full-branch diff vs merge-base is additive relative to round 1 (no rebase, nothing shrank). The SECURITY.md sentence on registered `redirect_uris` is in and says the right thing.

## Still open from round 1's "before merge" (then this is an approve)

1. The attempted username in the failed auto-login audit event.
2. Drop `docs/plans/auto-login.md`, folding the four `#250-assumption` call-outs into the PR description first. Your round-1 log inside it is a good review artifact - the PR thread now carries the same record, so nothing is lost.

The optional cleanups remain exactly that - take them or leave them. Once 1 and 2 land I will approve; the milestone call (3.0.0-in-rc vs 3.1) happens at merge per the policy on #250.

cdelmonte-zg (approved):
# Review round 3 (head cbacf3f) - APPROVED

Both remaining items are in, and you took the optional cleanups too. Verified at this head: suite 1977 passed (the +1 is your new audit test), ruff/mypy/import-linter clean, CI 9/9.

- **Audit username**: exactly right - the keyword-only `username` on `_authorize_error_redirect` with a `None` default is behavior-identical for every other caller (`audit_event` already defaulted `username=None`), and only the auto-login path passes it. Confirmed live: an unknown-persona probe now lands in the audit log as a structured `"username": "ghost3"` next to the free-text reason.
- **Plan doc**: removed. I have folded its four `#250-assumption` call-outs into the PR description (which still pointed readers at the deleted file), along with a short review-record section - nothing you need to do there.
- **Cleanups**: `_login_settings_defaults()` reads well, dropping `update_login_settings`'s early no-op branch is behavior-identical (the mutate closure was already a no-op with both fields `None`, and `expected_revision` is still honored), the shared test helpers close the `_base_form` drift risk, and `AUTO_LOGIN_QS = AUTHORIZE_QS + ...` keeps the two suites on the same request. Dropping the parameter count from the docstring was the right call over maintaining it.

Thanks for three fast, precise rounds - and for the honest root-cause notes along the way. The CHANGELOG entry and the milestone call (3.0.0-in-rc vs 3.1, per the policy on #250) happen at merge.

gprossliner:
I addressed the blocking and non-blocking issues.

A comment: While working on this PR warnings appeared when running the suite (`1974 passed, 8 warnings `). I analysed them and they seem unrelated to the PR, as they also exist when ran against `main`. It seems to be race conditions (`test_thread_safety`, `SlowInit` with `time.sleep(0.05) `) in test execution. Is this only on my system?

I validated that 59ebd355 introduced it on my system. This is the warning:

```
tests/test_thread_safety.py::TestSingletonConcurrency::test_get_crypto_service_creates_one_instance
  /Users/guenter.prossliner/proj/nanoidp/.venv/lib/python3.14/site-packages/_pytest/threadexception.py:58: PytestUnhandledThreadExceptionWarning: Exception in thread Thread-95 (get)
  
  Traceback (most recent call last):
    File "/opt/homebrew/Cellar/python@3.14/3.14.4/Frameworks/Python.framework/Versions/3.14/lib/python3.14/threading.py", line 1082, in _bootstrap_inner
      self._context.run(self.run)
      ~~~~~~~~~~~~~~~~~^^^^^^^^^^
    File "/opt/homebrew/Cellar/python@3.14/3.14.4/Frameworks/Python.framework/Versions/3.14/lib/python3.14/threading.py", line 1024, in run
      self._target(*self._args, **self._kwargs)
      ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/guenter.prossliner/proj/nanoidp/tests/test_thread_safety.py", line 106, in get
      obj = getter()
    File "/Users/guenter.prossliner/proj/nanoidp/src/nanoidp/services/crypto.py", line 519, in get_crypto_service
      service.keys_dir != Path(keys_dir) and not service.uses_external_keys
      ^^^^^^^^^^^^^^^^
  AttributeError: 'SlowInit' object has no attribute 'keys_dir'
```

cdelmonte-zg:
Not only your system - reproduced here on Linux/CPython 3.12, same `AttributeError` warnings, and your bisect is exact: 59ebd35 is the #281 `keys_dir` fix, which added `service.keys_dir` / `service.uses_external_keys` reads to `get_crypto_service`'s fast path and in-lock check. The generic `SlowInit` stub models the pre-#281 getter (which only checked `is None`), so every racing thread that observes the published stub now raises.

Two things worth stating precisely:

- **Not a product bug.** `CryptoService.__init__` assigns `self.keys_dir` as its first statement, `uses_external_keys` is a property, and the getter publishes the instance only after the constructor returns - a real caller never observes a half-built service. The race window your warning shows exists only against the attribute-less test double.
- **It is worse than a warning, though.** I instrumented the test's own scenario: of the 8 racing threads, 7 die on the AttributeError and only 1 reaches `seen`, so the "all threads got the same instance" assertion is checking a single element - the crypto leg of that concurrency test is currently vacuous.

Tracked as #320 (3.1) with the fix shape: teach the stub the post-#281 contract (store `keys_dir`, expose `uses_external_keys = False`) and pin `len(seen) == n` so the test cannot go silently vacuous again. Out of scope for this PR, and thanks for the report and especially for arriving with the introducing commit already bisected - that made this a ten-minute confirmation instead of an investigation.

cdelmonte-zg:
**v3.0.0-rc3 is out with this feature**: `pip install --pre nanoidp==3.0.0rc3` (or `docker pull ghcr.io/cdelmonte-zg/nanoidp:v3.0.0-rc3`). Verified on the published wheel: known persona -> code -> token, unknown persona -> the error redirect with the structured audit username. If you get the chance, pointing your real client-library test suite at it during the soak is exactly the confirmation the final needs - the [release notes](https://github.com/cdelmonte-zg/nanoidp/releases/tag/v3.0.0-rc3) have the rc3-specific asks. Thanks again for the contribution.

cdelmonte-zg:
@gprossliner **v3.0.0 is out** as a final: `pip install nanoidp` now resolves 3.0.0 and `ghcr.io/cdelmonte-zg/nanoidp:latest` moved to it. Auto-login personas ship in it exactly as merged here, with your credit in the [release notes](https://github.com/cdelmonte-zg/nanoidp/releases/tag/v3.0.0) and the CHANGELOG. The rc3-pinned commands in this thread are stale now - the plain install is the one to use. Thanks again for the contribution and for the soak feedback (the `SlowInit` warnings you bisected are tracked in #320 for 3.1).

gprossliner:
I'm using auto login now in my integration tests! Works nicely now.

cdelmonte-zg:
super! tnx for the good news :)

PR #323, opened 2026-09-05
feat: global two-step login (login.two_step) across /authorize, /login, /saml/sso and /device (#322)

### LINKED ISSUE #322: feat: two screen authorize
imho it is more modern way to enter username and password in two sequential screen. 1st user entered username , then click next and get screen to enter password.

for autologin, it will be just skip second step.

dshvedchenko:
Adds `login.two_step` (#322), a global opt-in that collects the username on one screen and the password on the next, the identifier-first shape of Entra ID, Okta, Google and Auth0, so a browser-driven integration suite written against one of those providers can rehearse the same page sequence locally.

## What it does

- **One global setting**, `login.two_step: true` (default `false`), next to `login.mode` and `login.auto_login`. Not per client: no real provider is identifier-first for one client and single-form for another.
- **Applied wherever the password form renders**: `/authorize`, `/login`, `/saml/sso` (POST and Redirect binding) and the device flow's `/device`. Inert under `login.mode: persona`, which is passwordless.
- **Stateless.** The username travels as a plain form field and is carried forward as a hidden input on the password screen. Whether a submission is the username-only step or a login attempt is derived from the fields it carries, by one shared helper (`routes/_auth.py`, `two_step_phase`). A POST that already carries username and password authenticates directly, so anything written against the combined form keeps working with the flag on.
- **"Change username"** returns to the first screen on every surface. On `/authorize` the link carries the request's own parameters; on `/saml/sso` it is a small POST form re-submitting the in-progress SP request's hidden fields, since that surface has no GET to link back to.
- **Configurable** through YAML, the Settings UI, the MCP settings tools and reflected by `/api/config`; `docs/schema/config.v1.json` regenerated; documented under "Two-step login" in the configuration reference; CHANGELOG under `[Unreleased]`.

## Review history

Four rounds. Round 1 replaced a session-based step sentinel with the stateless shape (a combined POST was being half-consumed, and a captured username could cross clients). Round 2 moved the flag from a per-client field to the global setting and extended it to all four surfaces. Round 3 fixed the SAML password screen losing the Redirect binding's original verb (under `saml.strict_binding: true` the AuthnRequest failed to parse and the response went to the default ACS; regression test added), restored the SAML "Change username", moved quick-fill onto the username screen, and consolidated step detection into the shared helper. Round 4 ordered the helper's blank-username check before the password short-circuit, made Deny clickable on the device screens (`formnovalidate`), fixed the device autofocus, unified the `/login` failure string and whitelisted `saml_original_verb` where it is read.

The pre-existing behaviour of the `/authorize` POST leg reading OAuth parameters from the form body with a session fallback (retarget, cross-tab clearing and overwrite) is tracked separately in #325. A follow-up may fold the three copies of the two-step markup into `_macros.html`.

Closes #322


cdelmonte-zg (changes requested):
Thanks - this is a clean, complete plumbing job: the field reaches YAML (omitted when default), the client UI form, both MCP tools and the serializer, the config schema, the e2e harness and the reference page, and the GET-time reset shows the #318 lesson was taken. Step 1 does not verify that the username exists, so the flow adds no username enumeration, and the `autocomplete` attributes make password managers work across the two screens. Verified on the head (57c7fcc): ruff/mypy/import-linter clean, 1987 unit tests, CI 9/9, and a live server with `two_step_login: true` driven through the happy path and the negatives (empty username, direct password POST, wrong password, change username, HTML in the username - escaped, new GET resets).

Two behaviours reproduced live are blocking, plus two before-merge items. Review round 1.

## Blocking

**1. A POST that carries a password is silently half-consumed.** For an opted-in client, every POST without `login_step=password` is the username step. A scripted client written against the existing combined form (GET, then POST `{username, password}` - exactly what `e2e/_run_auth_code_flow` does) gets a `200` password screen with "Signing in as admin", no code, no error, no audit event; the submitted password is discarded. Reproduced: flag off -> `302` with a code; flag on, same POST -> `200`. Whatever the final shape, a request carrying full credentials must either authenticate or be refused, never partially applied.

**2. The captured identity does cross clients, and the new comment says it never does.** The `session.pop` runs only on GET, while a password-step POST still rebinds `client_id`/`redirect_uri`/`state` from the form body through `_read_authorize_params`. Reproduced with two opted-in clients:

```
GET  /authorize?client_id=demo-client&...&state=sA
POST /authorize login_step=username&username=admin
POST /authorize login_step=password&password=admin&client_id=client-b&redirect_uri=http://evil.example/cb&state=sB
-> 302 http://evil.example/cb?code=...&state=sB
   log: Authorization code issued for user 'admin', client 'client-b'
```

The password is still verified and the POST-body retarget plus the permissive `redirect_uri` for clients without `redirect_uris` are pre-existing, so this is not an authentication bypass. But the user never saw a login page for client-b, and the invariant written at `_read_authorize_params` is false. Either bind the captured username to the client it was captured for (drop it when `p.client_id` differs), or narrow the comment to what the code guarantees - and pin whichever with a test.

**Recommended shape (fixes both, and more):** make the step stateless. Carry the username as `<input type="hidden" name="username">` on the password form, derive the step from the submitted fields (`two_step_login and not password` -> username screen, otherwise the ordinary `interactive_authenticate(username, password)`), make "Change username" a plain link to `GET /authorize` (the session fallback already restores the request parameters), and pass `login_username` from the form. That removes the `oauth_login_username` key, the three-valued `login_step` sentinel, the GET-time pop (which today also fires on an unrelated GET in another tab, dropping the captured username and showing a misleading "Username is required"), the config-changed-mid-flow edge (flip the flag or `login.mode` while a user sits on the password screen -> "Username and password are required" with the typed password lost), and the `two_step_login` predicate computed twice (handler and renderer - CONTRIBUTING "Domain invariants have one home"). No security trade-off: a tampered hidden username hits the same credential check as the combined form, and step 1 never verified the username anyway. `/login` and `/device` are already single-POST with no step state, so this is also the consistent shape. If you prefer to keep the session design, 1 and 2 need explicit fixes and tests instead.

## Before merge

**3. CHANGELOG placement.** Main released 3.0.0 (a570033) after this branch's base, so a trial merge puts the #322 entry inside `## [3.0.0] - 2026-09-06` (line 190 of the merged file) while `## [Unreleased]` is empty. Rebase onto main and re-home the entry under a new `### Added` in `[Unreleased]`.

**4. Unconditional `align-items-center` -> `align-items-start` on the #249 horizontal row** (57c7fcc). Every existing `layout: horizontal` client renders differently after this PR, while the CHANGELOG says existing clients are unchanged. Keep the centred alignment unless the "Signing in as" block is present, or document the change as intentional in the entry.

**5. Docs and scope.** `two_step_login` reaches the reference page only as a YAML comment; its two rules (inert under `login.mode: persona`; `/authorize` password form only - `/login`, `/device` and the SAML SP form keep the combined form) live only in the client-form help text. One or two sentences in the "Login page branding" paragraph, following the #250 subsection's "inert unless" wording. Related: #322 as written reads as a global presentation preference; the PR models it per client, `/authorize` only, which is consistent with #249 and fine by me, but say so on the issue and in the docs so nobody expects it on `/login`.

**6. Tests to pin the contract**: the tampering case (a form `username` on the password step is ignored in favour of the captured one - or, with the stateless shape, the combined-POST-with-flag-on case), and "Password is required" (the branch exists, no test reaches it).

**7. PR body**: the last line is truncated ("...and MCP"), and add `Closes #322` - this PR fully covers the issue as scoped above.

## Non-blocking

- The e2e assertions are folded into `test_client_branding`, so a password-step regression reports as a #150 branding failure. A dedicated `test_two_step_login` with its own client (only `two_step_login` on) and a fresh `requests.Session()` like `_run_auth_code_flow` keeps the names honest; the branding GET can then go back to bare `requests.get`.

Milestone: 3.1 (post-3.0.0 feature).

cdelmonte-zg:
Follow-up to the review, after the discussion on #322: the feature is accepted, in the global shape described there. Concretely for round 2:

1. Move the flag from `OAuthClient.two_step_login` to `login.two_step` (global, default `false`), following `auto_login` (#318) through `LoginDocument`, `Settings`, the MCP config handlers / schema / normalize list, `/api/config`, the UI settings form and the docs `login:` reference. Remove the client field, its UI checkbox, the MCP client-schema entry and the export/import wiring.
2. Apply it wherever the password form renders, not only `/authorize`: `/login`, the device flow, the SAML login.
3. Stateless step: hidden `username` on the password form, step derived from the submitted fields (username only = username step; username + password = authenticate), "Change username" as a link back to the GET. This drops the session key and closes the two blockers from the review.
4. The remaining points of the review stand: rebase on `main` so the CHANGELOG entry lands under `[Unreleased]`, keep the horizontal row alignment unchanged for existing clients, docs prose for the reference page, tests for tampering and for "Password is required", `Closes #322` in the body.

Milestone stays 3.1. Take your time; I will re-run the live probes on the new head.

dshvedchenko:
already rebased and work toward 3.1

cdelmonte-zg:
Verified 6512465c (ruff, mypy, import contracts, 1990 unit tests, plus the live probes from round 1 re-run against a two-step client, a plain client and a third client with a foreign redirect_uri): both blockers are gone. A combined username+password POST authenticates and mints the code, the username-only POST renders the password screen with the username as a hidden field, a blank password gives "Password is required", a wrong one stays on the password screen and is audited, and opening `/authorize` for another client shows no prefilled identity. CHANGELOG under `[Unreleased]`, docs, the conditional horizontal alignment and the tests all look right. Thanks, that was a clean round.

Two things for the next push:

1. The global shape from my comment above (`login.two_step` next to `mode` and `auto_login`, applied to every password form) is the only structural change still open. Your push crossed with that comment, so no surprise it is not in here yet.
2. Small one: the "Change username" link is a bare `url_for('oauth.authorize')`, so it only works while the session still holds the `oauth_*` keys from the GET; once those are consumed (a successful login in another tab) it answers 400 `client_id is required`. Carrying the request's query string on the link keeps it self-contained.

For the record, a POST whose body carries `client_id`/`redirect_uri` still retargets the code to that client after a valid password. That predates this PR (`_AuthorizeParams` reads POST params from the form with a session fallback), so it is not on you here; I will track it separately.

dshvedchenko:
Round 2 is fully implemented and verified: ruff/mypy/import-linter clean, 2012 unit tests passing, and live-server probes confirming every flow.

What changed:

Global setting: login.two_step (alongside login.mode/login.auto_login) replaces the per-client OAuthClient.two_step_login field entirely — removed from the model, ClientEntry, YAML serialization, MCP client schema/handlers/serializer, and the client-form checkbox. Added throughout LoginDocument/Settings (with a new two_step_login_active property — the single shared predicate), the YAML writer, MCP settings schema/handlers/normalize allowlist, /api/config, and the Settings UI checkbox. docs/schema/config.v1.json regenerated.

Applied to all four password-form surfaces: /authorize (already had it), /login, /saml/sso (reusing login.html, carrying SAMLRequest/RelayState/original_verb through both new screens), and /device (short-circuits before ever touching the device-code store, so a username-only POST can't disturb a pending code's status; Deny stays reachable without credentials).

Stateless everywhere: same design as round 1 — hidden username field, step derived from submitted fields, combined POST authenticates directly.

Your new comment's fix: "Change username" on /authorize now links to url_for("oauth.authorize", 
** current_request_params) instead of a bare URL, so it survives the session's oauth_* keys being cleared by an unrelated completed login. Verified live (reproduced the 400, then confirmed the fix) and added a regression test. /login's and /device's equivalent links never had this problem (no session dependency); SAML has no such link at all, documented as a deliberate gap.

Tests: rewrote test_two_step_authorize.py for the global setting, added test_two_step_login_flows.py (per-surface, mirroring test_persona_login_flows.py) and TestSettingsUiTwoStep (mirroring TestSettingsUiAutoLogin) in test_persona_login.py. e2e's test_two_step_login now toggles the global setting and covers both /authorize and /login, plus the cleared-session regression.

CHANGELOG/docs: rewrote the [Unreleased] entry and added a "### Two-step login" section in the configuration reference, removing the stale per-client mentions.

cdelmonte-zg (changes requested):
Round 2 on 210ea002. Thanks, this is the shape agreed on #322 and most of it holds up. Verified: ruff, mypy, import contracts, 2012 unit tests green; live on a fresh server with `login.two_step: true` I walked `/authorize`, `/login`, `/device` and `/saml/sso` (POST binding and Redirect binding) through both screens with the negatives: blank username, blank password, wrong password, combined POST, "Change username" with no cookie at all, device Deny from the username step, device code left pending by the username step, persona-mode inertness on all four surfaces, Settings UI uncheck/re-check round trip with the YAML and `/api/config` following, `two_step: maybe` rejected by validate-config. All of that behaves as described.

One blocker, then a handful of things to fix before merge, then nits.

**Blocker: the SAML password screen loses the original binding verb.** `_sso_authenticate_inline` renders every screen with `original_verb=request.method`. The username screen is rendered from the SP's GET, so it carries `saml_original_verb=GET`; the password screen is rendered from the username-only POST, so it carries `saml_original_verb=POST` while its hidden `SAMLRequest` is still the deflated Redirect-binding value. Reproduced live on a server with `saml.strict_binding: true`:

```
GET  /saml/sso?SAMLRequest=<deflated>&RelayState=rs      -> username screen, saml_original_verb=GET
POST /saml/sso  SAMLRequest=<deflated> username=admin saml_original_verb=GET
                                                        -> password screen, saml_original_verb=POST
POST /saml/sso  SAMLRequest=<deflated> username=admin password=admin saml_original_verb=POST
     log: "Failed to parse SAMLRequest: Start tag expected, '<' not found"
     -> 200, SAMLResponse auto-posted to the default ACS (http://localhost:8080/login/saml2/sso/...), not to the SP's ACS, InResponseTo lost
```

The same POST with `saml_original_verb=GET` (what the single-screen form sends) issues the response to `http://sp.example.com/acs`. With `saml.want_authn_requests_signed: true` the outcome is a 400 instead: a `POST` leg is re-verified as an enveloped signature on bytes that are not XML. Without strict binding it only works because the parser auto-detects compression, which is why the current tests pass. Redirect binding is what Spring Security SAML sends by default, so this is the flag's own use case.

Fix: the screens must carry the verb of the request that entered the flow, not the verb of the request that rendered them: in `render_login`, `original_verb=request.form.get("saml_original_verb") or request.method` (the parser already validates the value in `_sso_parse_request`), plus a test under `strict_saml_binding=True` that drives the Redirect binding through both screens and asserts the response goes to the AuthnRequest's ACS.

**Before merge**

1. **`test_password_step_ignores_client_swap_via_form_body` pins the wrong thing.** The name says the swap is ignored, but the assertion is that the code is issued on the swapped-in `redirect_uri`, i.e. the test locks in the pre-existing form-body retarget that #325 removes. Keep the point you wanted (the username is what this request submitted, nothing resurrected from a previous client), drop the `Location` assertion, and point the docstring at #325.
2. **`device.html`: focus lands in the device-code field on the password screen.** The `user_code` input keeps its `autofocus` and both new inputs add one; the first in DOM order wins, so on the password screen the user types the password into the pre-filled code field (maxlength 8, uppercased by the input listener) and overwrites the code. Make the `user_code` autofocus conditional on `not login_username` (or drop it on the two-step screens).
3. **SAML dead end on a mistyped username.** With `{% if not saml_request %}` there is no "Change username" on `/saml/sso`, and a wrong password re-renders the password screen with the same hidden username, so "admn" can only be escaped with the browser's Back button. The docs claim "no session-independent way back into an in-progress SP request", but the password screen already carries `SAMLRequest`, `saml_original_verb` and `RelayState` as hidden inputs, and `_sso_authenticate_inline` already renders the username screen for a POST that carries no `username` field. A second small POST form re-submitting only those hidden fields restores the affordance with no route change, and the docs sentence goes away.
4. **Quick-fill username disappears under two_step.** The "Quick fill username" buttons live only in the combined-form branch of `login.html`, so opting in removes them from `/login` and the SAML login even though both routes still pass `users`. The username screen is where they belong; `fillUsername()` needs a null guard on the password focus there.
5. **The step-detection rule has four hand-written copies** (`oauth.py` `_handle_authorize_login` and `device_verify`, `saml.py`, `ui.py`) and they already diverge (device adds `action != "deny"`, SAML adds `username_submitted`, the no-form fall-through differs). CONTRIBUTING's "domain invariants have one home" applies: one pure helper, e.g. in `routes/_auth.py`, returning `username_step | username_required | password_required | attempt` from the form, with each route keeping only its transport. Round 1's half-consumed POST was exactly the kind of drift this prevents.
6. **PR title** still says "per-client"; the squash inherits it. Something like `feat: global two-step login (login.two_step) across /authorize, /login, /saml/sso and /device (#322)`. The `210ea00` commit body also carries an em dash; the repo uses plain hyphens.

**Nits, take or leave**

- The username form, the password form, "Signing in as" and "Change username" are three byte-identical copies across `login.html`, `authorize.html`, `device.html`, and the SAML hidden-field block went from two to four copies. `_macros.html` exists for exactly this; a couple of macros next to `persona_picker` would give them one home. Fine to leave for a follow-up.
- `/login` reports a failed password inline under two_step but via `302 ?error=` otherwise, and a POST with a blanked hidden username plus a password falls through to the legacy redirect branch. One message and one transport would be simpler.
- The device username step returns before `verify()`, so a wrong or expired code is only reported after the password. If that is the deliberate trade-off (no non-mutating status lookup on the store), say so in the comment; today it reads as if the placement were forced.
- e2e `test_two_step_login` says it "restores the setting afterward" but the `finally` writes `two_step=false` unconditionally; either reword or read `/api/config` first, the way `test_saml_exclusive_c14n` restores the c14n setting.

For the record on #325 (pre-existing, not for this PR): the two-step forms on `/authorize`, like the combined form before them, rely on the session's `oauth_*` fallback, so a login completed in another tab still 400s a pending password POST; the self-contained link fixes the GET leg only. Whether the forms should carry the request or the POST leg should bind to the session is the design call in #325, so nothing to do here.

Once the SAML verb fix and the before-merge items are pushed I will re-run the strict-binding probe and the rest of the checklist on the new head.

dshvedchenko:
pls verify

cdelmonte-zg (changes requested):
Round 3 on 00f833b1. The blocker is closed: on a `saml.strict_binding: true` server the Redirect binding now goes through both screens with `saml_original_verb=GET` preserved, the SAMLResponse lands on the AuthnRequest's own ACS with the right `InResponseTo`, the SAML "Change username" re-POST returns to the username screen, and a tampered verb is rejected. The before-merge items are all in: the rewritten test no longer pins the #325 retarget, quick-fill sits on the username screen with the null guard, the device password screen focuses the password, the helper in `routes/_auth.py` replaced the four copies. Re-verified the rest as well: ruff, mypy, import contracts, 2020 unit tests, CI 9/9, the four surfaces live with the negatives, persona mode inert on all four, Settings UI round trip. Good round.

What is left is small, five one-line or near-one-line fixes, and then this merges:

1. **`two_step_phase` classifies a blank username plus a password as ATTEMPT**, because the `password` short-circuit runs before the username check. Each surface then handles that tampered POST differently: `/saml/sso` re-renders the blank username screen with no error and no audit event, `/login` audits a failed login with `username=''`, `/authorize` says "Username and password are required", `/device` reaches the code store. Test `username_submitted and not username` before the `password` short-circuit (the `not two_step_active` gate stays first, so the combined form is untouched) and all four answer "Username is required"; add that case to `test_two_step_phase.py`, where today no test has a blank username with a password. The `not two_step_login and (...)` guard in `ui.py` then becomes dead and can go.
2. **The Deny button cannot be clicked in a browser on either two-step device screen.** It shares the form with a `required` username / `required` password input and has no `formnovalidate`, so HTML5 validation blocks it until the user types a credential, while the route comment and `test_deny_available_from_username_only_step` promise the opposite (the test posts straight to the route, so it never sees the browser). `formnovalidate` on the Deny buttons, the combined form's one included since it has the same problem.
3. **Device username screen: the `user_code` autofocus guard has the wrong condition.** `not (two_step_login and login_username)` is true on the username screen, so focus lands in the pre-filled code field instead of the username field. `{% if not two_step_login %}` gives the code-first focus to the combined form only.
4. **`/login` reports the same wrong password with two strings**: "Invalid username or password" inline under two-step, "Invalid credentials" on the redirect otherwise. Nothing pins the second on POST, so one string for both.
5. **`render_login` in `saml.py` echoes whatever `saml_original_verb` the form carried** through two screens; the GET/POST whitelist in `_sso_parse_request` only runs at the password step, so a junk value is rejected after the user has typed credentials (on main the re-render reset it to `request.method`). Whitelist once where the closure reads it: upper-case, GET or POST, else the same 400.

Two things I am not asking for here: the cross-tab session interplay of the `/authorize` POST legs (another tab's GET for a different client overwrites the session's `oauth_*` keys, and the pending password POST then mints the code for that client) is the same mechanism as #325 and is documented there now, and the device username step not validating the code before the password is fine as the documented trade-off. Title and body I will rewrite at merge time.

dshvedchenko:
All 2022 tests pass (2020 + 2 new), ruff/mypy clean, import-linter contracts already confirmed clean earlier. Summary of the five before-merge fixes:

_auth.py — two_step_phase now checks blank username before the password short-circuit, so a tampered POST with an empty username + password answers USERNAME_REQUIRED on all four surfaces instead of ATTEMPT. Added the missing test case; updated the now-stale comment in ui.py describing the tampered-two-step path (that path no longer reaches the legacy combined-form guard, which itself stays — it's still live for actual combined-form submissions).

device.html — added formnovalidate to the three Deny buttons (username-step, password-step, combined form) so Deny is clickable without first satisfying the required username/password inputs.

device.html — fixed the user_code autofocus guard to {% if not two_step_login %}, so code-first focus is scoped to the combined form only.

ui.py — unified /login's two "wrong password" strings to "Invalid credentials" on both the two-step render and the redirect; updated the one test that pinned the old string.

saml.py — render_login's closure now whitelists saml_original_verb (uppercase, GET/POST or 400) at the point it's read, rather than trusting the raw form value until _sso_parse_request runs after credentials are typed. Added a regression test.

cdelmonte-zg (approved):
Round 4 on 1833c4d2, approved. Verified: ruff, mypy, import contracts, 2022 unit tests; live with `login.two_step: true`: a blank username plus a password answers "Username is required" on all four surfaces (SAML included, no silent re-render any more), Deny on both device screens carries `formnovalidate` and works without credentials, the device username screen focuses the username and the password screen the password, `/login` reports "Invalid credentials" on both legs, `saml_original_verb=PUT` is rejected at the first screen while lowercase `post` is normalised, and the strict-binding Redirect flow still lands on the SP's ACS with the right `InResponseTo`. Title and body rewritten to the merged shape. Merging once the last CI job reports.

Thanks for four fast, careful rounds.

PR #326, opened 2026-09-07
fix: the /authorize POST leg accepts client_id/redirect_uri from the …

### LINKED ISSUE #325: fix: the /authorize POST leg accepts client_id/redirect_uri from the form body, overriding the request validated on GET
Found while re-verifying #323 (it predates that PR and is not caused by it).

**What happens.** `_AuthorizeParams` in `routes/oauth.py` reads the OAuth parameters from `request.form` on POST, with the session values stored by the GET only as a fallback. So the login POST can carry its own `client_id`, `redirect_uri`, `state`, `scope`, `code_challenge`... and they replace the ones that were validated and shown to the user on the GET.

Reproduced on `main` (a5700339) and on #323's head:

```
GET  /authorize?response_type=code&client_id=plain&redirect_uri=http://localhost:9999/cb&state=st2&scope=openid
POST /authorize   username=admin&password=admin&client_id=evil&redirect_uri=http://evil.example/cb&state=x
-> 302 http://evil.example/cb?code=...&state=x   (audit: success, client: evil)
```

The password is still verified, so this is not an authentication bypass. It is a binding defect: the leg that mints the code is not bound to the authorization request the user saw. The same fallback also means a POST with no OAuth parameters at all works only while the session still holds the GET's keys, which is what the "Change username" link in #323 leans on.

**Expected.** The POST leg authenticates against the request captured by the GET and nothing else. OAuth parameters in the POST body are ignored (or rejected with `invalid_request`); a POST with no captured request answers `invalid_request` as today.

**Compatibility.** Scripted clients already do GET then POST with credentials only (`e2e/test_agent.py` `_run_auth_code_flow`), and no unit test posts OAuth parameters in the body, so binding to the session only is backward compatible for every known caller. A POST to `/authorize?client_id=...` without a prior GET already fails today (the query string is not read on POST), so nothing changes there.

**Scope.** Small: one place (`_AuthorizeParams` construction) plus a regression test that posts a foreign `client_id`/`redirect_uri` and expects the code on the GET's `redirect_uri` (or a 400). Keep the GET behaviour unchanged.

dshvedchenko:
…form body, overriding the request validated on GET

fix #325 

- src/nanoidp/routes/oauth.py: _read_authorize_params now reads client_id, redirect_uri, scope, state, code_challenge/code_challenge_method, nonce, claims, and resource exclusively from the session on the POST leg — never from request.form. GET keeps its existing session-fallback behavior (needed for bare url_for / "Change username" links). login_hint stays request-only per its existing #318 rule.

- Tests: added test_authorize_post_ignores_form_body_oauth_params (forges client_id/redirect_uri/state/scope in the POST body and asserts the code still redirects to the originally-approved URI/state). Fixed two pre-existing native-app tests that were unknowingly exploiting the bug (single POST with no prior GET) to instead do the correct GET-then-POST flow.

- Verified e2e/test_agent.py's pattern of resending auth_params in the POST body alongside credentials remains compatible, since the session already holds identical values from the preceding GET.

- CHANGELOG.md: added a Fixed entry under [Unreleased].

- Full suite: 2023 tests pass, ruff/mypy/import-linter all clean.


dshvedchenko:
Root cause of the CI failure and the fix:

What broke it: My session-only fix for #325 was correct, but three functions in e2e/test_agent.py (test_authorization_code_pkce, test_native_app_redirect_uris, and the persona-mode /authorize check) called bare requests.get/requests.post for the GET-then-POST /authorize pair instead of sharing a requests.Session(). With no cookie jar shared between the two calls, the POST leg never saw the GET's session cookie — previously that "worked" only because the (vulnerable) code let the POST body's client_id/redirect_uri/etc. fill in for the missing session, exactly the bug #325 reports. Once that fallback was removed, these three tests broke because they'd never actually been exercising a real cookie-persisted browser session.

Fix: updated those three call sites to use a local requests.Session() (the pattern already established elsewhere in the file, e.g. test_public_client_flow's flow), and stopped resending the OAuth params in the POST body since they're no longer honored there.

Verified locally end-to-end (mirroring every step of .github/workflows/e2e.yml): main e2e suite 63/63, oauth21 profile 3/3, signed-SAML 3/3, MCP-interop 9/9, MCP stdio smoke test, plus the full unit suite (2023 passed), ruff, and mypy — all clean. Only e2e/test_agent.py has uncommitted changes now; I haven't committed anything per your standing instruction.

cdelmonte-zg (changes requested):
Review round 1 on 5064f212. Verified on a worktree of the head: ruff, mypy and import-linter clean, 2023 unit tests green. Then live, against two instances of this head, one with `login.two_step` off and one on, with two registered clients, `client-a` (callback on :3000) and `client-b` (callback on :4000), plus a third instance on `d5bc352` as the baseline for anything I suspected was pre-existing.

**What is fixed, confirmed live.** The headline of #325 is closed. After a GET for `client-a`, a login POST carrying `client_id`, `redirect_uri`, `state` and `scope` for `client-b` in the body still mints the code for `client-a` with `state=tab-a`. Forged `nonce`, `code_challenge`, `code_challenge_method` and `resource` in the body are ignored too: the ID token carries the GET's nonce, and the code only redeems with the verifier of the GET's challenge, while a forged verifier answers `invalid_grant`. The new regression test is sound, and the e2e rewrite (one cookie jar per flow, no OAuth parameters in the POST body) is the right shape.

Four things to fix before merge.

**1. The PR says `fix #325`, but two of the three scenarios in the issue are still open on this head.** Both reproduced live, with two-step off and on alike.

- *Cross-tab clearing.* Tab A: GET `/authorize?client_id=client-a...`. Tab B, same cookie: GET for `client-b` and a completed login. Tab A now submits valid credentials on its own URL and gets a raw JSON `400 client_id is required`.
- *Cross-tab overwrite.* Tab A: GET for `client-a`. Tab B merely GETs `/authorize?client_id=client-b&redirect_uri=http://localhost:4000/callback&state=tab-b`, without logging in. Tab A submits valid credentials on its own URL and receives `302 http://localhost:4000/callback?code=...&state=tab-b`. The debug log shows the code created for `client-b` and the audit line records `client-b`. The user approved `client-a`; their code went to another client's callback.

With two-step on, the same clearing between step 1 and step 2 turns the password screen into the same 400.

The cause is the one the issue describes: on POST the request is read from a session slot that any other request on the same cookie can clear or replace. Reading only from the session closes the body vector but leaves that slot as the source of truth, so the new docstring's "ties the code to that original GET request unconditionally" overclaims. The login form has no `action`, so every POST already lands on the full `/authorize?...` URL of the page it came from. I confirmed the query string is currently ignored on POST: a POST to `/authorize?<client-b parameters>` after a GET for `client-a` still mints for `client-a`. Reading `request.args` first on the POST leg, with the session as fallback only for a bare URL, and never the body, binds each POST to its own page and closes all three at once. It also brings back the single-POST callers of point 2. The tests are two GETs and one POST on its own query string, plus a variant with a completed login in between.

**2. `existing callers are unaffected` in the CHANGELOG is not true.** A single POST carrying the OAuth parameters and the credentials with no prior GET worked on `main` and now answers `400 client_id is required`, an error naming a parameter the caller did send. That is the shape the PR itself had to rewrite in two shipped unit tests and in the e2e loopback helper, so it is a contract change, not a no-op. It belongs under Changed with the new contract stated: the form carries only credentials, and the request comes from the preceding GET in the same cookie jar. The `authorize()` docstring around line 761 still lists `client_id` and `redirect_uri` as POST parameters, and the endpoints reference says the same, so both need the same sentence. If point 1 is adopted the regression mostly disappears, since a single POST to `/authorize?...` with credentials works again, and the CHANGELOG entry gets simpler.

**3. `test_mismatch_enforced_on_login_post_too` now passes for the wrong reason.** In `tests/test_redirect_uri_matching.py` it single-POSTs every parameter with a mismatched `redirect_uri` and asserts only the 400. I instrumented it on this head: the error is `client_id is required` from the empty session, not the registry rejection the test exists to pin. The POST-leg registry re-validation is no longer covered. The fix is the one you already applied to the two sibling tests in `test_native_app_redirect_uris.py`: GET first with the mismatched URI, then POST the credentials, and assert `not registered` in the description, not just the status.

**4. `test_password_step_does_not_resurrect_a_stale_username` carries a docstring this PR makes false.** In `tests/test_two_step_authorize.py` it still says the form-body retarget is "pre-existing, out-of-scope behavior - tracked separately as #325 - so this test does not pin where the redirect goes". That is exactly what this PR changes. The test already POSTs a forged `client_id`, `redirect_uri` and `state` on the two-step password screen, so asserting that the redirect still goes to `http://localhost:3000/callback` with `state=two-step` is free, and it is the only guard on the two-step second POST: the new regression test never enables two-step.

Smaller points, not blocking.

**5. Forged POSTs are no longer audited.** On `main`, a POST with `client_id=evil-client` and no session was rejected as `Unknown client_id` and produced an audit line naming the client. On this head it stops earlier at `client_id is required` and produces no audit entry at all, so the log loses sight of exactly the attack #325 describes. One line when a POST carries OAuth parameters it should not, or when no pending request exists, restores it.

**6. `login_hint` on POST is now hard-coded to empty.** On `main` it came from the form, so a POSTed `persona-auto-login:` hint triggered auto-login. No caller in the repo uses that shape and the restriction is defensible, but the CHANGELOG does not mention it and the security document describes the hint with no method restriction. One sentence covers it.

**7. The second commit message says it "stopped resending the OAuth params in the POST body", but three POST legs still send them** (`e2e/test_agent.py` lines 734, 989, 1655), and the comment above the second one still claims to mirror "the exact shape `_run_auth_code_flow` sends", which is no longer that shape. They pass because each does a GET first on a shared session. Either strip them or keep them deliberately with a comment saying they are there to prove they are ignored. Also, the new `flow.get` in the loopback helper discards its response with no status check, unlike its sibling, so a regressed GET surfaces as a confusing POST 400.

**8. The `_read_authorize_params` docstring now spends three paragraphs on the history of the bug.** One sentence per leg is enough, the issue number carries the rest. If you keep the session-only branch rather than adopting point 1, the same semantics are one token away from the base code: `params = request.args if request.method == "GET" else ImmutableMultiDict()`, leaving every `.get(key, session.get(...))` to fall through on POST. That shrinks the diff in `oauth.py` from about sixty lines to two and removes the fourth place where the ten `oauth_*` keys are spelled out. Note the identical line also appears in `end_session()` and must not be touched there.

**On the commit messages.** The second commit's subject is a 250-character pasted chat reply ending with a note about standing instructions, and the PR title is GitHub's truncation of a subject that describes the bug rather than the fix. I will rewrite the squash subject at merge, so no need to force-push for that alone, but please keep the next commits to a normal subject line.

**One thing that is not yours.** While probing I found that a fresh `/authorize` request inherits `state`, `nonce`, `code_challenge` and `resource` from an earlier abandoned request in the same browser session, through the per-field session fallback on the GET leg. A stale challenge makes the new client's code unredeemable. I verified it behaves identically on `d5bc352`, so it predates this PR and is not in scope here. Opened as #328. It is a sibling of #325 and the two fixes touch the same function, so if you want to take both in this PR I have no objection, but splitting is fine too.

I will re-run the same probes on the next head.

dshvedchenko:
Thanks for the thorough live probing — all four blocking points and the four smaller ones are addressed.

1. The actual fix, not just the body vector. _read_authorize_params now reads from request.args on both GET and POST — never request.form — falling back to the session only when the current request's own query string is empty. The login form has no action, so a POST always lands back on the exact /authorize?... URL of the page that rendered it; reading that query string first binds each POST to its own request instead of to a session slot any other request on the same cookie can clear or replace. This closes cross-tab clearing and cross-tab overwrite in the same change that closed the body vector. Added two regression tests that reproduce your exact repro steps: test_authorize_post_survives_another_tab_clearing_the_session, test_authorize_post_is_not_hijacked_by_another_tabs_get.

2. CHANGELOG / docstrings corrected. Real (narrower) contract change stated: a single POST /authorize packing OAuth params into the body with no preceding GET no longer works — send those on the query string instead, or GET first as before. Docstrings no longer overclaim "session, unconditionally." Same sentence added to book/src/reference/endpoints.md.

3–4. Both tests fixed to GET-first / assert the right thing, matching the sibling pattern and pinning what each test actually exists to guard.

5–8. Audit line restored for forged POST params, login_hint-on-POST restriction documented + tested, e2e **auth_params spreads stripped and the missing status check added, docstring trimmed.

Verified locally: full unit suite 2026 passed (2023 + 3 new), ruff/mypy/import-linter clean.

cdelmonte-zg (approved):
Round 2 on `8a0057c`: approved. All four blocking points and the four smaller ones from round 1 are closed, and I checked each of them live.

**What I ran:** ruff, mypy and import-linter clean; 2026 unit tests green; four live instances (this head with `two_step` off, with `two_step` on, and in persona mode with `auto_login`, plus `main` at `d5bc352` as the baseline), driven with the round-1 probe scripts.

## Closed

1. **The fix.** Reading `request.args` on both legs closes all three vectors. The baseline still shows all three:

   | probe | `main` d5bc352 | this head |
   |---|---|---|
   | cross-tab clearing (tab B completes, tab A posts on its own URL) | `400 client_id is required` | `302` to `:3000` with `state=tab-a` |
   | cross-tab overwrite (tab B merely GETs) | code for **client-b** at `:4000` | code for client-a at `:3000` |
   | forged `client_id`/`redirect_uri`/`state`/`scope` in the POST body | redirect to `:4000` with `state=evil` | body ignored, code for client-a |

   The same three results hold with `two_step: true`, including a forged body on the password step and a tab-B completion between the two steps. A POST with no query string of its own still resolves from the session, and a wrong password followed by a retry on the same URL still works.

2. **CHANGELOG and docs.** The contract change is stated where a reader will hit it, and the cold `POST /authorize` with body-only parameters returns `400 client_id is required`, matching the text.

3, 4. **Both tests** GET first and assert what they exist to pin. `test_mismatch_enforced_on_login_post_too` reaches the registry check and asserts "not registered".

5. **Audit.** Both branches observed on a running instance: a forged body gives `failed` with `{"fields": ["client_id", "redirect_uri", "scope", "state"], "reason": "OAuth parameters in POST body are ignored"}`; a cold POST gives `{"reason": "POST with no pending /authorize request"}`.

6. **`login_hint`.** On the persona instance, a POST landing on a URL whose query carries `login_hint=persona-auto-login:nonexistent` processes the picker's own selection and issues a code, while the same hint on a GET still produces `error=invalid_request`.

7, 8. **e2e** call sites no longer resend `**auth_params`, the loopback flow gates on the GET's status, and the docstring is shorter.

## One finding, which belongs to #328

The session fallback is per field, so a tab whose own query string omits a parameter takes that parameter from what another tab left in the session:

```bash
# tab A: no resource, no nonce            tab B: both, for a different client
A="response_type=code&client_id=client-a&redirect_uri=http://localhost:3000/callback&scope=openid&state=tab-a"
B="response_type=code&client_id=client-b&redirect_uri=http://localhost:4000/callback&scope=openid&state=tab-b&resource=https://api.example/b&nonce=nonce-b"
curl -sc j -b j "http://127.0.0.1:8326/authorize?$A" -o /dev/null
curl -sc j -b j "http://127.0.0.1:8326/authorize?$B" -o /dev/null
curl -sc j -b j -d username=admin -d password=admin "http://127.0.0.1:8326/authorize?$A" -D -   # code for client-a
# exchanging that code:
#   access_token aud = https://api.example/b      <- tab B's resource
#   id_token nonce   = nonce-b                    <- tab B's nonce
```

Tab A's user approves a request for client-a and receives an access token whose audience is a resource server they never requested, plus an ID token carrying another request's nonce, which a strict client rejects.

This is pre-existing: `main` produces the same two claims in the equivalent old-contract flow, which I checked before writing this. It is the mechanism of #328, so it stays there, and I have posted the reproduction and a tested fix on that issue.

The consequence for this PR was one sentence: "Binding each POST to its own page's query string closes all three at once" holds for the redirect target and reads as covering the whole request. I have narrowed it, and added the same note to the `_read_authorize_params` docstring, in `70d8d7f` on your branch. Nothing needed from you.

## Non-blocking, also going to #328

The store now runs on both legs, so a POST writes the session as well. A failed login in another tab rebinds it, where on `main` only a GET did:

```bash
curl -sc j -b j ".../authorize?$A" -o /dev/null                                      # session = A
curl -sc j -b j -d username=admin -d password=WRONG ".../authorize?$B" -o /dev/null  # failed login in tab B
curl -sc j -b j -d username=admin -d password=admin ".../authorize" -D -             # POST with no query of its own
# this head: code for client-b   |   main: code for client-a
```

This is reachable only from a page whose URL has no query string, which is the #328 window, and the #328 fix covers these lines too.

Merging once CI is green on the branch head. I will write the squash subject, as agreed.

cdelmonte-zg:
@dshvedchenko tnx for the contribution!

Issue #328, opened 2026-09-09
fix: a fresh /authorize request inherits state, nonce, PKCE and resource from an earlier abandoned request in the same session


cdelmonte-zg:
## What happens

`_read_authorize_params` resolves every OAuth parameter on the GET leg as `request.args.get(key, session.get(...))`, then re-stores the result. A fresh `/authorize` request that omits an optional parameter therefore inherits whatever an earlier, unrelated request in the same browser session left behind, and the inherited value is what gets bound to the issued code.

Reproduced live on `d5bc352` (current `main`) with two registered clients, `client-a` (callback on :3000) and `client-b` (callback on :4000):

```
GET /authorize?response_type=code&client_id=client-a&redirect_uri=http://localhost:3000/callback
    &scope=openid&state=S1&nonce=N1&resource=https://api.example.com/v1      # abandoned, user closes the tab
GET /authorize?response_type=code&client_id=client-b&redirect_uri=http://localhost:4000/callback&scope=openid
POST /authorize?...client-b...   (username/password, the plain browser form)
-> 302 http://localhost:4000/callback?code=...&state=S1
```

The code redeems, and the tokens carry the abandoned request's parameters:

| field | client-b sent | token carries |
|---|---|---|
| state | nothing | `S1` |
| id_token nonce | nothing | `N1` |
| access_token aud | nothing | `https://api.example.com/v1` |

With a `code_challenge` on the abandoned request instead, the inherited challenge sticks to the code and `client-b` cannot redeem it at all: `/token` without a verifier answers `invalid_grant`, and the client has no way to produce the verifier because it never generated the challenge.

## Why it matters

- A client that does not send `state` receives one and, if it checks, sees a mismatch it cannot explain.
- A client that does not send `nonce` gets an ID token with someone else's nonce; a strict OIDC client rejects it.
- A stale `code_challenge` makes a perfectly normal flow unredeemable, and the error at `/token` names neither the cause nor the request that caused it.
- The audience of the access token is not what the client asked for.

Scope, verified: the per-client `allowed_resources` check does run on the POST leg against the effective client, so a configured allowlist still rejects the inherited resource with `invalid_target`. The audience leak only reaches clients with no allowlist, which is the default. Everything here is same-browser and same-user, so it is not a cross-user issue.

## Not a regression

Pre-existing, and not introduced by #325 or its fix in #326. Verified identical on `d5bc352`. The fallback predates the two-step work.

## Suggested fix

[...]

PR #329, opened 2026-09-10
fix: isolate authorization request parameters and preserve pending GET state (#328)

dshvedchenko:
Implemented issue #328 across oauth.py:146, test_oauth_flows.py:201, and CHANGELOG.md:23.

Key changes:

Requests with their own client_id are treated as complete and cannot inherit stale optional parameters.
Unrelated query parameters no longer disable whole-request session fallback.
Only GET requests update pending authorization session state.
Failed POSTs in another tab cannot rebind a pending bare submission.
Tests now cover state, nonce, scope, PKCE, claims, resource, same-client retries, unrelated query parameters, and failed cross-tab logins.
Validation:

Full suite: 2,030 passed, with 8 unrelated warnings.
Affected authorization slice: 193 passed.


cdelmonte-zg (changes requested):
Verified on head `02601d3`: ruff, mypy, import contracts, 2030 unit tests, CI 9/9, the e2e agent live (63/63), and a probe matrix on five instances: this head at `:8340`, `main` at `2479bed` as the baseline at `:8341`, a `login.two_step: true` instance at `:8342`, and a `login.mode: persona` + `auto_login` pair, head at `:8343` and main at `:8344`.

The fix is right, and the shape you chose is stricter than the sketch I posted on the issue. Three changes requested, all inside `_read_authorize_params`, plus documentation.

## The defect is closed

Baseline `2479bed`, tab A carrying no `resource` and no `nonce`, tab B carrying both for another client, tab A then logging in on its own URL:

```
main: Location: .../callback?code=...&state=tab-b
      access_token aud = https://api.example/b     id_token nonce = nonce-b
head: Location: .../callback?code=...              (no state)
      access_token aud = my-app                    id_token nonce = absent
```

Same for PKCE: on `main` a `code_challenge` abandoned in another tab makes the exchange fail with `invalid_grant` for a client that never sent one; on this head it returns 200. The same-client retry drops the abandoned `state`/`nonce` too.

## Reading the whole request from one source was the better call

Making the session an all-or-nothing fallback closes more than the issue describes. A request that carries parameters of its own but no `client_id` can no longer inject them into the pending request:

```bash
curl -sc j -b j "$B/authorize?response_type=code&client_id=client-a&redirect_uri=http://localhost:3000/callback&scope=openid&state=tab-a" -o /dev/null
curl -sc j -b j -d username=admin -d password=admin "$B/authorize?state=evil&resource=https://api.example/evil" -D -
```

```
main: Location: .../callback?code=...&state=evil    access_token aud = https://api.example/evil
head: Location: .../callback?code=...&state=tab-a   access_token aud = my-app
```

The per-field shape I sketched on #328 (`params.get("state", fallback.get("oauth_state", ""))`) would have kept `state=evil` and the attacker's `resource`, because the current request's own values still win field by field. Your two-branch version reads nothing from the query string in the fallback branch, so it does not have that hole. Worth keeping as it is.

## Changes requested

**1. The gate tests the value instead of the key, and ignores every other OAuth parameter.** `params.get("client_id")` is falsy both when the key is absent and when it is present but empty, and nothing else on the query string is consulted. Two measured consequences, with a client-a request pending in the session:

```
GET  /authorize?response_type=code&client_id=&redirect_uri=http://localhost:4000/callback&state=evil
  main: 400          head: 200, the login page renders client-a under a URL naming another redirect_uri
POST same URL
  main: 400          head: 302, a code for client-a

POST /authorize?response_type=code&redirect_uri=http://localhost:9999/evil&state=NEW
  head: 302 to http://localhost:3000/callback?code=...&state=tab-a
```

No token reaches the wrong place, because the redirect target comes from the resumed request, so this is not a leak. But a request that says something about the OAuth request and gets answered as a *different* request is worse than `client_id is required`: the client sent `state=NEW` and gets back `state=tab-a` for a `redirect_uri` it never named, with no error anywhere. It also contradicts this PR's own rule, since a query string carrying `client_id=` does carry it.

The gate wants to be "does this query string say anything about the OAuth request", which also keeps your `tracking=abc` test passing:

```python
_AUTHORIZE_REQUEST_FIELDS = (
    "response_type", "client_id", "redirect_uri", "scope", "state",
    "code_challenge", "code_challenge_method", "nonce", "claims", "resource",
)
...
    if any(f in params for f in _AUTHORIZE_REQUEST_FIELDS):
```

**2. An incomplete GET now destroys the pending capture.** The session is still written before validation, and with the request read all-or-nothing that write can no longer be partial in a harmless way: it replaces a complete capture with an incomplete one, and the bare request that was meant to resume it cannot complete any more.

```
GET  /authorize?response_type=code&client_id=client-a&redirect_uri=http://localhost:3000/callback&scope=openid&state=S1
GET  /authorize?client_id=client-a          main: 200    head: 400
POST /authorize   (bare)                    main: 302 ...&state=S1    head: 400
```

On `main` the second GET filled its missing fields from the session and left it coherent. Now any stray or truncated `/authorize` link in the same browser session strands the login in flight. Guarding the write on the fields that make a capture usable is enough:

```python
    if request.method == "GET" and p.client_id and p.redirect_uri and p.response_type:
```

**3. `login_hint` is dropped when the request has no `client_id` of its own.** The fallback branch hard-codes `login_hint=""`, so a request that does carry one loses it. Measured on the persona + `auto_login` pair, with a pending request in the session:

```
GET /authorize?login_hint=persona-auto-login:admin
  main :8344 -> 302 with a code       head :8343 -> 200, the picker
```

This contradicts the field's own comment on `_AuthorizeParams` ("read from the CURRENT request only, never the session fallback the other params use") and `book/src/guides/SECURITY.md:334` ("only ever read from a `GET` request's own query string"). Reading it from `params` in both branches restores the documented rule in one line; the alternative is to state the removal in both places.

### The three together, measured

```diff
-    if params.get("client_id"):
+    if any(f in params for f in _AUTHORIZE_REQUEST_FIELDS):
...
-            login_hint="",
+            login_hint=params.get("login_hint", "") if request.method == "GET" else "",
 
-    if request.method == "GET":
+    if request.method == "GET" and p.client_id and p.redirect_uri and p.response_type:
```

With those three lines: the `#328` probes above keep their results, `?client_id=` and `?state=evil`-style partial requests are 400 on both legs, a partial GET leaves the capture intact so the bare POST still completes with `state=S1`, `login_hint` behaves as on `main` on both shapes, the `tracking=abc` and bare-POST resume paths still work, two-step at `:8342` is unchanged through both screens and "Change username", and the suite stays at 2030 passed with mypy and ruff clean. Tests pinning the empty `client_id` and the partial GET would keep all of it from coming back, next to the four you added.

## Documentation

**4. `book/src/reference/endpoints.md:19` still states the old rule**: the POST reads OAuth params "from the query string only - its own, or the preceding `GET`'s via the session when the POST carries none". After this PR the trigger is not a missing query string: a POST with `?tracking=abc` carries one and still falls back, which your own new test asserts. The reference page is the contract for the endpoint, so it should say what the code says.

**5. Two contract changes the CHANGELOG entry does not mention.** The entry lists the optional parameters, but the branch is taken on `client_id` alone, so the required ones change too:

```
GET /authorize?client_id=client-a   (nothing else, a pending request in the session)
  main: 200, the login page      head: 400
```

and a client that only ever POSTs can no longer retry with a bare submission after a failed attempt:

```
POST /authorize?<full query>   with a wrong password, then POST /authorize with no query string
  main: 302 with a code          head: 400
```

The second is the intended consequence of the GET-only store and browsers are unaffected, since the login form has no `action` and a retry carries the query string. Both deserve a clause, the way the #325 entry documents its own cold-POST change.

## Nits

6. The ten parameter names are now written three times in `_read_authorize_params`, once per branch and once in the session store, where it was two before: adding a parameter means keeping three lists in agreement. A small accessor per branch would leave one constructor without weakening the all-or-nothing semantics.
7. One CHANGELOG line runs past the file's wrap ("its own - and never from the POST body. Previously the POST leg fell back to the").

## Also verified, unchanged

The `#325` probes still hold on this head: OAuth parameters in the POST body are ignored, the code goes to the approved client, and the audit records the forged field names on one branch and "POST with no pending /authorize request" on the other. A cold POST with no session and no query string is 400. Two-step login at `:8342` is intact end to end, including the bare-POST variant of both steps and the negative paths (`Username is required`, `Invalid username or password`). The persona instance stays inert where it should. The remaining "another tab's GET rebinds a bare POST" window is untouched, and outside what this PR claims.

dshvedchenko:
Implemented all feedback from review 1.

Changes include:

Detect any OAuth request field by key presence, including empty values.
Preserve whole-request fallback for unrelated query parameters.
Store pending requests only for complete GETs.
Read login_hint from the current GET even when OAuth fields come from session.
Consolidate parameter extraction into one constructor.
Add regressions for empty client_id, incomplete GETs, and fallback-branch auto-login.
Update the endpoint reference and changelog, including both contract changes and line wrapping.

cdelmonte-zg (changes requested):
Round 2 verified on head `3b37e6b`: ruff, mypy, import contracts, 2033 unit tests, the e2e agent live (63/63), and the full probe matrix from round 1 re-run on four instances, plus a `/code-review high` on this head.

All three changes are in, and you took the one-home nit further than I asked: one constructor with a `value()` accessor instead of two parallel field lists. The three regression tests pin exactly the three behaviours. Measured:

```
?client_id= (empty)                          400 on both legs
?redirect_uri=evil&state=NEW (no client_id)  400
?state=evil on the bare branch               400
GET ?client_id=X alone                       400, and the capture survives:
                                             the bare POST still completes with state=S1
login_hint, full URL and bare GET            both mint a code again, as on main
login_hint on POST                           still ignored
```

Two edge cases I had not probed before also behave: a query string carrying only `?resource=...` or only an empty `?scope=` is treated as its own incomplete request (400) without touching the pending capture. Two-step login is intact through both screens, "Change username" and the bare-POST variant; persona instances behave on both shapes.

What is left is a documentation and consistency pass. No functional change requested.

## Not yours: the capture is still written before validation

The `/code-review` pass flagged that a *rejected* but complete GET still evicts the pending capture: after a GET with an invalid scope, the bare POST answers that rejected request, error and `state=evil` included. I reproduced it on `main` at `2479bed` with the identical result, so it predates this PR, which narrows the family rather than causing it. Filed as #331, with the ordering trap (`_validate_authorize_scope` mutates `p.scope` to the granted value) written down. Nothing to do here.

## Requested

1. **The CHANGELOG's cross-tab sentence claims more than the code does.** "POST requests never [update the capture], so a failed login in another tab cannot rebind a later bare form submission" is true about POST, but in a browser that other tab's failed login is preceded by that tab's own GET, and the GET rebinds the capture:

   ```
   GET  /authorize?...client-a...&state=tab-a    -> captured
   GET  /authorize?...client-b...&state=tab-b    -> recaptured
   POST /authorize (bare)                        -> code for client-b, state=tab-b
   ```

   The protection holds only for a client that POSTs with its own query string and never GETs, which is the shape `test_failed_post_does_not_rebind_pending_request` uses. Narrow the sentence to that case so nobody later reads it as cross-tab isolation. A sibling test covering the GET-preceded tab would make the boundary explicit in the suite too.

2. **`book/src/reference/endpoints.md:20` omits `response_type` from the trigger list**, which is in `_AUTHORIZE_REQUEST_FIELDS`. A request carrying only `response_type=code` does flip to query mode and is rejected, which a reader of that enumeration would not predict.

3. **The `authorize()` docstring was not updated with `_read_authorize_params`.** It still says the query string is read "falling back to the session when absent", while the fallback now also applies when the query string is present and carries no OAuth field, which is the `tracking=abc` case this PR adds on purpose.

4. **State the `login_hint` exception where the contract is written, not only in a docstring aside.** Reading it from the current request in both branches is what I asked for in round 1 and matches `main`, so keep the behaviour. But it does mean a query string carrying only `login_hint` is not "a request in its own right" the way the new rule says every OAuth parameter is, and under `login.mode: persona` with `auto_login` that bare GET completes the pending request as the named persona. One sentence in `_read_authorize_params` and in the reference page, so the exception is visible next to the rule it excepts.

5. **`_FORGEABLE_POST_OAUTH_FIELDS` (9 fields) and the new `_AUTHORIZE_REQUEST_FIELDS` (10) now sit 40 lines apart and disagree**: `response_type` is missing from the first, so a forged `response_type` in a POST body is the one OAuth field `_audit_suspicious_authorize_post` stays quiet about. The divergence predates you, but two near-identical tuples in the same file is a maintenance trap now that the second one exists. Derive one from the other, or say in a comment why they differ. Auditing `response_type` too looks like the simpler end of that.

6. **One e2e pair for the new contract.** The project keeps `e2e/test_agent.py` in step with contract changes, and this one is unit-tested only. The two-step block around `e2e/test_agent.py:1619` already drives bare `POST`/`GET /authorize` in a live session, so it is a natural home for two checks: a bare POST with no preceding GET is refused, and a POST whose query string carries only an unrelated parameter still resumes the pending request.

dshvedchenko:
Narrowed the changelog claim: direct POSTs cannot rebind the capture, but another tab’s GET still can.
Added a unit test explicitly documenting that GET-driven cross-tab boundary.
Added response_type to the endpoint contract.
Updated the authorize() docstring to describe OAuth-field-based fallback.
Documented the GET-only login_hint exception beside the fallback rule.
Derived _FORGEABLE_POST_OAUTH_FIELDS from _AUTHORIZE_REQUEST_FIELDS; response_type is now audited and tested.
Added live e2e checks for:
rejecting a bare POST without a preceding GET;
resuming a pending request when the POST query contains only an unrelated parameter.

cdelmonte-zg (approved):
Round 3 verified on head `ddb53c2`. All six items are done, and the two I cared most about are done the honest way: the CHANGELOG now says a failed *direct* POST cannot rebind the capture and that a GET in another tab still replaces it, and there is a test asserting that GET-driven boundary instead of leaving it implied.

Verified: ruff, mypy, import contracts, 2034 unit tests, CI 9/9 on this head, the e2e agent live (63/63, Two-Step block green with the two new checks), zero em dashes, CHANGELOG under `[Unreleased]`.

Live, on a head instance plus the two-step and persona ones:

```
#328 probes                       aud back to default, no nonce, no state, PKCE not inherited
empty client_id / partial request 400 on both legs
partial GET                       400, and the capture survives: bare POST completes with state=S1
unrelated query parameter         still resumes the pending request
bare POST after its own GET       still completes  (the e2e shape)
cold bare POST                    400
body forgery                      ignored, code still bound to the approved request
two-step                          both screens, "Change username", bare variant, negatives
persona + auto_login              login_hint honoured on the full URL and on a bare GET,
                                  still ignored on POST
```

The `response_type` unification shows up where it should:

```
POST /authorize?<approved request>  body: response_type=token&state=evil
  -> code for the approved request, and the audit records fields ['response_type', 'state']
```

I checked that no template posts an OAuth field in the login form body, so aliasing `_FORGEABLE_POST_OAUTH_FIELDS` to `_AUTHORIZE_REQUEST_FIELDS` cannot produce a false positive on a legitimate submit.

Approved. The two new e2e checks both also pass on `main`, so they are guards on the contract rather than regression tests, which is the right role for them there.

Thanks for four careful rounds on this, and in particular for replacing my sketch with the stricter whole-request shape and then taking the one-home consolidation further than I asked. #331 (the capture is written before validation, so a rejected request still evicts a pending one) stays open as pre-existing work, not yours.

dshvedchenko:
@cdelmonte-zg merge ?

cdelmonte-zg:
Merged as `0cec34c`, closing #328. Verified on `main` after the merge: 2034 unit tests green, the CHANGELOG entry under `[Unreleased]`, and the probe matrix re-run against an instance built from the merge commit, including the two cross-tab scenarios and the empty/partial request cases.

Thanks for four rounds on this one. Two things were yours rather than mine: the whole-request fallback instead of the per-field sketch I put on the issue, which closed an injection path my version would have left open, and the single-constructor consolidation, which went further than the nit asked for. #331 stays open as pre-existing work.

Issue #327, opened 2026-09-08
feat: Provide a helm chart for Kubernetes deployment


gprossliner:
By providing a helm chart, users can deploy and update nanoidp on Kubernetes very easly. 

I'd suggest / provide:

* A helm chart following [Chart Best Practices](https://helm.sh/docs/chart_best_practices/)
* Allows standard helm values, like labels, replicas, ....
* Allows to provide configuration files like `users.yaml` / `settings.yaml` as a whole. The will be emitted as `Secret` object and mounted automatically
* Providing explicit single configuration settings and building the resulting file I would defer to a new feature if needed
* A workflow that pushes a new chart to the gh OCI Registry as `ghcr.io/cdelmonte-zg/nanoidp-chart` (https://helm.sh/docs/topics/registries/)
* The chart is located in `/charts/nanoidp` which is a common convention for in-tree helm charts.

## Versioning

* Use only `Version`, which corresponds 1:1 with nanoidp releases.
* As the chart is in-tree to the actual project, use **only** `Version`, and keep `AppVersion` not set (see https://helm.sh/docs/v2/developing_charts/#charts-and-versioning).  Based on my personal experience, versioning chart and app differently is very confusing for the user. The `AppVersion` doesn't hold any semantics, and is purely informational. So no `helm upgrade` will check this.
* A new OCI helm image is pushed whenever a new release is pushed, even if nothing in the structure of the chart changed. The new chart will have `image.tag` set by default to the actual release. This allows the end user to run `helm install` without explicit version spec to install the latest version, and `helm upgrade` to see and upgrade to a new version if published and not fixed.

## Example installation:

```sh
helm install nanoidp oci://ghcr.io/cdelmonte-zg/nanoidp-chart --values ./values.yaml
```

Please note that this is not a full list of all supported values. I'd take this from best practices and let you review. I only have some examples, and the one that matter / are specific to nanoidp.

```yaml
# values.yaml
#########

# some standard values, see comment above
replicas: 2
ingress:
  create: true

# config: controls the generated config secret
config:
  # users.yaml has a top-level users: array, so I would use this directly here.
  # this only is recommended it users: stay the only root property of users.yaml,
  # otherwise we have to to (the ugly) users.users
  users:
    anton:
      email: 'anton@localhost.test'
      description: Admin user, full access, single site
      roles:
      - ADMIN
    eva:
      email: 'eva@localhost.test'
      description: Editor user, single site
      roles:
      - EDITOR

[...]

PR #330, opened 2026-09-11
feat: helm chart

gprossliner:
see #327 

Call-outs beyond the discussed contract:
* `values.schema.json` alongside `values.yaml`.
* `commonLabels/commonAnnotations`, `service.type/labels/annotations, ingress.labels`.
* Ingress fails fast on `ingress.create: true` with an empty ingress.host, fixed by default to `create: false`
* `Chart.yaml` commits `version: 0.0.0` as an explicit placeholder, not a real version, which is set by arg on public
* CI validates rendered manifests with `kubeconform` plus behavioral `yq` assertions, not just that they render.
* publish workflow in plan doc, it's up to you to actually create this

Details in plan doc, which I'll drop after review


cdelmonte-zg (changes requested):
Thanks, this is a well built chart. The contract from #327 is respected point by point, the README is honest about the three limitations, and `ci/check.sh` asserts behaviour rather than just rendering.

I ran it for real rather than only templating it: a kind cluster (k8s 1.36) with ingress-nginx, the image built from this branch and side-loaded, then six installs (default values, each of the three `ci/` fixtures, an `existingSecret` I created myself, and one with a management secret), plus namespaces enforcing `baseline` and `restricted`. `charts/nanoidp/ci/check.sh` passes locally against helm 3.19 and kubeconform 0.7.0.

Three things to fix before merge, then some smaller ones. Everything below is measured on the cluster, with the commands that produced it.

## 1. `helm upgrade` with a changed `configFiles` is a silent no-op

The README says configuration changes "go through `values.yaml` and a `helm upgrade`, not through the running instance". That path does not currently take effect: the Secret is updated, but nothing rolls the pod, and nanoidp reads its config once at boot.

```
users seen by the running pod:            ['admin']
helm upgrade -f values-with-alice.yaml    (adds a second user)
pod before: t3-nanoidp-56f4f776b9-wwgcg
pod after:  t3-nanoidp-56f4f776b9-wwgcg   <- same pod
users seen by the running pod:            ['admin']
```

The kubelet does refresh the projected Secret, so the file on disk is correct while the process serving requests is not, which is the worst version of this:

```
kubectl exec deploy/t3-nanoidp -- cat /app/config/users.yaml   ->  alice is there
curl /api/users                                                ->  ['admin']
kubectl rollout restart deploy/t3-nanoidp                      ->  ['alice', 'admin']
```

The usual fix applies: a `checksum/config` annotation on the pod template, computed from the rendered Secret, so a config change rolls the pod. With `strategy: Recreate` that is a short gap rather than an overlap, which is the behaviour this chart wants anyway. With `configFiles.existingSecret` the checksum cannot see content the chart does not render, so that case stays the user's responsibility and is worth one README line.

## 2. The "Complete example" publishes an unauthenticated admin surface

`values-full.yaml` and the README's complete example render an Ingress, and nanoidp's `/api/*` and UI are unauthenticated by default. Through the Ingress the chart itself creates:

```
GET  /api/users              200   the user list
POST /api/users/admin/token  200   an access token for admin, no credentials
POST /api/keys/rotate        200   signing keys rotated
POST /api/audit/clear        200   audit log cleared
GET  /                       200   the admin UI
```

The pod says so itself at boot: "the /api/* management endpoints are unauthenticated by design, so any reachable host can mint admin tokens and rotate signing keys ... or set management_secret to require a shared secret for mutations."

The README's Security section recommends keeping the Ingress internal, which is right but is advice about the network, not about the instance. The two switches that exist should be in the one example that puts nanoidp on a hostname, and `values.yaml` already names the first in a comment. Verified in the chart's own shape, with `NANOIDP_MANAGEMENT_SECRET` in `env` and `session.require_ui_login: true` in `configFiles.settings`:

```
POST /api/keys/rotate                   401
POST /api/users/admin/token             401
GET  /                                  302   to the login page
GET  /api/health                        200   the probes keep working
GET  /.well-known/openid-configuration  200
pod                                     1/1 Running
```

So it costs nothing operationally: the gate covers mutations, not the safe methods the probes use. The example already creates a Secret for `CLIENT_SECRET`, so the same `secretKeyRef` pattern covers the management secret with two more lines.

## 3. `helm install ./charts/nanoidp` from a clone cannot start

`image.tag` defaults to `.Chart.Version`, which is the committed `0.0.0` placeholder:

```
helm install t0 charts/nanoidp
Failed to pull image "ghcr.io/cdelmonte-zg/nanoidp:0.0.0": not found
ImagePullBackOff
```

Your reasoning for the placeholder is sound and I am not asking you to commit a real version. But the first thing anyone does with a chart in a repo is install it from the checkout, and `0.0.0: not found` does not say what went wrong. Either a README line ("installing from a git checkout needs `--set image.tag=<release>`, the committed chart version is a publish-time placeholder") or, better, a `NOTES.txt` that prints the resolved image and says this when the version is `0.0.0`. A `fail()` would be the loudest, but it would also make `ci/check.sh` need a `--set`, so your call.

## Smaller

**4. No `ingress.className`.** The rendered Ingress has no class, and it only worked here because kind's bundled controller runs with `--watch-ingress-without-class=true`; the `ADDRESS` column stayed empty throughout. ingress-nginx installed from its own chart defaults to `watchIngressWithoutClass: false`, so on a normal cluster this Ingress is silently ignored unless the installer adds the deprecated `kubernetes.io/ingress.class` annotation by hand. An `ingress.className` value, rendered as `spec.ingressClassName` when set, is the standard shape.

**5. No `securityContext`.** Your `baseline`/`restricted` claim is exactly right, I checked both:

```
namespace enforcing baseline:    1/1 Running
namespace enforcing restricted:  FailedCreate: allowPrivilegeEscalation != false,
                                 unrestricted capabilities, runAsNonRoot != true,
                                 seccompProfile
```

Three of those four are image-independent. Setting them keeps `baseline` working (I patched the running Deployment and it came back 1/1 Running):

```yaml
securityContext:
  allowPrivilegeEscalation: false
  capabilities: { drop: ["ALL"] }
  seccompProfile: { type: RuntimeDefault }
```

and with only those three applied, `restricted` admission narrows to one line: "runAsNonRoot != true". That is the image-side change we agreed is mine, so this would leave the chart one decision away from `restricted` instead of four. Worth adding now, ideally as a value users can extend.

**6. Default values render a Secret with two empty files.** `configFiles.users`/`settings` default to `""`, and the Secret mounts over `/app/config`, which the image ships populated. The result starts and passes its probes while being empty:

```
Users: 0        OAuth Clients: 0        Issuer: http://localhost:8000
```

A bare install therefore looks healthy and can do nothing, and the demo config a `docker run` gives you is shadowed rather than used. Either make the two keys required in `values.schema.json`, or default them to the minimal pair from `ci/values-minimal.yaml`, or say in the README that the mount replaces the image's bundled config. Any of the three is fine, the current silence is what I would rather avoid.

**7.** `nanoidp.labels` still has an `{{- if .Chart.AppVersion }}` branch that can never fire now that the chart deliberately has no `appVersion`.

**8.** No `NOTES.txt`, so `helm install` prints nothing about how to reach the thing it just installed.

## Verified working

For the record, everything else I tried behaved as documented:

- `${INGRESS_URL}` expansion is real: with `values-full.yaml` the boot log shows `Issuer: http://idp.example.com`, and a full authorization code flow through the Ingress returns an ID token with `iss=http://idp.example.com`, `aud=myapp`, `sub=admin`. The README's central claim holds end to end.
- The read-only mount fails exactly as the README says, and visibly: the settings form returns "Failed to update settings: [Errno 30] Read-only file system: '/app/config/.nanoidp-write.lock'" as an error alert, not a silent success.
- Signing keys do rotate on restart: `kid 83406868...` became `27563381...` after deleting the pod, as documented.
- `configFiles.existingSecret` mounts the external Secret and renders no Secret of its own.
- `ingress.create: false` by default, and the `fail()` on an empty `ingress.host` fires.
- Probes hit `/api/health`, which stays ungated even with a management secret set, and `startupProbe` covers first-boot key generation.
- `ci/check.sh` passes locally, including the `values.schema.json` rejection and the `INGRESS_URL` scheme assertions.

## Two notes on my side, not review comments

The plan doc says the release checklist gains a bump of both `Chart.yaml` `version` and the default `image.tag`. Your implementation is better than what we agreed there: since `image.tag` falls back to `.Chart.Version`, and `helm package --version` sets that from the tag, a release needs no file edit at all. I will write the checklist that way.

The publish workflow sketch is a good starting point and I will take it from there, so nothing for you to do on it. And yes, please drop `docs/plans/helm-chart.md` before merge as you suggested.

gprossliner:
I have addressed all the points you mentioned. 

* Verification by static `ci/check.sh` and in a kind cluster. 
* Plan doc reflects the changes. After a final review I'll drop the plan.

cdelmonte-zg (changes requested):
Round 2 verified on head `34ee82e`, the same way as round 1: `ci/check.sh` locally, then a kind cluster with ingress-nginx and the image built from this branch.

All six items are fixed, and each one is now pinned by an assertion in `ci/check.sh` rather than left to a future reader. Measured on the cluster:

```
bare helm install        -> schema error naming /configFiles/users and /configFiles/settings
install from a checkout  -> NOTES.txt: WARNING: the resolved image tag is "0.0.0" ...
ingress.className=nginx  -> CLASS nginx, ADDRESS localhost   (it had no class and no address before)
helm upgrade, new user   -> checksum af4dc6e2 (was 7bf63676), pod rolled, "Users: 2"
values-full via Ingress  -> /api/users/admin/token 401, /api/keys/rotate 401, UI 302,
                            /api/health 200, and the authorization code flow still
                            completes: iss=http://idp.example.com aud=myapp sub=admin
existingSecret           -> installs with no users/settings, no checksum annotation
PSS baseline             -> 1/1 Running
PSS restricted           -> violates PodSecurity "restricted:latest": runAsNonRoot != true
```

That last line is the whole point: `restricted` now fails on exactly one thing, and it is the image-side change I said was mine. I will open an issue for a non-root image, and at that point the chart needs only `runAsNonRoot`/`runAsUser` in the default `securityContext`.

Two things left, both small.

## 1. CI validates the chart with Helm 4 only, and `check.sh` no longer runs on Helm 3

`azure/setup-helm@v4` with no `version:` installs `latest`, which is now **v4.3.0** (from the job log: "Installing v4.3.0"). The new NOTES.txt assertions pass there because Helm 4's `--dry-run=client` renders without contacting a cluster. Helm 3 still does:

```
helm 4.3.0, no kubeconfig:   WARNING: the resolved image tag is "0.0.0" ...
helm 3.19.0, no kubeconfig:  Error: INSTALLATION FAILED: Kubernetes cluster unreachable
```

So `ci/check.sh` fails on Helm 3 at the NOTES step unless a cluster happens to be reachable, which is how I hit it, and its own header says it "runs identically locally and in .github/workflows/helm.yml". Two consequences worth separating: the script's promise is no longer true, and the chart we ship is tested only against the Helm version most users do not have yet.

I would pin `azure/setup-helm` to a 3.x version so CI matches what the README tells people to run, and guard the NOTES step so it skips with a message when neither a cluster nor Helm 4 is available. Testing both in a matrix is also fine by me if you would rather keep 4 covered. Your call on the shape, as long as an unpinned `latest` is not what decides it.

For the record I diffed `helm template` between 3.19.0 and 4.3.0 on `values-full.yaml`: only trailing whitespace differs, and `checksum/config` is identical, so there is no cross-version pod roll hiding in this.

## 2. The complete example does not set `ingress.className`

The new paragraph says to set it, and it is right, but the example right below renders `ingress.create: true` without it, so anyone copying it gets the class-less Ingress the paragraph warns about. One line in the example closes that.

## Then

Drop `docs/plans/helm-chart.md` as you planned and I will merge. Thanks for turning every point into an assertion, that is the part that will keep this working after we both stop looking at it.

gprossliner:
I have addressed the changes. Please note I'm using `nginx` in the sample. I left this out from the default `values.yaml`, as `ingress-nginx` in deprecated, and I didn't want to choose just a random other (we migrated to traefik).

Personally I like it also that all changes are asserted. Nowadays it's easier / more efficient to write test covered code, because not every boilerplate code has to manually written, and everyone of us doesn't have unlimited time. It's just if you care and instruct the machine to do the right things.

The plan doc is deleted as well.

cdelmonte-zg (changes requested):
Round 3 verified on head `9f71e9b`. Both items are done, the matrix is the better answer to the Helm question, and the plan doc is gone with no references left behind.

Verified on the kind cluster: `values-full.yaml` now renders `CLASS nginx` on its own, the pod rolls, and the flow through the Ingress still ends in an ID token with `iss=http://idp.example.com`, `aud=myapp`, `sub=admin`, with `/api/keys/rotate` at 401 and `/api/health` at 200. `ci/check.sh` passes locally on both Helm 3.19.0 and 4.3.0, and CI is green on both matrix legs.

On leaving `ingress.className` out of `values.yaml`: agreed, and for the reason you give. A default that names a controller the cluster may not run is worse than no default, because it looks configured. The README paragraph already says a class-less Ingress is ignored by most controllers, which is the part that matters.

## One thing to fix: `notes_or_skip` cannot fail the script

Its comment says it "prints the failure and exits the whole script, same as any other assertion here". It does not. The call site wraps it in a command substitution, so `exit 1` ends that subshell; the `if` reads the non-zero status as the skip branch, and the script carries on to its success line.

Reproduced by making `helm install` fail for a reason that is not a missing cluster (a wrapper that passes everything to Helm 4 except `install`, which returns a template error):

```
=== NOTES.txt ===
Error: INSTALLATION FAILED: template: nanoidp/templates/NOTES.txt:12:4: ... nil pointer
FAIL: NOTES.txt (default values)
All nanoidp chart CI checks passed.
exit=0
```

So a genuine NOTES.txt breakage prints `FAIL` and leaves the job green, on the Helm 4 leg as much as the Helm 3 one. The skip path you were protecting works; the hard-failure path next to it is the one that got lost.

The smallest fix is to stop calling the function inside `$( )` and hand the output back through a variable, which makes the comment true:

```bash
notes_or_skip() {
  local desc="$1"
  shift
  if NOTES_OUT="$(helm install ci-check "$CHART_DIR" -f "$CI_DIR/values-minimal.yaml" --dry-run=client "$@" 2>&1)"; then
    return 0
  fi
  if [ "$HELM_MAJOR" -lt 4 ] && grep -q 'Kubernetes cluster unreachable' <<<"$NOTES_OUT"; then
    echo "skip: $desc (Helm $HELM_MAJOR's --dry-run=client needs a reachable cluster; none available here)" >&2
    return 1
  fi
  printf '%s\n' "$NOTES_OUT" >&2
  echo "FAIL: $desc" >&2
  exit 1
}

if notes_or_skip "NOTES.txt (default values)"; then
  ...grep "$NOTES_OUT"...
  notes_or_skip "NOTES.txt (image.tag set)" --set image.tag=v3.0.0
  ...grep "$NOTES_OUT"...
fi
```

Measured with that change, in the four shapes that matter:

```
helm 4, install fails for a template error   exit 1, FAIL printed, no success line
helm 4, healthy                              exit 0, both NOTES assertions run
helm 3, no cluster                            exit 0, skip message, rest of the suite runs
helm 3, cluster reachable                     exit 0, both NOTES assertions run
```

Worth a look at the second call site too: as written it sits outside the `if` condition, so `set -e` does abort there, which is the behaviour you wanted in both places.

Nothing else from me on the chart.

gprossliner:
done

cdelmonte-zg (approved):
Round 4 verified on head `8e7b58f`. The fix is the one the round-3 comment described, plus a comment on the function that states the constraint it depends on, which is the part that keeps the next edit from reintroducing this.

`ci/check.sh` re-run in the five shapes that matter, using a `helm` wrapper that passes everything to the real binary except the one subcommand made to fail:

```
helm 4, first install fails for a template error    exit 1, FAIL printed, no success line
helm 4, only the second install (image.tag) fails   exit 1, FAIL printed, no success line
helm 4, healthy                                     exit 0, both NOTES assertions run
helm 3.19, no cluster                               exit 0, skip message, rest of the suite runs
helm 3.19, cluster reachable                        exit 0, both NOTES assertions run
```

The second line is the call site outside the `if`, which now aborts through `set -e` as intended rather than being masked. CI is green on both matrix legs.

Approved. Thank you for the four rounds, and for adding an assertion for each review point. The OCI publish workflow and the non-root image (#332) are on my side.

cdelmonte-zg:
Merged as `268365b4`, closing #327. Thank you for the chart and for the four rounds.

On my side next: the chart-only OCI publish workflow, then the non-root image (#332) and the `runAsNonRoot` default that follows from it.

Issue #331, opened 2026-09-11
fix: a rejected /authorize request still replaces the pending-request capture in the session


cdelmonte-zg:
`/authorize` writes its pending-request capture to the session *before* the request is validated, so a request that is then rejected still replaces whatever was captured before it. The in-flight login that was relying on that capture cannot complete any more, and inherits the rejected request's `state` along with its error.

Reproduced on `main` at `2479bed` and on the head of #329 (identical behaviour, so this is not from that PR, which only narrowed the family by no longer capturing *incomplete* requests):

```bash
A="response_type=code&client_id=client-a&redirect_uri=http://localhost:3000/callback&scope=openid&state=tab-a"
EVIL="response_type=code&client_id=client-a&redirect_uri=http://localhost:3000/callback&scope=not-a-real-scope&state=evil"

curl -sc j -b j "http://127.0.0.1:8340/authorize?$A" -o /dev/null          # captured
curl -sc j -b j "http://127.0.0.1:8340/authorize?$EVIL" -o /dev/null       # 302 invalid_scope
curl -sc j -b j -d username=admin -d password=admin "http://127.0.0.1:8340/authorize" -D -
```

```
Location: .../callback?error=invalid_scope&error_description=...&state=evil
```

The user's pending `state=tab-a` request is gone, and the bare submission answers the rejected request instead. With an unknown `client_id` the same sequence ends in `400 Unknown client_id`.

Nothing is minted and the response goes to a URI registered for the client named in the rejected request, so this is not a token leak. It is a denial of the in-flight authorization, reachable from any page that can make the browser issue a top-level GET to `/authorize`.

## Shape of the fix

The capture exists so that a later request with no OAuth parameters of its own can resume the one in flight. A request that was rejected is not in flight, so it has no business being captured. Write the session copy only once the validation chain has passed, rather than inside `_read_authorize_params` before any of it has run.

The ordering to watch: `_validate_authorize_scope` and `_validate_authorize_resources` replace `p.scope`/`p.resources` with the *granted* values, so a capture written after them stores the granted request rather than the requested one. Decide deliberately which of the two a resumed request should see (storing the granted values re-validates fine, but it is a semantic change worth stating in the docstring).

Tests to add with it: a rejected GET (bad scope, unknown client, unregistered `redirect_uri`) leaves an earlier complete capture intact, and the bare POST that follows still completes the original request.

Found by the round-2 review of #329.

PR #334, opened 2026-09-11
fix: preserve pending authorization after all rejected requests (#331)

dshvedchenko:
Implemented the fix for issue #331.

Session capture now occurs only after the complete /authorize validation chain succeeds. It stores the original requested scope and resources, so resumed requests retain existing validation semantics.

Regression coverage includes:

Invalid scope
Unknown client
Unregistered redirect URI
Unsupported response type
Invalid PKCE method
Invalid resource
Rejected requests without an existing capture
Requested values before normalization
Existing tests updated to POST to their original authorization URL
Updated CHANGELOG.md, oauth.py, and the relevant authorization tests.

Session capture now occurs after persona auto-login handling, so an unknown persona cannot replace an earlier pending request.
The rejected-request regression compares every oauth_* session field.
The regression requires a successful authorization code response and checks the stored code’s client, redirect URI, scope, and state.
Added coverage for unknown-persona rejection followed by a bare POST.


cdelmonte-zg (changes requested):
Verified on head `97963bd`: ruff, mypy, import contracts, 2043 unit tests, CI green, the e2e agent live (63/63), and a probe matrix on four instances (this head, `main` at `268365b4` as the baseline, a `login.two_step` instance, a persona + `auto_login` instance), plus a `/code-review high` pass on the head.

The production change is correct and is the shape the issue asked for: a snapshot of the request before validation, captured only after every validator and after the persona auto-login step, on GET only. Measured, with a valid `state=S1` request pending and then a rejected GET carrying `state=evil`:

```
rejected GET            main: bare POST answers the rejected request     head: bare POST completes S1
invalid scope           error=invalid_scope&state=evil                    code=...&state=S1
unknown client          400                                               code=...&state=S1
unregistered redirect   400                                               code=...&state=S1
unsupported response    error=unsupported_response_type&state=evil        code=...&state=S1
invalid PKCE method     error=invalid_request&state=evil                  code=...&state=S1
```

A rejected GET with nothing pending creates no capture (bare POST 400, where `main` answered it). Under persona + `auto_login`, an unknown hint with `state=evil` leaves the pending `state=tab-a` request in place, and a valid hint still auto-logs in. Under two-step login, a rejected GET between the username and the password screen no longer breaks the bare password submission. The #328/#329 probe matrix and the body-forgery audit are unchanged.

What is left is on the tests and the documentation, not the fix.

## Tests

1. **`tests/test_iss_authorization_response.py:47`: the new early return in `_authorize_redirect` moves `test_invalid_scope_redirects_with_error_and_iss` from the POST leg to the GET leg.** A GET with `scope=definitely-not-a-scope` is itself the 302, so `if response.status_code != 200: return response` returns before `client.post` runs. On `main` that test asserted on the bare POST; now no test covers the POST leg's error redirect carrying `iss`. Dropping the early return and always POSTing to `"/authorize?" + query` keeps the module green and restores the coverage (that POST returns the 302 with `error=invalid_scope`, `state` and `iss`).
2. **`tests/test_oauth_flows.py`, `test_valid_get_captures_requested_values_before_normalization`**: the docstring says "resumed requests revalidate the original scope and resources", but the test issues one GET and inspects the session; nothing is resumed, and the test passes unchanged on `main`. A bare POST after that GET, asserting the issued code has `scope == "openid"` and the single normalized `resource`, would pin what the docstring promises. The `login_hint` lines can go: `auto_login` is off there, and `oauth_login_hint` is a key no code has ever written.
3. **`tests/test_redirect_uri_matching.py:86`**: the rewritten docstring cites #331 for "the rejected GET is not captured for a bare POST", which the body does not check (it POSTs to the full URL). Either assert the bare POST too (400, `client_id is required`; today no unit test pins that outcome after a rejected GET, `test_rejected_get_without_pending_request_creates_no_capture` looks at the session only), or let the docstring describe what the test checks.
4. **One live check in `e2e/test_agent.py`**, next to the two #329 added in the two-step block: a valid GET, a rejected GET with a bad scope, then the bare POST still completing the original request. `e2e/README.md` asks for the scenario in the same PR, and the two sibling fixes on this surface both added theirs.

## Documentation

5. **The #328 entry in the CHANGELOG now contradicts this one**, in the same `[Unreleased]` section: "Only a complete GET updates that capture" and "A GET in another tab still replaces the shared session capture" are no longer true for a complete but rejected GET. "Only a complete, accepted GET" and "A valid GET in another tab" is enough.
6. **`book/src/reference/endpoints.md:24`** has the same sentence ("captured by the preceding `GET`"); one word there too.

## Nits, take or leave

- `_capture_authorize_params`'s completeness guard is dead at its only call site (the validators before it already reject an empty `client_id`, `redirect_uri` or `response_type`). An unconditional helper called under `if request.method == "GET":` in `authorize()`, beside the existing `if request.method == "POST":` audit call, makes "captured iff a GET is about to render the login page" visible in one place, and removes the temptation to move the call earlier because "the guard checks completeness".
- `tests/test_persona_login_flows.py:231` reads `response.headers["Location"]` before asserting the status, so a regression fails with a `KeyError` instead of the status; the parametrized sibling in `test_oauth_flows.py` has the order right.
- The `oauth_*` session snapshot comprehension appears four times and `parse_qs(urlsplit(...).query)` three times across the two test files; a small `oauth_session(client)` helper in `conftest.py` would give the next capture test one place to reuse.

Thank you for taking this one; it was on my list and yours is the better use of the time.

dshvedchenko:
Restored POST-leg invalid_scope, state, and iss coverage.
Extended the normalization test through bare POST and authorization-code inspection.
Added the rejected-GET bare POST assertion to redirect URI tests.
Added the live #331 e2e scenario under two-step login.
Corrected conflicting changelog and endpoint documentation.
Removed the dead capture completeness guard and made the GET-only call explicit.
Fixed assertion order in the persona regression.
Added shared OAuth session and authorization response parsing helpers.

cdelmonte-zg (approved):
Round 2 verified on head `1a7b4d4`: ruff, mypy, import contracts, 2043 unit tests, the e2e agent live (63/63, the new `rejected_get_preserves_pending_request` check inside the two-step block), and the probe matrix re-run on three instances: the five rejection kinds all leave the pending `state=S1` request in place and the bare POST completes it, a rejected GET with nothing pending still yields 400, the #328/#329 matrix is unchanged, two-step survives a rejected GET between its screens, and the persona instance behaves on all four shapes.

All six items and the three nits are in: `_authorize_redirect` POSTs unconditionally again, the normalization test now resumes and inspects the issued code, the bare-POST outcome after a rejected GET is pinned, the two documentation sentences say "accepted"/"valid", the capture call is GET-only at the call site, the assert order is fixed, and the two conftest helpers replace the copies.

Approved. Merging once the last CI legs finish. Thank you.

cdelmonte-zg:
Merged as `6bf5832`, closing #331. Verified on `main` after the merge: the full unit suite green and the rejection probes re-run against the merge commit. Thank you for the two rounds.

PR #337, opened 2026-09-12
fix:  Close #301 single setter for the auth_method session key (two writers, one …

### LINKED ISSUE #301: fix: single setter for the auth_method session key (two writers, one reader)
From the audit: the `auth_method` session key ("persona" vs "password", read by SAML SSO at `routes/saml.py:588` to choose the AuthnContextClassRef) is written identically in TWO places - `routes/ui.py:129` (dashboard login) and `routes/saml.py:485` (SAML inline login). A third login surface that forgets the line would silently declare `PasswordProtectedTransport` for persona logins - a "metadata never lies" hazard with no test on the pairing.

Proposal (3.1, small): one shared `record_session_auth_method(persona_mode)` helper next to the other session helpers, both writers call it, and a test pins that a persona login through EACH surface yields the persona AuthnContextClassRef.

dshvedchenko:
…reader)
Close #301 

What changed

[_auth.py](routes/_auth.py) gains establish_login_session(username, *, persona_mode), which writes user, auth_method, and the permanent flag together, plus the reader session_authenticated_via_persona(). The key literal now lives only in this module.
[ui.py](src/nanoidp/routes/ui.py) and [saml.py](src/nanoidp/routes/saml.py) call the helper instead of assigning the session keys by hand. The SAML success path uses the reader for its AuthnContextClassRef choice.
I went one step past the issue's literal proposal. A helper that only writes auth_method could still be skipped by a third surface. Making the helper the sole writer of session["user"] removes that failure mode structurally.
Tests

New [test_login_session_contract.py](tests/test_login_session_contract.py) pins what the helper writes and greps src/nanoidp for any direct access to either key outside _auth.py. I confirmed it fails when a stray session["user"] = ... is added to ui.py, then restored the file.
Persona logins on both surfaces now also assert the session key itself, not just the assertion XML, in [test_persona_login.py](tests/test_persona_login.py) and [test_persona_login_flows.py](tests/test_persona_login_flows.py).


cdelmonte-zg (commented):
Thank you, this is a clean fix, and I agree with the step past the issue's literal proposal: making the helper the sole writer of `session['user']` as well is what removes the failure mode; a helper for `auth_method` alone would still be skippable.

What I verified on 79949c9 (merge-base is current main):

- Full unit suite: 2048 passed. The three changed test files: 77 passed. `ruff`, `mypy src`, `lint-imports` clean.
- The tripwire does what its docstring says. A stray `session["user"] = ...` in ui.py fails it with file and line; so does a stray `session.pop("auth_method", None)`. A key held in a variable passes, as the docstring admits. The file was restored after each probe.
- Census of every session access under `src/nanoidp` on the PR head: `auth_method` is written in `_auth.py` only and read through `session_authenticated_via_persona` only; `session['user']` is written in `_auth.py` only. The other session writers (oauth.py's `oauth_*` capture, saml.py's `saml_verified_redirect`, `management_verified`) are different keys, and both logouts clear the whole session.
- Live, two servers started from the branch (default config, and the persona preset): dashboard login then SAML SSO reusing that session gives `PasswordProtectedTransport` in password mode and `unspecified` in persona mode; the inline SAML login gives the same pair; after logout the SSO shows the login form again; an unknown user through the inline login and a wrong dashboard password issue no assertion and no session. No error in the debug logs. The default e2e agent suite against the password-mode server passes.

One small round, all in the test and in prose; no runtime change:

1. **The regex fires on reads.** Verified against `DIRECT_SESSION_ACCESS` as written: `if not session["user"]: return redirect(url_for("ui.login", error=e))` matches (the annotated-assignment branch `(?::[^=]*)?=` takes the `:` of the `if` as an annotation and runs to the keyword's `=`), and so do `if session["user"]: x = 1`, `session.update(actor=session.get("user"))` and `session.setdefault("hint", user.username)` (via `[^)]*\buser\b`). The next person adding a legitimate read like that gets a red contract test. Tightening the annotation branch to a type (`(?::\s*[\w\[\], .|]+)?=`) and requiring `\buser\s*=` or a quoted `"user"` inside `update`/`setdefault` fixes the four; adding them as negatives to `test_the_guard_pattern_matches_what_it_claims_to` keeps it that way. If you would rather go the way `tests/test_token_issuance_parity.py` does (an `ast.walk` over the same `rglob`, flagging a `Subscript` store on `session` with the `user` constant and any `auth_method` constant outside `_auth.py`), that also removes the comment/docstring false positive and catches the multi-line `session.update(` form the docstring lists as missed. Either is fine with me; the regex fix is the smaller change.
2. **`_AUTH_METHOD_KEY`** is the one spelling of the key the regex cannot see (`from ._auth import _AUTH_METHOD_KEY; session[_AUTH_METHOD_KEY] = ...` matches nothing), and the literal is still spelled in the tests' assertions, so the constant buys no rename safety either. Either inline the literal in the helper and the reader, or add the constant's name as a fourth regex alternative; the comment above it then needs to say which.
3. **Helper docstring:** "Every interactive login surface" reads as four in this repo (`config.interactive_authenticate` names `/login`, `/authorize`, `/saml/sso`, `/device`), and `/authorize` and `/device` authenticate through it without establishing a UI session and never call this helper. "The two surfaces that establish a UI session (ui.login's POST handler, the SAML SSO inline-login form); a future one that does must go through this helper" is what is true.
4. **CHANGELOG:** the entry ends with "No behavior changes for existing logins", and every such consolidation in this file sits under `Changed` (the `/userinfo` one-home entry in 2.4.0, the #229 phases); `Fixed` is where a 3.1 reader looks for regressions that reached them. Under `Changed`, please, and "can therefore no longer record a user without recording the method" softened to what the test guarantees ("is held to the helper by a contract test, direct forms only").
5. **`book/src/project/architecture.md`, line 67:** the `routes/_auth.py` row still describes the module as the management-gate choke point only. One clause naming `establish_login_session` as the single writer of the login session is where a contributor adding a login surface would find it.

I will add `Closes #301` and retype the title and body when merging: the PR does cover the issue completely (the helper, both writers, and the per-surface AuthnContextClassRef tests already on main, `test_prior_persona_dashboard_login_reused_gets_unspecified_context` and `test_persona_mode_selection_completes_sso_with_unspecified_context`, now with the session-key assertions next to them), and the current title is cut with an ellipsis whose tail became the body's first line.

Two things I noted that are not for this PR and that I may open as follow-ups: the absent-key default in `session_authenticated_via_persona` still resolves to the stronger claim (`PasswordProtectedTransport`), a choice that predates this PR (#158); and the caller, not the authenticator, is the one telling the helper how the login authenticated (`persona_mode=`), so a copy-pasted literal is the one mistake the tripwire cannot see.

dshvedchenko:
All five review points are applied, verified, and left uncommitted on iss_301 for you to review.

What changed

Tripwire rewritten as an AST walk in [test_login_session_contract.py](tests/test_login_session_contract.py), taking the reviewer's second option. It flags, outside _auth.py: a store or del on session["user"], a session.update/setdefault naming user as a keyword, dict-literal key, or first positional, and the string constant "auth_method" anywhere. Reads like if not session["user"]: and actor=session.get("user") now pass, comments and docstrings are never seen, and the multi-line session.update( form is caught. The four false positives from the review are negatives in the self-check test, alongside new positives for del, flask.session[...], .pop, and the split update call.
_AUTH_METHOD_KEY removed from [_auth.py](src/nanoidp/routes/_auth.py). The literal is inlined in the writer and the reader, and the helper docstring says why a named alias is deliberately avoided.
Helper docstring now names the two surfaces that establish a UI session, says a future one must go through the helper, and notes that /authorize and /device authenticate without establishing a session.
CHANGELOG entry moved to a new ### Changed section under Unreleased, with the guarantee softened to what the test enforces (direct forms only, listed).
Architecture doc row for routes/_auth.py names establish_login_session as the single writer of the login session.
Verification

Full unit suite: 2048 passed. The three touched test files: 77 passed. ruff, mypy src, lint-imports clean.
Live probes against ui.py: a stray session["user"] = ..., session.pop("auth_method", None), and a multi-line session.update({...}) each fail with file and line. if not session["user"]: and session.get("user") pass. File restored after each.
The markdownlint warnings the IDE surfaced on CHANGELOG.md are the file's existing heading style, not new.

What changed in [test_login_session_contract.py](tests/test_login_session_contract.py)

Unpacked dict keywords are inspected. A keyword with no name whose value is a dict literal naming user now flags, so session.update(**{"user": username}) is caught. Added as a positive case.
setdefault inspects only its key argument. Dict-key inspection is limited to update. session.setdefault("hint", {"user": username}) no longer flags. Added as a negative case.
Docstrings are excluded. A pre-pass collects the docstring constant of every module, class, and function, and the auth_method constant check skips those nodes. Added exact-key negatives for a function docstring, a class docstring, and a module docstring.
Module docstring updated to describe the **{...} form and the docstring exclusion.
Architecture row in [architecture.md](book/src/project/architecture.md#L67) now says "every surface that establishes a UI session calls it", matching the helper docstring and excluding /authorize and /device.

Verification

Full unit suite: 2048 passed. Contract test plus the two persona test files: 77 passed.
ruff check ., mypy src, and lint-imports all clean.
The uv.lock file that uv run generates as a side effect was removed again; the working tree holds only the four intended files.

cdelmonte-zg (commented):
Round 2 on a2b8aad: the five points are addressed, and the AST rewrite is the better of the two options. Verified:

- Full suite 2048 passed; `ruff check .`, `mypy src` clean.
- The tripwire, probed by editing ui.py and restoring it each time: a stray `session["user"] = ...`, `session.pop("auth_method", None)`, a multi-line `session.update(`, `del session["user"]`, `flask.session["user"] = ...` and `session.update(**{"user": ...})` each fail the contract test with file and line. The four reads that tripped the round 1 regex (`if not session["user"]: return redirect(url_for(..., error=e))`, `if session["user"]:`, `session.update(actor=session.get("user"))`, `session.setdefault("hint", ...)`) and a plain `name = session["user"]` pass. The two documented misses (key in a variable, dict built elsewhere) still pass, as documented.
- `_AUTH_METHOD_KEY` gone, literal in the writer and the reader with the reason in the docstring; the helper docstring names the two surfaces and says what `/authorize` and `/device` do; the architecture row names the single writer.

One thing to fix before merge, and it is not your doing: `main` moved under the PR. #342 added a `### Changed` section to `[Unreleased]` at the same spot yours goes, and a merge of the current `main` into this branch applies cleanly but leaves two `### Changed` headings back to back. Either rebase on `main` and fold your entry into the existing section (after the Helm chart entry), or say so and I will make that one edit on your branch when merging; the entry's text stays as it is.

Three small things on the tripwire itself, all verified by running `login_session_key_offenders` on the snippet; the first two are worth a few lines each, the third is wording:

1. `session.update(dict(user=username))` passes: `_update_call_writes_user` inspects `ast.Dict` only, and `dict(...)` is a Call. The CHANGELOG entry ("`session.update` ... naming the key") and the docstring read as if this form were covered. Either match a `dict(...)` call with a `user` keyword as well, or list it with the other misses.
2. `del session["user"]` is flagged but `session.pop("user", None)` and `session.pop("user")` are not: the two spellings of the same operation get opposite verdicts. Match `pop` with the `user` key in the Call branch next to `setdefault`, or drop the `Del` case and its self-check line.
3. The docstring's "what still passes" list could name the aliased import (`from flask import session as s; s["user"] = ...`, `_is_session` sees only `session` and `<x>.session`), and "exactly one reader" holds within `src/nanoidp` (the test files read `sess["auth_method"]` directly, which is fine, so say "in src").

Two observations, no change asked: any bare `"auth_method"` constant trips the test, so an audit `details={"auth_method": ...}` next to the login call would go red on code that touches no session key; the message says what to do, and I would rather keep the strong check than weaken it, but the docstring could say that an unrelated use of the literal is expected to be renamed. And `_is_session` matches any `.session` attribute (`db.session.update(user=1)` flags); nothing in `src/nanoidp` has one today, narrowing it to `flask.session` is optional.

With the heading folded and 1 and 2 in, this is ready; I will not ask for another round on the tripwire's reach beyond that, it is a tripwire and says so.

dshvedchenko:
Round 2 is addressed. The branch is rebased onto main and the edits are in the working tree, uncommitted, per your standing rule.

Rebase and CHANGELOG. iss_301 now sits on main at 3.1.0rc2. The rebase applied cleanly, and I removed the duplicate ### Changed heading so the #301 entry follows the Helm chart entry in the existing section. The entry's text is unchanged.

Tripwire fixes in test_login_session_contract.py:

dict(...) calls now match. _dict_names_user accepts a dict(user=...) or dict({"user": ...}) call alongside a dict literal, so session.update(dict(user=username)) and session.update(**dict(user=username)) are flagged. The docstring and the CHANGELOG claim about session.update are now true.
pop joins del. The Call branch treats pop like setdefault and checks the key argument, so session.pop("user", None) and session.pop("user") get the same verdict as del session["user"]. session.pop("hint", None) still passes.
Docstring wording. The module docstring now names the aliased import as a miss, says why <anything>.session is matched wide and that nothing in src has such an attribute today, says an unrelated "auth_method" literal is expected to pick another name, and scopes "exactly one reader" to src/nanoidp with the tests reading the key directly.
Self-checks. New positive lines for the three covered forms, negative lines for the non-user pop and dict(...) cases, and the three documented misses pinned as passing so a change in reach shows up in this test rather than in review.

cdelmonte-zg (approved):
Round 3 on 6bd4249, verified: rebased on the current `main` (merge-base 9bc964e), one `### Changed` section with the #301 entry after the Helm chart one; full suite 2048 passed, `ruff`, `mypy src` clean; the tripwire on sixteen snippets gives the expected verdict on each: `session.update(dict(user=...))`, `session.update(**dict(user=...))`, `session.update(dict({"user": ...}))`, `session.pop("user", None)` and `session.pop("user")` are now caught alongside `del`; `session.pop("hint", None)`, the four reads from round 1 and a plain `name = session["user"]` pass; the two documented misses pass as documented, now pinned in the self-check.

Approving and merging, with `Closes #301` and the title retyped (GitHub had cut it with an ellipsis). Thank you for the three rounds; the AST tripwire is a better piece than the issue asked for.

cdelmonte-zg:
@dshvedchenko, `v3.1.0-rc3` is out and carries all of your work from this cycle: the two-step login (#323), the three `/authorize` corrections (#326, #329, #334) and this single writer (#337). Release notes: https://github.com/cdelmonte-zg/nanoidp/releases/tag/v3.1.0-rc3

```
pip install --pre nanoidp==3.1.0rc3
docker pull ghcr.io/cdelmonte-zg/nanoidp:v3.1.0-rc3
```

Plain `pip install` and `:latest` stay on 3.0.0 until the final. What deserves a real run before that, and where you are the best placed to try: your PoC demo flows with `login.two_step: true`, a client that used to drive `/authorize` with a single POST carrying the OAuth parameters in the body (that form is no longer honoured, the parameters go on the query string, see the CHANGELOG's Migration notes), and, if you run the image with a keys volume created by an earlier release, the one-time chown the notes describe. If it soaks clean, 3.1.0 final follows.

cdelmonte-zg:
@dshvedchenko, 3.1.0 final is out with all of your work from this cycle: the two-step login (#323), the three `/authorize` corrections (#326, #329, #334) and the single login-session writer (#337). It is the rc3 tree unchanged, so the pinned rc3 commands in my previous comment still install the pre-release; for the final:

```
pip install nanoidp==3.1.0
docker pull ghcr.io/cdelmonte-zg/nanoidp:v3.1.0   # or :latest
```

Plain `pip install nanoidp` and `:latest` now resolve 3.1.0. Release notes, with the credits: https://github.com/cdelmonte-zg/nanoidp/releases/tag/v3.1.0

With the final out, #350 can land after its next round, with the CHANGELOG entry under the new `[Unreleased]` as agreed.

Issue #194, opened 2026-08-23
n8n end-to-end: OIDC login and AI Agent -> MCP Client Tool -> mock MCP server -> nanoidp


cdelmonte-zg:
**Why.** A real agent host with real expectations about discovery, claims, PKCE, DCR and MCP authorization is the only honest test of this milestone. Separate workflow: an n8n container takes 60-90 s to start.

**Scope.**
- Workflow `n8n-e2e.yml` (manual + nightly, not on every PR): nanoidp, the mock MCP server and n8n (pinned image) on one Compose network; n8n configured through its REST API/CLI with a workflow that uses the MCP Client Tool with OAuth2 against nanoidp (with and without DCR) and a second one using generic OAuth2 credentials; assertions through n8n's execution API.
- **No real LLM in CI.** Two n8n paths, different purposes:
  1. **Foundation (success criterion)**: n8n **MCP Client (step) node** -> mock MCP server -> nanoidp. Deterministic, tests OAuth/MCP interoperability from a real host.
  2. **Agentic integration (additional)**: n8n **AI Agent + MCP Client Tool**, with a deliberately stupid OpenAI-compatible mock chat model as a fixture (`examples/mock_chat_model.py`): the minimum the pinned n8n version needs, a two-state machine (request with tools -> one `tool_call(read_document)`; request with the tool result -> a fixed final answer). Tests that n8n propagates the authenticated MCP tool into the agent's tool-calling loop, not OAuth itself. Not a general OpenAI emulation; it must stay a fixture.
- **`examples/agentic-stack/docker-compose.yml`**: nanoidp + mock MCP server + n8n (+ the mock chat model) so that `docker compose up` lets anyone open n8n, run the workflow and watch discovery -> token (`aud`/`resource`/`scope`) -> `tools/call` end to end. For hands-on exploration with a real model the compose offers an optional `ollama` profile (`docker compose --profile ollama up`) wired as the AI Agent's chat model, with a small local model that supports tool calling pinned in the README; no hosted LLM provider is configured anywhere in the repo, so the stack costs nothing to run. CI never uses the Ollama profile: the deterministic mock chat model remains the only model in automated tests. Architectural rule: the compose file assembles components that are already testable on their own and contains no logic the tests need; #191 runs without Docker or n8n, and #194 runs the same mock MCP server, never a variant of it.
- Two deliverables, deliberately separate:
[...]

Issue #190, opened 2026-08-23
Dynamic Client Registration (RFC 7591/7592), opt-in and open, backed by the runtime store


cdelmonte-zg:
*Rewritten 2026-09-16 against the code as it stands after #235 and #192 (3b10d2d). Three premises of the previous text had been overtaken by that work, and `save_to_config` turned out to duplicate a capability #192 already ships. The earlier version is in the edit history.*

**Why.** n8n's `MCP OAuth2 API` credential has "Use Dynamic Client Registration" on by default: the host is given the MCP URL and registers itself. nanoidp has no `/register`, so that path stops. MCP 2026-07-28 makes CIMD the SHOULD and keeps DCR a MAY for backward compatibility, so DCR is implemented for today's ecosystem while the client-resolution seam stays ready for CIMD (#196).

The observed sequence, documented in `book/src/guides/n8n-end-to-end.md`: n8n discovers nanoidp from the resource metadata, asks for `/.well-known/oauth-authorization-server`, **falls back** to `openid-configuration` when that is a 404, and stops at the missing `registration_endpoint`. The 404 is therefore not what blocks it; the missing endpoint is. Both are fixed here, in that order.

## What the codebase already provides

Three things the previous text asked for exist today, and asking for them again would add a second home for a rule that already has one.

- **The client registry is `IdentityResolver`.** `resolve_client` composes the declared configuration with the runtime store and returns `ResolvedClient(client, origin)` with `Origin = Literal["declared", "runtime"]` (`services/identities.py:41,50,154`). Every protocol path already resolves through it: `/authorize`, `/token`, `/introspect`, `/revoke`, `/device_authorization`, `services/token.py`. **No `ClientRegistry` class is added.** #196 extends the same `Origin` with `"cimd"`.
- **There is no `/api/clients`, and none is invented for DCR.** Runtime clients are already listed with their origin by `GET /api/runtime/clients` (#192), so a registered client appears there by construction.
- **MCP stays declared-only.** #192 settled this: the MCP server is a separate process with its own `ConfigManager` and its own in-memory store, so routing `list_clients` through the resolver would show that process's empty store, not the clients registered against the running IdP. The MCP-visible change here is the one that comes through the shared builder: `get_oidc_discovery` gains `registration_endpoint`, because HTTP and MCP both call `services/discovery.build_discovery_document`.

## PR A - RFC 8414 authorization server metadata

[...]

PR #338, opened 2026-09-12
feat: n8n end to end, path 2 of #194 (AI Agent -> MCP Client Tool -> mock MCP server -> nanoidp)

cdelmonte-zg:
Path 2 of #194: n8n's **AI Agent** with the **MCP Client Tool** node inside its tool-calling loop, against the same nanoidp + mock MCP server stack as path 1 (#336), with a mock chat model as the agent's model so that CI needs no LLM. Path 1 tested the OAuth from a real MCP host; path 2 tests that n8n propagates the OAuth-protected tools into the agent and that the resource server's authorization decisions come back through the loop. No change to nanoidp.

## What is in

- `e2e/mock_chat_model.py`: an OpenAI-compatible fixture, standard library only, deliberately stupid. `GET /v1/models` and `POST /v1/chat/completions`; offered tools, it always answers with one tool call (`read_document` unless the prompt says "delete" or "admin"); handed a tool result, it answers with a fixed sentence quoting it. Every request is kept and served on `GET /requests` so the script can assert on what n8n sent. `stream: true` is answered with 400 on purpose: the gap is visible, not silent. Not a general OpenAI emulation, and meant to stay that way.
- `examples/agentic-stack/docker-compose.yml`: a `chat` service (nanoidp image + the mounted fixture, like the mock MCP server), n8n now depends on it; and an optional `ollama` profile (`ollama/ollama:0.34.0`, a volume for the model) for running the same loop by hand with a real local model. CI never uses the profile.
- `examples/agentic-stack/n8n-import/credentials.json`: the `openAiApi` credential pointing at the mock (`url: http://chat:9200/v1`), imported with the other two.
- `e2e/n8n_e2e.py`: `create_agent_workflow` (Webhook -> AI Agent 3.1 with `enableStreaming` off and `maxIterations` 3; OpenAI Chat Model 1.3 with the Responses API off; MCP Client Tool 1.4, all tools, the path 1 credential) and four checks after the path 1 ones; the workflow-creation tail is shared between the two builders.
- Docs: the guide gains "Path 2: the AI Agent's tool-calling loop" and "A real model, by hand"; the stack README, `examples/README.md`, `e2e/README.md`, the workflow header and the CHANGELOG follow.

## What the four new checks assert

| check | what it shows |
| --- | --- |
| the model is offered `MCP_Client_Tool_read_document`, `..._delete_document`, `..._admin_operation` | n8n propagates the OAuth-protected server's tool list into the agent; listing is not gated on scope, calling is |
| asked to read `doc-42`, the model calls `read_document`, the tool result carries the document, the agent's answer quotes it, execution `success` | the token n8n obtained does the same job inside the loop as from the step node |
| asked to delete, the tool result is the server's `insufficient_scope ... requires the 'documents:write' scope`, quoted in the answer; execution `success`, the tool node's run marked as error | the resource server's decision reaches the loop as a tool result the model can react to, not as a crash |
| with the wrong-audience credential the execution fails on `tools/list` (`Authentication failed`, `invalid_token` in the tool node's error) and the mock's request log is empty | a token for another resource is refused before the agent knows what tools exist |

## Facts measured on n8n 2.38.7 that the fixture and the script encode

- The MCP Client Tool node offers every tool under the node's sanitized name as a prefix (`MCP_Client_Tool_read_document`) and maps the call back; the mock matches on the suffix.
- The OpenAI Chat Model node defaults to the Responses API in 1.3; the workflow turns it off. The request the mock receives is `model`, `messages`, `tools` (each `strict: false`) and `stream: false`; no `tool_choice`, no `response_format`. The credential's Base URL is honoured without an allowlist.
- A tool result reaches the model as a JSON string of the MCP `content` list; an `isError` result as `[{"error": "..."}]` with the server's text, and the run continues. A connection failure at `tools/list` is fatal and carries n8n's own wording at execution level, with the 401 body in the tool node's error.

## Verification

- Fresh stack (`down -v`, `up --build --wait`) then `python e2e/n8n_e2e.py`: 11/11 (7 from path 1 unchanged, 4 new). Re-runs against the same stack: 11/11.
- The checks are not vacuous: with the mock's request log unreachable (`--chat` pointing at a closed port) the run exits 1 at the first path 2 check. The first run against n8n failed the wrong-audience check on wording: the execution-level error is n8n's own sentence and the 401 body is in the tool node's error, so the check now reads both.
- `ruff check e2e/` clean; `mypy` clean on the fixture.
- The `ollama` profile, by hand: `--profile ollama up`, `ollama pull qwen2.5:3b`, then the agent workflow with an Ollama Chat Model node (credential base URL `http://ollama:11434`) instead of the mock: the model called `MCP_Client_Tool_read_document` with `document_id: doc-42` and answered with the document in 6 s on CPU. The service publishes no port: n8n reaches it on the Compose network, and a host that already runs its own Ollama keeps its 11434.

## Not in this PR

- Path 2b of #194, the OIDC login to n8n itself (Enterprise, docs only). #194 stays open for it.
- Dynamic client registration (#190), unchanged from path 1.


cdelmonte-zg:
Both points addressed in 4a84166.

- The insufficient_scope check now asserts the whole documented property: the refusal is fed back to the model and quoted in the answer, the execution status is `success`, and the MCP Client Tool node's run carries the refusal as its error (`node_error(execution, "MCP Client Tool")`). The script's docstring for check 7 says the same.
- The mock chat model builds the arguments from the tool it calls: `document_id` for read_document and delete_document, `action` for admin_operation, matching the mock MCP server's signatures.

Verified: full run 11/11 on the running stack. The admin path, which no check exercises, run once by hand through the agent with the prompt "Run the admin audit.": the model called `MCP_Client_Tool_admin_operation` with `{"action": "audit"}`, the server refused it with `insufficient_scope: 'admin_operation' requires the 'admin' scope`, the execution succeeded and the tool node's run carries that error, the same shape as the delete case.

cdelmonte-zg:
One more commit, 1480968, for the nightly's red run (34681960004 on `a960c20`, the path 1 code), since the fix is in the files this PR owns.

**Cause, reproduced on a fresh n8n 2.38.7 container.** n8n starts in three phases: at about 2 s `/healthz` is 200 JSON while every other path, `/rest/settings` included, gets a 200 `text/html` "n8n is starting up. Please wait" page; at about 5.5 s the SPA answers 404 HTML while the REST routes are still unmounted; at about 6.6 s the JSON arrives. The Compose healthcheck (`wget -qO-`, any 200) and the script's `wait_http` both accepted the HTML page. In the failed run `--wait` returned 5.5 s after the n8n container started, the script ran its whole bootstrap in 200 ms against the HTML page, and failed decoding JSON on `/rest/api-keys/scopes`; the stack logs show n8n still running migrations at that moment. Locally n8n never happened to be probed in that window.

**Fix.** Readiness is `/rest/settings` answering the JSON envelope: the healthcheck is now `wget -qO- ... | grep -q '"data"'`, `wait_http` takes a `json_key` and waits for it, and the scopes call checks its status before decoding. The fact is recorded in the guide next to the other n8n facts.

**Verified.** On a fresh container the old rule returned at 2.0 s with the HTML page, the new rule at 8.2 s with the JSON; the new healthcheck command passes inside the container. Fresh stack (n8n data volume removed) with the new healthcheck: 11/11. Whether the nightly stays green is what the next scheduled runs will show; the mechanism it failed on is closed.

Issue #348, opened 2026-09-12
feat: declarative per-user TOTP second factor for demoing an MFA login (secret in users.yaml, no enrolment, no secret management)


dshvedchenko:
> **Scope as agreed (maintainer, 2026-09-12):** the declarative form described in [this comment](https://github.com/cdelmonte-zg/nanoidp/issues/348#issuecomment-5646374298) replaces the original request below; the nine-step production plan is not the scope.

**Title:** Add optional per-user TOTP authentication

## Summary

Add Time-based One-Time Password (TOTP) as an optional authentication factor that users can enable individually. TOTP should complement the existing multi-step authentication flow.

dshvedchenko:
need open discussion 
new description 
---

# feat: declarative per-user TOTP second factor for demoing an MFA login

## Why

nanoidp is used to prototype a client's login experience (VISION, "two secondary, supported uses"). A client that has to handle an IdP with MFA cannot exercise that today: there is no second screen after the password, no login that fails on a wrong code, and no `amr` claim to tell the two logins apart. This adds the visible flow and the claims that follow from it, as data in the configuration files, so a client can be demoed with a known secret and a real authenticator app.

This is not IdP security. Nothing here protects anything: the secret sits next to a plaintext password in a disposable instance, and the non-goals (no hardening guarantees, personas rather than people) stay as they are. The scope below is the one agreed in [this comment](https://github.com/cdelmonte-zg/nanoidp/issues/348#issuecomment-5646374298); the original production-style request is withdrawn.

## Scope: a declarative second factor

Everything is data in the configuration files, nothing is managed at runtime, like every other property of a user.

### Configuration

- `login.totp: true|false`, global, default `false`, next to `login.mode`, `login.auto_login` and `login.two_step`. Same treatment as those: YAML, the Settings UI, the MCP settings tools, `config_documents` and the generated schema.
- Per user, `totp_secret: <Base32>` as a plain field of the user entry (`users` is keyed by username), `${ENV_VAR}` allowed as for any value. Validation: Base32 or rejected; a user with a secret but no password is rejected (a factor has to follow a password). No `enabled` per user: the presence of the secret is the enrolment.
- The operator writes the secret in the file. The users form and the MCP user tools neither accept nor return it; an edit through them keeps the secret the file has. No read surface (`/api/users`, MCP `get_user`/`list_users`, `User.to_dict()`, the audit log) returns it, in the same way none returns `password`.

```yaml
# settings.yaml
[...]

PR #350, opened 2026-09-12
feat: declarative TOTP second factor (login.totp, users[].totp_secret)

dshvedchenko:
**Declarative TOTP second factor** (#348, opt-in, off by default):
  `login.totp: true` requires a time-based one-time code after a successful
  password check, across every interactive login surface - `/authorize`,

  `/login`, `/saml/sso` and the device flow's `/device` - the same
  surfaces `login.two_step` covers, riding its phase machinery. A
  declarative demo factor, not IdP hardening: the secret is a plain

  `totp_secret` field of the user entry in `users.yaml` (its presence is
  the enrolment, no per-user `enabled` flag, `${VAR}` allowed like any
  value), never accepted or returned by the users form or any MCP user
  tool. Verification is RFC 6238 (6 digits, 30s period, HMAC-SHA1, one
  step of clock skew), standard library only, no new dependency, and
  deliberately has no replay protection. Inert under `login.mode: persona`,
  which checks no password; the OAuth password grant and client
  credentials are unaffected. While `login.totp` is on, the ID Token
  carries an OIDC `amr` claim (RFC 8176): `["pwd"]` for a plain password
  login, `["pwd", "otp"]` once the code is verified, preserved across a
  refresh like `auth_time`; with it off no new claim appears. A third SAML

  `AuthnContextClassRef` value,
  `urn:oasis:names:tc:SAML:2.0:ac:classes:TimeSyncToken`, next to the
  existing password/unspecified pair. Configurable through YAML, the
  Settings UI, and the MCP settings tools.


cdelmonte-zg (changes requested):
Thank you, this follows the #348 scope closely and the shape is right: one rule in `_auth` (`authenticate_interactively` + `SecondFactorPhase`), each route only renders the phase, the session vocabulary extended through the single writer/reader from #301, `amr` carried like `auth_time`, the secret kept off every read and write surface but the YAML. Verified on the head (dfc1d05): 2145 unit tests, ruff, mypy, import-linter, the e2e agent (64/64 with `NANOIDP_E2E_TOTP_SECRET` set), and live probes on three instances (`login.totp` alone; with `login.mode: persona`; with `login.two_step` and `saml.strict_binding: true`): the four surfaces, wrong/blank/malformed/stale codes, the skew boundary, a tampered hidden password with a valid code (fails closed), the device flow's dead-`user_code` oracle, deny from the code screen, `amr` on code/device/refresh and its absence on the access token, userinfo, the password grant and persona mode, `TimeSyncToken` on the inline login, on a reused pwd+otp session and under the Redirect binding, the users form leaving the secret in place, the audit entries, and the debug logs (no secret, no password). All as documented.

Two blocking, both small; then a suggestion and some cleanup.

### Blocking

**1. Non-ASCII digits turn every code screen into a 500.** `verify_totp` guards the code with `len(code) == 6 and code.isdigit()`, and `str.isdigit()` is true for fullwidth (`１２３４５６`), Arabic-Indic or superscript digits; `hmac.compare_digest` then raises `TypeError` on a non-ASCII `str`. Reproduced on the head: that POST to `/login` (and `/authorize`, `/saml/sso`, `/device`) answers 500 with a traceback instead of "Invalid code". Fix: require `code.isascii()` too (or compare bytes), and a test with a fullwidth code.

**2. An unset `${VAR}` on `totp_secret` silently disables the factor.** `validate_totp_secret` maps `""` to `None`, so `totp_secret: "${ALICE_TOTP}"` with the variable unset loads without a word and alice logs in with the password alone, `amr: ["pwd"]`, `PasswordProtectedTransport`. The same mistake on `password: "${VAR}"` aborts the load (reproduced both ways on the head), and configuration.md promises placeholders "work exactly as they do for `password`". Fix: let the empty string reach `canonical_secret`, which already raises "totp_secret must not be empty" (the `""` -> `None` mapping serves no surface: the form and MCP never send the field).

### A suggestion, not a request

The code screen carries the password forward as a hidden field on all four surfaces, and the route re-checks it. That is the literal reading of "nothing is stored server-side" in the scope, and it is documented, so it can ship as is. If you feel like it, there is a neat way to keep the step stateless without the password in the page: after the password check, hand the screen an HMAC tag over `username|time-bucket` keyed with the app secret (stdlib `hmac`, like `verify_totp`), and accept the code screen's POST when the tag verifies. No store, nothing secret in the DOM, `no_store` no longer needed, and a demo audience looking at the page source sees only a tag. Your call, here or later.

### Cleanup, non-blocking

- `device_verify`: on a wrong password for a live code the credentials are checked twice, `authenticate_interactively` outside the lock and `store.verify()` again inside it, because `authenticate` is rebound only on the success path (counted on the head: `['outside', 'in_lock']` vs `['in_lock']` on the base). Once the pre-check ran, hand `verify()` a callable that returns `login.user` (None on failure) in every case; under a password-hashing profile this is one bcrypt round per attempt instead of two, one of them serialising the store.
- `authenticate_interactively` reads `totp_code` without `.strip()`, unlike `username` on the same forms; a pasted or autofilled code with a trailing space is audited as a failed attempt. One-liner.
- The pending-phase block (audit with a re-spelled `"Invalid code"` literal, `no_store(render(..., totp_step=True, login_password=...))`) is hand-written in all four surfaces, and `no_store` is a per-caller obligation that tests assert only for `/login` and `/authorize`. Apply `no_store` inside the four render closures when `totp_step` is set, take the audit reason from `login.phase.error`, and add the `Cache-Control` assertion for `/saml/sso` and `/device`.
- `DeviceCodeStore.pending_status` re-spells the user_code classification (`INVALID_CODE` / `ALREADY_USED` / `EXPIRED`) that `verify()` has thirty lines below; one private `_status_of(grant)` used by both keeps the oracle guard and the transition on the same rule.
- `authorize.html` now has two identical "Change username" blocks (TOTP branch and password step), while the same PR consolidated login.html's into one block gated on `(two_step_login or totp_step) and login_username`; the same shape works here.
- `claims_supported` lists `amr` whether or not `login.totp` is on, while tokens.md says the claim is opt-in "on the wire too". Gate it on `settings.totp` (the function already has `settings`), or soften the sentence.
- Docs: docs/SECURITY.md links `configuration.md#second-factor-totp`, which exists in neither rendering (GitHub: no `docs/configuration.md`; the book: the guide lives in `guides/` via symlink, siblings use `../reference/configuration.md#...`). And configuration.md's TOTP section sends the reader to `[SAML](saml.md)` for the `AuthnContextClassRef` value, but saml.md is untouched and never mentions the three values; either add them there or point at the security guide.
- CHANGELOG: `totp_secret` is now a reserved top-level user key (before, it folded into `attributes`); one clause, as #247 did for `description`. And the entry sits under `[Unreleased]`, which today is the 3.1.0 content: as agreed on #348 this lands after the 3.1.0 final, and the entry then moves under the new `[Unreleased]`. Nothing to do on your side for that now; I will ping when main is ready and set the milestone.

dshvedchenko:
Both blocking issues and all six non-blocking cleanup items from cdelmonte-zg's review are addressed, with the full suite green (2148 tests, ruff, mypy, import-linter).

Blocking

verify_totp now requires code.isascii() before isdigit(), so fullwidth/Arabic-Indic/superscript digits are rejected instead of raising through hmac.compare_digest. Added a unit test and a route-level regression test hitting all four surfaces (/login, /authorize, /saml/sso, /device) with a fullwidth code, asserting 200 + "Invalid code" instead of 500.
validate_totp_secret no longer maps "" → None; an empty string now reaches canonical_secret, which raises "totp_secret must not be empty" — same failure mode as an unset ${VAR} on password. Added a model-level rejection test and a YAML-load test with the env var deliberately unset.
Cleanup

device_verify: authenticate is now rebound unconditionally after the pre-check (returning login.user, None on failure), so a wrong password is no longer hashed twice.
authenticate_interactively now strips totp_code like every other form field.
no_store moved inside all four render closures (ui.login, saml's inline login, _render_authorize_login, device_verify's render_device); audit reason now reads from login.phase.error instead of a re-spelled literal; added the missing Cache-Control: no-store assertions for /saml/sso and /device.
DeviceCodeStore got a shared _status_of(grant) used by both pending_status and verify.
authorize.html's two identical "Change username" blocks consolidated into one, gated the same way login.html already does.
claims_supported now only lists amr when settings.totp is on, matching tokens.md's "opt-in on the wire too."
Docs

docs/SECURITY.md's link to configuration.md#second-factor-totp now points to ../reference/configuration.md#second-factor-totp, matching the sibling-guide convention (this fixes the book rendering; the GitHub-raw rendering of docs/SECURITY.md has no reference/ counterpart to link to at all — a pre-existing structural gap, not something a single link can close).
configuration.md's TOTP section now points to the security guide (which documents all three AuthnContextClassRef values) instead of saml.md (which documents none of them).
Left untouched, per the review's own framing:

The HMAC-tag statelessness suggestion — explicitly "your call, here or later," a design change rather than a requested fix.
The CHANGELOG [Unreleased] split — reviewer said there's nothing for me to do until they ping about the 3.1.0 milestone.

cdelmonte-zg (changes requested):
Round 2, head b7f7e36.

**Verified on this head**

- 2148 unit tests, ruff, mypy, import-linter: all green.
- e2e agent against a live instance with `login.totp` on and a secret on `user1`: 64/64, "TOTP Login" included.
- Live probes on four instances (totp on / totp on + persona / totp on + two_step + strict_binding / totp off): 91 checks, all pass once four probe artefacts (session reuse, a grant consumed by the token poll) were taken out. Zero tracebacks in the four logs.

Both blockers are fixed:

- Blocking 1: fullwidth, Arabic-Indic, Devanagari, superscript and mixed ASCII/fullwidth codes now answer 200 "Invalid code" on all four surfaces, with `Cache-Control: no-store` on every one of those responses. A padded valid code (`"  123456 "`) verifies, so the strip is in.
- Blocking 2: `totp_secret: "${UNSET}"` aborts startup with "totp_secret must not be empty", and `POST /api/config/reload` on the same file fails and keeps the previous configuration in service (the code screen still shows, a valid code still logs in). The same reload with `password: "${UNSET}"` behaves identically (HTML 500, old config kept); that path is untouched by the PR, so the reload presentation is pre-existing.

The cleanup items check out live too: one credential check per `/device` attempt, the dead-user-code guard holds, `Change username` renders exactly once on both the password and the code step of `/authorize`, `amr` is absent from `claims_supported` with `login.totp` off, and both doc links resolve in the book (the `second-factor-totp` anchor exists).

**Three small things to close in this round**

1. `models.py` - the Base32 rejection and the "requires a password" rule are plain `ValueError`s inside pydantic validators, so the error text appends `input_value=<the secret>`. Reproduced: `totp_secret: "JBSWY3DPEHPK3PX1"` (one wrong character, still a secret) makes `nanoidp validate-config`, the startup log and the MCP validate/reload results print `input_value='JBSWY3DPEHPK3PX1'`; a persona-only user with a secret echoes the whole user dict with the secret at its tail. `totp_secret.py` promises the opposite ("The message never repeats the value"). One line closes it: `hide_input_in_errors=True` in `User.model_config` (I checked pydantic drops the `input_value` part for the whole model; no test asserts on `input_value`).
2. `discovery.py` - `claims_supported` gates `amr` on `settings.totp`, while the only minting site (`authenticate_interactively`) gates on `settings.totp_active`. With `login.mode: persona` and `login.totp: true`, discovery advertises `amr` but no ID Token can ever carry it (verified on the persona instance). `settings.totp_active` there makes the two agree.
3. `.github/workflows/e2e.yml` - the e2e agent's code-screen, wrong-code, valid-code and amr checks only run with `NANOIDP_E2E_TOTP_SECRET` set, and no job sets it, so CI and the nightly only exercise the "user without a secret is unaffected" half. One more job in the e2e-saml/e2e-mcp shape (copy `config/`, append a demo `totp_secret` to a user, export the same value) closes the gap; a line in `e2e/README.md` for the variable would help too, it is documented nowhere yet.

**Non-blocking, take what you like**

- `oauth_grants.py` refresh: `amr` is read from the refresh payload unvalidated and `token.py` does `list(amr)` after the jti has already been claimed. A payload with `"amr": "pwd"` refreshes into `["p","w","d"]`, `"amr": 1` is a 500 after the token is spent. Key-holder scenario only (the token must be signed with the IdP key), `auth_time` is equally unguarded, so a cleanup: normalise to a list of str, else drop it.
- `device_code.py` `verify()`: the docstring still says the credential check runs inside the lock with `authenticate`, but after your fix the route decides outside and passes a constant closure. Either update the docstring or take the natural signature `verify(user_code, action, user: Optional[User], *, amr)`, which also removes the lambda and its `noqa`. Separately, after `_status_of` the three `if status is X: return X, None` branches plus the `assert` can collapse to: early `if not grant: return INVALID_CODE, None`, then `status = self._status_of(grant)`, set `grant.status = "expired"` when `EXPIRED`, `if status is not None: return status, None`. Behaviour and lock scope unchanged; a fourth non-live classification added later would then not fall through to the deny/authorize path.
- `tokens.md` `amr` section: "issued through an interactive login surface (/login, /authorize, /saml/sso, the device flow)". Only `/authorize` and the device flow issue ID Tokens; `/login` and `/saml/sso` feed the session and the SAML context, and `/authorize` never reuses the `/login` session. Worth one sentence.
- CHANGELOG, whenever the entry moves under the new `[Unreleased]` after the 3.1.0 final, two clauses: the round-1 one (`totp_secret` is now a reserved top-level user key, before it folded into `attributes`, the way #247 noted `description`), and a new one: `amr` joined `_RESERVED_CLAIMS`, so a user attribute named `amr` requested via the `claims` parameter is now silently dropped regardless of `login.totp` (the #110 convention). Placement itself stays as agreed, nothing to do there now.
- `no_store` is now spelled as `no_store(response) if totp_step else response` with the same comment in four closures, keyed on `totp_step` while the sensitive payload is the hidden password. Making every interactive login render no-store unconditionally removes the four conditionals and comments; nothing asserts an absent `Cache-Control`. Your call.
- A whole-entry save (form edit of an unrelated field, MCP `save_config`) writes `totp_secret` back through `user_to_yaml`, so a `${VAR}` placeholder is materialised and a non-canonical spelling (`jbsw y3dp ...`) is rewritten to the canonical one. `password: "${VAR}"` already behaves this way on main, so no change asked; a sentence in configuration.md next to "`${VAR}` allowed like any value" would save an operator a surprise.
- `_handle_authorize_login` now returns a 3-tuple (seven `, False` returns) and `authorize()` re-reads the password from the form to forward it. `/login`, `/saml/sso` and `/device` render their pending screen inside their own closure; `/authorize` could too by passing `client` into the helper. Fine as a follow-up issue rather than this PR.

The hidden password on the code screen stays as you decided; the HMAC tag was a suggestion and remains one. CHANGELOG placement waits for the 3.1.0 final, as agreed.

dshvedchenko:
All work is done and verified — full suite green (2150 passed), ruff/mypy/import-linter all clean, and a live smoke test confirmed the new CI job (64/64, matching the reviewer's own result). No commit made per your standing preference.

Summary of what changed, across the scope you picked:

3 blocking:

[models.py](vscode-webview:///src/nanoidp/models.py) — hide_input_in_errors=True on User.model_config; verified totp_secret no longer leaks into validation error text.
[discovery.py](vscode-webview:///src/nanoidp/services/discovery.py) — claims_supported now gates amr on settings.totp_active instead of settings.totp, so persona+totp no longer advertises a claim nothing can mint. Added two regression tests in test_totp_login_flows.py.
[.github/workflows/e2e.yml](vscode-webview://.github/workflows/e2e.yml) + [e2e/README.md](vscode-webview:///e2e/README.md) — new "Prepare/Start/Run TOTP" steps put a demo secret on user1 (not the agent's own default user, to avoid breaking the "no secret" checks) and export NANOIDP_E2E_TOTP_SECRET, so test_totp_login's code-screen checks actually run in CI.
amr refresh cleanup:

[token.py](vscode-webview://src/nanoidp/services/token.py) — new _sanitize_amr helper drops non-list/tuple values and non-string entries before create_token uses amr, so a hand-crafted refresh payload ("amr": "pwd" or "amr": 1) is dropped instead of exploding into characters or crashing after the token is spent.
device_code.py cleanup:

verify() now takes an already-resolved user: Optional[User] instead of username/password/authenticate, removing the lambda _u, _p: decided_user # noqa: E731 at the [oauth.py](vscode-webview://src/nanoidp/routes/oauth.py) call site and matching docstrings to what the code actually does. Collapsed the status-branch logic as suggested. Updated the two direct-store tests in test_resource_indicators.py.
Docs:

[tokens.md](vscode-webview://src/reference/tokens.md) — corrected which surfaces actually mint the ID Token carrying amr (/authorize and the device flow, not /login//saml/sso).
[configuration.md](vscode-webview://book/src/reference/configuration.md) — noted that a whole-entry save materializes a totp_secret ${VAR} and canonicalizes its spelling, same as password.

cdelmonte-zg (commented):
Round 3, head 4c0a778.

**Verified on this head**

- 2150 unit tests, ruff, mypy, import-linter: all green. CI green, and the new TOTP step in `http-e2e` runs the agent 64/64 against the TOTP server.
- The new job is load-bearing: with `NANOIDP_E2E_TOTP_SECRET` set against an instance where `user1` has no secret, the agent exits 1 with "TOTP Login" failing; without the variable it takes the skip path. So CI now really exercises the code screen.
- Live probes on four instances (totp on / totp on + persona / totp on + two_step / totp off), zero tracebacks in the logs.

All three asks from round 2 are closed:

1. A near-miss secret and a persona-only user with a secret no longer echo the value in `nanoidp validate-config`, the startup log, the MCP validate/reload results or the reload traceback.
2. `claims_supported` follows `totp_active`: no `amr` with persona + totp, and it appears and disappears when `login.mode` is switched at runtime.
3. See the e2e job above.

The non-blocking items you took check out too:

- `DeviceCodeStore.verify(user_code, action, user, *, amr)`: missing credentials, wrong password, unknown user, dead code (same answer with right and wrong password, no code screen), code screen without a status transition, wrong / fullwidth / valid code, already used before the token poll, deny, a user without a secret, persona and two_step all behave as before.
- `_sanitize_amr`: hand-signed refresh tokens with `amr` as a string, int, dict, nested list or empty list refresh with 200 and no `amr`; `["pwd", 1]` refreshes into `["pwd"]`; a warning is logged each time and nothing raises.

**Three small things left**

1. `book/src/reference/configuration.md` - the new sentence wraps so that a line starts with `  - spaces, lowercase - ...`. Inside the list item that is a nested bullet: the built book renders `<ul><li>spaces, lowercase - to the canonical one). To set the secret up in an authenticator app...` as a sub-list, cutting the parenthesis and moving the setup-key/QR guidance into it. Rewrapping so no line starts with `- ` (or using commas instead of the dashes) fixes it.
2. `book/src/reference/tokens.md`, `amr` section - "Only `/authorize` and the device flow mint ID Tokens" is not right, and that wording came from my round-2 comment, sorry about that: the password grant also returns an ID Token with `scope=openid` (verified), as do refresh and MCP `generate_token`, and the same paragraph mentions the password grant a few lines later. Something like "Of the interactive login surfaces, only `/authorize` and the device flow mint ID Tokens" keeps the point.
3. `hide_input_in_errors` on `User` covers the validators, but `users.yaml` is validated first by `UsersDocument`/`UserEntry` in `config_documents.py`. A secret that YAML reads as a number (an unquoted all-digit Base32 value, e.g. `totp_secret: 2345672345672345`) fails there with `string_type`, and the startup traceback prints `input_value=2345672345672345` through the chained exception. My round-2 suggestion only named `User`, so this is a gap in what I asked, not in what you did. The same happens on main today with a numeric `password`, so it is not new to this PR; setting the flag on `UserEntry` (and `UsersDocument`) closes it for both. Unlikely for a generated secret, so fine here or as a follow-up, your call.

**Nit**

- `oauth_grants.py`: the new comment says `create_token` runs `amr` through `sanitize_claim_names`; it is `_sanitize_amr`.

CHANGELOG as agreed: when the entry moves under the new `[Unreleased]` after the 3.1.0 final, with the two clauses from round 2 (`totp_secret` is now a reserved top-level user key; `amr` joined the reserved claims). We are very close now.

cdelmonte-zg (approved):
Round 4, head 8e22c0d, plus two commits of mine on top.

**Verified on 8e22c0d**

- 2150 unit tests, ruff, mypy, import-linter: all green, CI green.
- `configuration.md`: the built book now renders the setup options as an intentional two-item sub-list, with the sentence before it intact.
- `tokens.md`: the `amr` paragraph now scopes the claim to the interactive surfaces.
- `hide_input_in_errors` on `UserEntry` and `UsersDocument`: an all-digit `totp_secret` no longer appears in the startup traceback, `nanoidp validate-config`, or the MCP validate/reload results. The same change also covers a numeric `password`, which main still echoed.
- The `oauth_grants.py` comment names `_sanitize_amr`.

**What I pushed to your branch** (fast-forward, your commits untouched)

With 3.1.0 now released, the branch needed main, and a plain merge was not enough: git auto-merges `CHANGELOG.md` without a conflict and drops the TOTP entry inside the released `[3.1.0]` section, because its context lines still belong to the old `[Unreleased]`. So:

1. `faffc9f` merges main (607d003) into `iss_348`. Outside the CHANGELOG the branch diff is identical before and after (41 files, +2825/-208), and the suite is green on the merged tree.
2. `97242cd` moves the entry under the new `[Unreleased]` and adds the two clauses from round 2. I checked both against main before writing them: `totp_secret` used to fold into `attributes`, and a user attribute `amr` used to resolve through the `claims` parameter. The `[3.1.0]` section is byte-identical to main.

Approving. Thanks for carrying this through four rounds, and for keeping it to the declarative scope we agreed in #348.

cdelmonte-zg:
TOTP shipped in 3.2.0: `pip install nanoidp==3.2.0`, `docker pull ghcr.io/cdelmonte-zg/nanoidp:v3.2.0`, or chart `--version 3.2.0`. Release notes: https://github.com/cdelmonte-zg/nanoidp/releases/tag/v3.2.0. Thanks again for the work across the four rounds.

Issue #358, opened 2026-09-15
jwt.external_keys and jwt.max_previous_keys are documented but no configuration reaches them


cdelmonte-zg:
*Revised 2026-09-15: widened to `jwt.max_previous_keys` and the direction decided (first as a comment); acceptance criteria added. Earlier text in the edit history.*

**Defect.** `docs/SECURITY.md` documents two `jwt` settings that no configuration can reach:

```yaml
jwt:
  external_keys:
    private_key: /path/to/private.pem
    public_key: /path/to/public.pem
    kid: "my-custom-key-id"
  max_previous_keys: 2
```

Verified on main (4d1160c):

- `JwtSection` accepts only `algorithm` and `keys_dir`. A `settings.yaml` with either block loads with `unknown key jwt.external_keys (ignored)` / `unknown key jwt.max_previous_keys (ignored)`; `Settings.external_private_key`, `external_public_key` and `external_key_id` stay `None` and `Settings.max_previous_keys` stays 2. The flat spellings (`jwt.external_private_key`, ...) are ignored the same way.
- The `Settings` fields exist (`models.py`), `create_app()` passes them to `init_crypto_service()`, `CryptoService` loads external keys (covered by `tests/test_key_management.py`, which constructs the service directly), and the keys page, `GET /api/keys/info` and the MCP `get_keys_info` tool display `max_previous_keys`. Nothing between the YAML document and those fields sets them. `git log -S` traces the fields (since moved into `models.py`, #86) and the documented block back to the first commit.
- The docs say that without `kid` "one is generated from the key fingerprint"; `CryptoService` uses `uuid.uuid4().hex`, so the same key pair gets a different `kid` on every start.
- `CryptoService` checks that both PEM files parse, but not that the public key belongs to the private key.

A user following the docs gets generated keys and a warning they may not read, while the docs say their own key signs the tokens.

**Direction: wire them.** The capability exists almost entirely, and fixing the signing key is useful for a test IdP. Depends on #359, whose activation step keys the signing service by all its inputs.

**Acceptance criteria**

- `jwt.external_keys` (`private_key`, `public_key`, optional `kid`) and `jwt.max_previous_keys` are part of the settings document model, mapped onto the existing `Settings` fields, and the committed JSON schema (`docs/schema/config.v1.json`) is regenerated.
- `private_key` and `public_key` are given together or not at all (a load error otherwise); `max_previous_keys` keeps its model bounds (0-10).
[...]

PR #360, opened 2026-09-15
fix: a reload no longer activates a configuration whose signing service cannot be built

### LINKED ISSUE #359: A reload accepts an unusable jwt.keys_dir and answers 200
*Revised 2026-09-15: the body now carries the agreed design (first posted as a comment, then corrected on the reader order, the commit order and `validate-config`). Earlier text in the edit history.*

**Invariant.** A configuration does not become active if the essential services derived from it cannot be built. A writable-directory check would not cover it (malformed or external keys, #358), and a generic dependency-validation framework is more than this needs.

**Defect.** A configuration reload accepts a `jwt.keys_dir` the process cannot use, and answers 200. Reproduced with a real server: start on a working `keys_dir`, set `keys_dir: "/proc/nanoidp-no-such-dir"` in `settings.yaml`, then `POST /api/config/reload`:

| | before #357 (db64195) | since #357 |
|---|---|---|
| `POST /api/config/reload` | 200 | 200 |
| `/.well-known/jwks.json` | 500 (HTML) | 500 (HTML) |
| password grant at `/token` | 200, signed with the previous key, which the JWKS no longer serves | 500 (HTML) |
| server log | `FileNotFoundError` traceback | `FileNotFoundError` traceback |

At startup the same directory makes the process exit 1 with the same traceback.

**Facts on main (4d1160c)**

- A configuration becomes active on four paths, all through `ConfigManager._load_config`: the first load, `reload()` (`POST /api/config/reload`, MCP `reload_config`) and `reload_local()` (after every UI write and inside `ConfigManager.save()`). `_load_config` stages and validates both files before committing anything; that is the transaction to extend.
- `_commit_directory()` configures the hooks and plugins first, the only step that can still fail (strict plugin load, rolled back by the registry), and then assigns settings, users, default user and revisions.
- `config` must not import `services` (import-linter layers), so the manager cannot build the signing service itself.
- `POST /api/config/reload` handles only `HookError` (503 JSON); a configuration that fails validation answers Flask's HTML 500, with the previous configuration kept.
- `get_crypto_service(keys_dir)` rebuilds the service when `keys_dir` differs (#281), with the default `max_previous_keys`, so a request holding an older `settings` snapshot would rebuild the old service over a newly published one.
- `TokenService.create_token()` resolves the signing service before it reads `self.config.settings`.

**Design**

1. **One activation step inside the load transaction.** `init_config(..., activate=...)` takes a single callable, supplied by the process composition (`create_app` and the MCP server's `_ensure_config`). Its only use is the signing service: `prepare(candidate settings)` builds a `CryptoService` for the candidate without publishing it, and returns the published one when its inputs are unchanged, so an unrelated reload neither rebuilds nor regenerates keys. The load then runs in this order, with nothing fallible after the publish:

   ```text
   stage + validate settings.yaml and users.yaml
         |
   prepare the candidate signing service      (may fail: nothing committed)
         |
   configure hooks and plugins                (may fail: registry rolls back, nothing published)
         |
   publish the candidate signing service
         |
   assign settings, users, default user, revisions
   ```

2. **The signing service is keyed by all its inputs:** `keys_dir`, the external key paths and key id, and `max_previous_keys`. #358 then gets "a changed `jwt.external_keys` or `jwt.max_previous_keys` reinitialises the signing service transactionally on reload" without separate logic.
3. **Publish order and read order.** Activation publishes the signing service before the settings. Readers observe them in the opposite order: token issuance captures the settings snapshot first and resolves the published signing service after it, so `create_token()` reads `self.config.settings` before `self.crypto`, and a route that reads settings earlier in the request keeps that order. A concurrent request can then observe only old settings with the old signing service, old settings with the new one, or new settings with the new one, never new settings with the old signing service. `get_crypto_service()` returns the published service and no longer rebuilds on a `keys_dir` argument; the argument is removed from its 19 call sites (mechanical), since callers holding the old settings would otherwise undo the publish. No runtime snapshot object.
4. **One error surface.** A rejected configuration, whether by validation or by activation, answers `422` with a JSON body (`status`, `error`, `kind`) from `POST /api/config/reload` and a `success: false` result from MCP `reload_config`, the previous configuration staying in effect; `HookError` keeps its 503. At startup the same condition is a configuration error message instead of a traceback. A failed `reload_local()` after a UI write keeps its existing contract: the write is durable, `ReloadAfterSaveError`.

**Out of scope, deliberately**

- `nanoidp validate-config` does not run activation: it must not create directories or generate keys. It keeps validating structure and values without side effects, so an operationally unusable `keys_dir` is reported by startup and reload, not by `validate-config`.
- A per-request `settings` snapshot across every route: point 3 is enough for signing.
- Other derived state (Flask's `secret_key`, the rate limiter): the activation step can take them later if a case appears.
- Side effect to accept: `prepare` may create the new `keys_dir` and generate keys there even when the load then fails on a strict plugin load; those keys stay unused and harmless.

**Tests**

- A reload to an unusable `keys_dir` answers 422 JSON, the previous key keeps signing and the JWKS keeps serving it; the same through MCP `reload_config` (`success: false`).
- A reload with unchanged key inputs keeps the same service instance and `kid`.
- A strict plugin load failure after a successful prepare publishes nothing: the previous signing service and settings stay.
- Startup with an unusable `keys_dir` fails with the configuration error message and no traceback.
- Read order pinned: with the publish interleaved between the two reads of `create_token()`, the token is signed by the service matching the settings it read or a newer one, never an older one; and no call path rebuilds an older service over a published one.
- The #357 keys_dir reload test keeps passing.

Found while verifying #357. #358 builds on this.

cdelmonte-zg:
Closes #359, implementing the design in its body: a configuration does not become active if the signing service it needs cannot be built.

## What changes

- **One activation step in the load transaction.** `init_config(..., activate=...)` takes a single callable; `create_app` and the MCP server pass `activate_crypto_service` (config still does not import services). The load now runs:

  ```text
  stage + validate settings.yaml and users.yaml
  prepare the candidate signing service      (fails -> nothing committed)
  configure hooks and plugins                (fails -> nothing published)
  publish the signing service
  assign settings, users, default user, revisions
  ```

  `prepare` reuses the published service while its inputs (`keys_dir`, the external key paths and id, `max_previous_keys`) are unchanged, so an unrelated reload neither rebuilds it nor touches keys.
- **Loads run one at a time** (a reentrant lock on the manager). Found by `/code-review` on the first head and reproduced: two concurrent reloads to a fresh `keys_dir` each generated keys over the other, and the published service held a private key and a public key from different pairs in 8 of 10 rounds; 0 of 20 with the lock. It also keeps one load's signing service from being published next to another load's settings.
- **Readers take the settings before the signing service.** `get_crypto_service()` takes no argument and returns the published service, so no reader rebuilds an older one over it (its 19 call sites lose the `keys_dir` argument). `create_token()` takes one settings snapshot before resolving the signing service, and uses it for the authorities and the ID Token audience; the SAML response builder, metadata and AttributeQuery signing, the refresh grant, `/userinfo` and the keys page read settings first too (the last three since round 2). The keys page takes the key's directory and retention from the service itself.
- **One error surface.** A failed load raises `ConfigurationRejected` (a `ValueError`, original error as cause) with `kind` `invalid` (unreadable or invalid files) or `activation`. `POST /api/config/reload` answers `422` JSON `{status, error, kind}`, MCP `reload_config` an error result with the same `kind`; a strict hook or plugin failure keeps its `503`. The server and the MCP server print `error: configuration rejected: ...` and exit 1 at startup instead of a traceback. The refresh after a UI write goes through the same load.
- `init_crypto_service` is gone (replaced by the activation step); docs: `architecture.md`, `guides/extending.md`, `reference/configuration.md`; CHANGELOG under Fixed and Changed.

Out of scope, as the issue settles: `validate-config` (no side effects, no activation), a per-request settings snapshot across every route, other derived state.

## Before and after, real server

`jwt.keys_dir` moved to a path inside a regular file, then `POST /api/config/reload`:

| | main (4d1160c) | this branch |
|---|---|---|
| reload | 200 | 422 `{"kind": "activation", "error": "...settings.yaml: JWT signing configuration cannot be activated (NotADirectoryError: ...)"}` |
| JWKS | 500 HTML | 200, previous key |
| `/token` | 500 HTML | 200, previous `kid`, introspection active |
| a following valid reload | 200, new `kid` | 200, new `kid` served and signing |
| log | `NotADirectoryError` tracebacks | no traceback |

MCP over real stdio: `reload_config` returns `success: false`, `kind: activation`, the active `kid` unchanged; `nanoidp-mcp` started on that directory exits 1 with the message and no traceback.

## Tests

- `tests/test_reload_activation.py` (14): HTTP 422 with the previous key still signing and served; MCP `reload_config`; invalid YAML answers 422 `invalid`; an unreadable `settings.yaml` answers 422 `invalid`; the refresh after a UI write rejects it too; an unchanged-input reload keeps the service instance; a strict plugin failure after a successful prepare publishes nothing (the candidate's keys exist, the old service stays); loads never overlap; no function under `src` reads settings after resolving the signing service (AST); a reload changing `oauth.audience` right after `/userinfo` and the refresh grant resolve the service does not change the audience they verify with; startup with an unusable `keys_dir` and with invalid YAML exits 1 with the message.
- The read order is pinned by a test that reloads right after `create_token()` reads its settings: the token is signed with a key the JWKS serves. With the two reads swapped the test fails on the `kid`, checked; a first version that reloaded inside the signing-service read passed either way and was rewritten. The loads-never-overlap test fails without the lock, checked.
- `tests/test_key_management.py::TestSigningServiceActivation` (5) replaces the #281 tests: reuse on unchanged inputs, a changed `keys_dir` or `max_previous_keys` prepares a new service without publishing it, external key inputs are part of the identity, readers never rebuild over the published service.
- 2204 unit tests, ruff, mypy, lint-imports; `e2e/test_agent.py` 64/64 against a branch server; `e2e/mcp_smoke_test.py` over stdio.

The unusable directory in the tests is a path inside a regular file, so the cases hold on every platform.


cdelmonte-zg:
Round 2 (981627a), from a review of the first head (f9fc75c).

- **Reader order, three sites.** The refresh grant and `/userinfo` resolved the signing service and then read `oauth.audience`; the keys page described the key with `keys_dir` and `max_previous_keys` read after it. Each now takes one settings snapshot first, and the keys page takes the key's directory and retention from the service itself (it still shows the configured spelling, `./keys`, while that names the service's directory). A census of every function calling `get_crypto_service()` or `self.crypto` found no other site reading settings afterwards.
- **Regression tests.** A structural test walks every function under `src` and fails on a settings read after the signing service is resolved, so a new call site cannot reintroduce the pairing. Two behavioural tests land a reload changing `oauth.audience` right after `/userinfo` and the refresh grant resolve the service. All three fail on f9fc75c and pass on 981627a.
- **Neutral activation error.** `JWT signing configuration cannot be activated (<Error>: ...)`, since external keys and retention become signing inputs with #358.

2204 unit tests, ruff, mypy, lint-imports; `e2e/test_agent.py` 64/64 against a branch server.

PR #361, opened 2026-09-15
fix: jwt.external_keys and jwt.max_previous_keys are read from settings.yaml

### LINKED ISSUE #358: jwt.external_keys and jwt.max_previous_keys are documented but no configuration reaches them

cdelmonte-zg:
Closes #358: `jwt.external_keys` and `jwt.max_previous_keys`, both documented in the security guide and ignored by the loader as unknown keys, now reach the signing service, through the activation step of #359/#360.

## What changes

- **Document model and schema.** `JwtSection` gains `external_keys` (`private_key`, `public_key`, optional `kid`; both paths required together) and `max_previous_keys` (0-10), mapped onto the existing `Settings` fields; `docs/schema/config.v1.json` regenerated. `${VAR}` placeholders work in the paths. UI and MCP saves keep the `jwt` block (the writer only rewrites the keys it owns; tested).
- **Signing inputs.** Both settings are inputs of the signing service, so they apply at startup and on every reload, in the server and in the MCP server: the configured key signs every token and is the only key the JWKS serves; a reload that changes either setting publishes a new service.
- **Key validation, at activation.** The two keys must be RSA and the public key must belong to the private key (a mismatched pair would sign with one key while the JWKS serves the other). A missing, unreadable or malformed file, or a mismatched pair, rejects the configuration: startup exits with `error: configuration rejected: ...`, a reload answers 422 and the running key stays.
- **Stable `kid`.** Without `kid`, the key id is the RFC 7638 thumbprint of the public key. The docs promised "generated from the key fingerprint"; the code used `uuid4()`, a different kid on every start.
- **Rotation refused for external keys.** `POST /api/keys/rotate` answers 409, the keys page and MCP `rotate_keys` return an error with one fixed message, and nothing is written. Before, a rotation generated a pair into `keys_dir` and signed with it, against the operator's configuration, until the next start brought the external key back. The files are not watched: a key replaced at the same paths is read at the next start (documented and tested).

### From `/code-review` on the first head (all reproduced, then fixed)

- **SAML certificate.** The external key's certificate overwrote `keys_dir/idp-cert.pem`, and the generated-key path only regenerated that file when missing, so dropping `external_keys` again left SAML signing with a certificate for the wrong key (signxml verification against the metadata failed). The external key's certificate now lives in `external-cert-<thumbprint>.pem`, generated once and reused (stable for SPs that pin it); a certificate that does not belong to the signing key is regenerated; certificate writes are atomic; SAML signing uses the published service's certificate instead of re-reading the file on every request.
- **Retention.** A lowered `max_previous_keys` left the JWKS unchanged until the next rotation; it now trims the served previous keys when the service is built. The public-key files of dropped keys may remain in `keys_dir/previous/`, unserved; the next rotation rewrites `keys.json` with the retained set (documented).

## Verification

- `tests/test_jwt_signing_settings.py` (23). The first 16 all fail on main (661cccc). Disabling each guard in turn fails its tests: the pair check (2), the thumbprint kid (1), the rotation refusal (1). The 5 review tests (a signed SAML Response verified against the `/saml/metadata` certificate across generated -> external -> generated keys, a stable external certificate, a foreign `idp-cert.pem` replaced, a strict-plugin-rejected reload leaving the active signing service and certificate unchanged, lowered retention trimming the JWKS) all fail on the first head.
- 2227 unit tests, ruff, mypy, lint-imports.
- Real server: tokens signed with the external kid and only that kid served; rotate 409; reload to a mismatched pair 422 with the running key kept; reload to another valid pair switches. Startup with a mismatched pair exits 1 with the message. MCP over stdio: `generate_token` signs with the external kid, `rotate_keys` is refused.
- **E2E.** The agent gains `test_external_signing_keys` (gated by `NANOIDP_E2E_EXTERNAL_KID`, optional `NANOIDP_E2E_EXTERNAL_PUBLIC_KEY`), and `test_key_rotation` skips on such a server. `e2e.yml` gains a server on :8005 configured with a generated external pair. The whole `http-e2e` job, run step by step from the workflow file in a copy of the tree, exits 0 on every suite (default, external keys and TOTP each 65/65). A wrong expected kid fails the new check (64/65).
- `docs/SECURITY.md` "Key Management" rewritten against the code, including the generated key file names, which were wrong.


cdelmonte-zg:
Round 2 (52c93e1), from a review of f18f7a6 and CodeQL.

- **Retention docs.** Lowering `max_previous_keys` trims only the served list, so the next rotation cannot delete the files of keys it no longer knows; the security guide and the code comment said it would. Both now say what happens, checked with a probe of the sequence: lowered 3 -> 1, the JWKS drops to 2 keys while 3 files stay; raised back before a rotation, the dropped keys are served again; after a rotation `keys.json` lists only the retained key and raising the value no longer serves the orphaned files. No filesystem cleanup added; the JWKS contract is unchanged and tested.
- **CodeQL alert 28** (`py/stack-trace-exposure`, `routes/api.py`). The 409 returned `str(exc)`. The refusal text is now a module constant (`EXTERNAL_KEYS_NOT_ROTATABLE`) that the exception, the API, the keys page and MCP `rotate_keys` use directly.

2227 unit tests, ruff, mypy, lint-imports.

Issue #354, opened 2026-09-14
Optional SQLite RuntimeStore for several nanoidp processes on one host


cdelmonte-zg:
**Context.** nanoidp is a test identity provider ([VISION](https://github.com/cdelmonte-zg/nanoidp/blob/main/VISION.md#what-comes-next)), and the runtime state chain is **#363 -> #354 -> #193**. This is step 2 of 3: an optional SQLite backend behind the runtime store boundary, so that several nanoidp processes on one host share runtime state. **#193 is the concrete consumer**: a Streamable HTTP MCP companion is a separate process, and without a shared store its tools would report runtime state the IdP does not have. The other needs stay: several workers of one IdP, and a test environment whose runtime objects survive a restart of the IdP.

**Where the line is.** The project is bounded by test versus production, not by size. This store is for one host, one local filesystem, one test environment. Memory stays the default. Nothing here moves nanoidp towards operating as a production IdP: no HA, no multi-node state, no durability promise beyond "the file is disposable".

Everything this issue was waiting for is done: the repository contract (#404), promotion holds on the entry (#405), protocol state and the audit behind the boundary (#363), the cost of a cleanup (#417). **The census below changes what this issue asks for**: a second backend is the smaller half. The larger half is what stops being true when a second process exists.

## Census

Measured on main (`6a78d09`), SQLite 3.45.1, ext4.

**What one decision costs in SQLite.** A decision-shaped write (`BEGIN IMMEDIATE`, delete what is due through a partial index on the time, an existence check, an insert, `COMMIT`) over ten thousand rows:

```text
WAL    + synchronous=NORMAL    22 us median, p99 0.24 ms
WAL    + synchronous=FULL      1.5 ms
DELETE + synchronous=FULL      7.7 ms
a read by name                 3.3 us
```

With WAL and NORMAL a write costs less than the in-memory backend's at the same size, whose write copies the index.

**A process with a stale declared configuration disagrees with its peer.** Reproduced with two `ConfigManager`s over one directory and one store: A promotes the runtime client `x`, which is then declared; B, which has not reloaded, still creates a *runtime* `x`; A resolves `x` to the declared redirect URIs, B to the runtime ones. Today a process learns of a change to the declared configuration only from its own writes and from an explicit reload. This is the requirement #404 and #405 handed to this issue.

[...]

PR #362, opened 2026-09-15
feat: users and clients resolve through one identity resolver over a runtime store

cdelmonte-zg:
Implements **phase 1 of #235**: an in-memory home for runtime-created users and clients, and one resolver that every protocol surface uses, with no surface that creates runtime objects yet (that is #192). Phase 2 of #235 (authorization codes, device codes, revocations and the audit behind the same boundary) is not part of this PR, so the issue is referenced, not closed.

## What changes

- **`services/runtime_identities.py`: the store.** Exactly two repositories, `users` and `clients` (create, get, list, delete, delete_all), in process memory behind one lock. By value: every object crosses the boundary as a copy both ways (create, get, list), so a change is a delete and a create, as with a backend that serializes. The repository contract is tested against the interface, parametrized over store factories, so a later backend (#354) targets the same suite.
- **`services/identities.py`: the `IdentityResolver`.** It composes the declared configuration with the store, and the rules live there only:
  - **declared first**: a name resolves to the declared object when there is one;
  - **no shadowing on creation**: a runtime object cannot be created under a declared name (`DeclaredNameCollision`);
  - **declared wins on reload**: after every load, runtime objects whose names the configuration now declares are removed with a warning (`init_config(after_load=...)`, passed by `create_app`; no audit event until #192);
  - listings carry the origin (`ResolvedUser(user, origin)`), as a contract for #192; they are not wired into any HTTP, UI or MCP listing.
- **Every protocol resolution goes through the resolver**: `/authorize` (including persona auto-login), `/login`, `/token` and all its grants, client authentication (basic, post, public), `/userinfo`, `/introspect`, `/revoke`, device authorization and polling, SAML SSO and AttributeQuery, and the ID Token audience in `create_token` (from the response's settings snapshot, #360). `ConfigManager.authenticate`, `interactive_authenticate` and `check_client` moved to the resolver; `ConfigManager.get_user`/`get_client` stay and mean declared objects only.
- **Management surfaces stay declared-only** until #192: UI pages and forms, the persona picker, `/api/users` including `POST /api/users/<username>/token`, and the MCP server (a separate process with its own, empty store).
- **Concurrency across a reload.** A runtime object is created with loads held off (`ConfigManager.holding_loads()`), so the declared-name check and the insert cannot straddle a reload that declares the name. Lookups read the store before the declared state; the reload assigns the declared state before it removes a shadowed runtime object, so a lookup spanning that reload finds one of the two. A client lookup from an older settings snapshot falls back to the current settings.

No HTTP, UI or MCP behaviour changes: nothing creates runtime objects yet.

## Verification

- `tests/test_runtime_identities.py` (55): repository contract (including 8 threads creating one name: exactly one succeeds, and objects returned by create, get or list changed in place leaving the repository unchanged); resolver rules; runtime identities through every protocol surface; one observable behaviour per resolution site (scope ceilings rechecked at redemption after the client is replaced by delete and create, public and `client_secret_post` clients, refresh, device flow, revocation, introspection policy, `additional_audiences`, default user, AttributeQuery, auto-login); management surfaces still declared-only; reconciliation on reload, after a UI write and on a rejected reload.
- **Mutation run.** Each of the 27 resolver call sites in `routes/` and `services/token.py` was replaced in turn by a declared-only lookup. On the first head 8 failed a test and 19 did not, so a test was added per site; now all 27 do. Removing the `after_load` wiring or the collision check also fails a test.
- **Concurrency** (from `/code-review` on the first head, reproduced before fixing). With two real threads, a create that paused between check and insert while another thread reloaded left `ci-bob` under a declared name, and it came back with its own password once un-declared; a lookup spanning a reload returned `None`. The three regression tests (two-thread create vs reload, lookup spanning a reload, older settings snapshot) fail on the first head and pass now.
- 2282 unit tests (24 test calls of the moved `ConfigManager` methods retargeted), ruff, mypy, lint-imports.
- The whole `http-e2e` job, run step by step from the workflow file on a copy of the tree, exits 0 on every suite (default, TOTP and external keys each 65/65); `e2e/mcp_smoke_test.py` passes.


cdelmonte-zg:
Round 2 (f7c2734), from a review of 06fac70.

- **Repositories are by value.** `create()` copied the incoming object but returned the stored instance, and `get()`/`list()` returned stored instances too, so a caller could change a stored user or client without a repository operation and outside the lock, and a serializing backend (#354) could not honour the same interface. Every object now crosses the boundary as a copy both ways. New contract cases change an object returned by `create`, `get` or `list` in place and check the repository is unchanged, for users and clients; all six fail on 06fac70.
- **Policy changes in tests go through the repository.** The tests that narrowed a runtime client's scopes or resources between authorization and redemption mutated the stored instance. They now delete the client and create the replacement, the operations #192 will have, so they show that code, refresh and device redemption re-read the current client. The mutation run over the 27 resolver call sites still fails a test for every site.
- **`AfterLoad` contract.** The comment now states it is a post-commit callback that must not raise, since the configuration is already in effect when it runs; a fallible backend (#354) has to revisit this.

2282 unit tests, ruff, mypy, lint-imports.

PR #364, opened 2026-09-16
feat: disposable runtime users and clients through /api/runtime

### LINKED ISSUE #192: Runtime-created users and clients: disposable test state separate from declared configuration

cdelmonte-zg:
Closes #192, on the store and resolver of #235/#362: a CI job or an integration test creates users and clients on a running IdP, uses them in every protocol flow, and removes them without touching `users.yaml` or `settings.yaml`, unless one is explicitly promoted.

## The surface

`POST /api/runtime/users` and `/api/runtime/clients` take a declared entry (plus `username` for a user), validated by the same models, defaults and rules as a YAML entry; `GET` and `DELETE` per object; `DELETE /api/runtime` removes everything and answers the counts; `POST .../promote` writes one into the declared file. There is no update: the store is by value (#235), so a change is a delete and a create. Errors are JSON with an `error` and a `kind` (`invalid`, `not_found`, `declared`, `exists`, `promotion_in_progress`, `conflict`, `lock_unavailable`, `write_failed`, `reload_failed`). No password or client secret is ever returned or audited. Writes follow the `management_secret` gate of `/api`; reads stay open.

Create, delete, reset and promote are audit events, as are a reload removing a runtime object and an abandoned promotion.

## Promotion

Two rules carry it:

- **The promotion holds reloads off** (`ConfigManager.holding_loads`, #235), so the only reload that can see its mark in the writing state is the one its own write triggers. Someone declaring the same name at that moment either wins the race to the file (the promotion answers `409` and that declaration's own reload removes the runtime object as a collision) or writes after it.
- **The mark has two states.** Written but not yet reloaded survives the request: every later successful load resolves it, as `promoted` when the name is declared by then (exactly one event, no collision warning) or as abandoned, with a warning and an audit event, when the entry is gone. No mark outlives a successful load.

A failing strict `on_config_saved` hook does not undo a promotion (`mirror_hook_error` in the response); any other hook failure, and a rejected configuration, are `reload_failed`. `DELETE /api/runtime` waits for a promotion in progress and keeps an object whose promotion is waiting for a successful reload.

## Effective views

`GET /api/users` and `GET /api/users/{username}` show declared and runtime users with their `origin`; `POST /api/users/{username}/token` resolves the user and the optional `client_id` like any login; the persona picker lists runtime users (moved from `ConfigManager` to `IdentityResolver`); the web UI lists them with a `runtime` badge and no edit or delete, since their lifecycle is `/api/runtime`'s; the dashboard counts them separately. `GET /api/config` and the MCP server stay on the declared configuration.

## Verification

- `tests/test_runtime_api.py` (32) and the reworked effective-views tests: flows for created identities with the declared files byte-identical, reads, delete, reset counts, six invalid payloads and no echo of a rejected secret, `409` for declared and existing names, promotion into both files, a name declared in the file but not reloaded (`409`, object kept), delete and second promotion during a promotion, a failing mirror hook, a failed reload then retirement as promoted, an abandoned promotion, a strict plugin failing on the post-write reload, a reset waiting for a promotion and keeping a written-marked object, a concurrent declaration ending as a collision, a malformed file as `write_failed`, the reconciliation audit, and the gate. 2318 unit tests, ruff, mypy, lint-imports.
- **Mutation run**: each of the 13 promotion and reconciliation rules disabled in turn fails at least one test. The seven tests written for the review findings fail on the pre-review head.
- **Real server**: create, use, reload, reconciliation with its audit event and warning, promotion with a broken `settings.yaml` (`500`, delete refused, then retired as promoted on the repaired reload), zero tracebacks.
- **E2E**: the agent gains a runtime identity scenario (66/66) covering creation, a password grant with the runtime client, `origin` in `/api/users`, surviving a reload, the authorization code flow, SAML SSO, a declared name answering `409`, a UI declaration removing a runtime object with its audit event, promotion with exactly one `promoted` event, and a reset. The whole `http-e2e` job, run step by step from the workflow file, exits 0 on every suite. Two items of the issue's e2e list are not reachable over HTTP and stay covered by the unit tests: the declared files staying byte-identical, and a name declared in the file but not yet reloaded.
- Docs: a new guide, "Disposable test identities"; the endpoints reference; `/api/runtime` named in the security guide and in both network-binding warnings; `architecture.md`.


cdelmonte-zg:
Round 2 (201e28a), from CodeQL on the first head.

`py/stack-trace-exposure` on the runtime routes' error helper: four flows carried a caught exception's `str()` into the JSON body, the two payload validations and the two creation collisions.

- The validation message stays, since it is the point of the endpoint, but it travels explicitly: `config_documents` raises `EntryInvalid` carrying the text it composed from the model's finding, and the route answers `exc.message`, as `/api/config/reload` already does with `ConfigurationRejected`.
- Every other case builds its own sentence from the kind, the name and the file. The writer's conflict and lock errors, and any unexpected write failure, no longer quote the library's text; that goes to the log instead.

2318 unit tests, ruff, mypy, lint-imports; the runtime API and identity tests unchanged and green.

cdelmonte-zg:
Round 3 (e1769fc), from a review of the first head.

**The blocker.** `YamlWriter._atomic_write` replaces the file first and notifies the mirror and reloads after it, so an unexpected failure in those steps reached the promotion as a plain exception: the mark was discarded as if nothing had been written, the response said the entry could not be written, and the caller could delete a runtime object whose entry was already in the file. The next reload would then record it as a collision instead of the promotion it was.

The writer now states which side of the replace failed: everything after it that callers do not already classify (a strict `on_config_saved` hook, a rejected configuration) is wrapped in `PostWriteError`. The promotion marks the object as written for that error too and answers `reload_failed`, so a later successful load retires it as promoted, or abandons it if the entry is gone by then; a failure before the replace still answers `write_failed`, with the object untouched and unmarked. The new test fails when only that wrapping is removed.

**The smaller point** was already in round 2 (201e28a, pushed before this review): the generic `write_failed` no longer returns the exception's text, which goes to the log; that round also answered CodeQL's `py/stack-trace-exposure` on the same helper.

2319 unit tests, ruff, mypy, lint-imports; the whole `http-e2e` job run step by step from the workflow file, all suites exit 0 (66/66); `mcp_smoke_test` green.

PR #365, opened 2026-09-16
refactor: MCP tool arguments take their value shape from the domain models (#297)

### LINKED ISSUE #297: refactor: derive MCP field schemas from the domain models (shape only), and stop under-reporting domain constraints
*Rewritten 2026-09-16 after a census of the 26 tools and their 106 properties; the original text (in the edit history) assumed duplicated descriptions and drifting types, which the numbers do not show.*

**What the census found.**

| | |
|---|---|
| Types, on the 89 properties that map 1:1 to a domain field | 89 agree, none drifted |
| Enums | `token_endpoint_auth_method` and `layout` duplicated verbatim in two tools each, and equal; `login_mode` and `saml_c14n_algorithm` declared in the tool schema while the domain model is a plain `str` |
| Constraints | 30 properties carry one on either side; 3 agree (`description`, `maxLength` 200), and **27 are constraints the domain enforces and the tool schema does not advertise**: `minLength: 1` on every name and secret, the `^#[0-9a-fA-F]{6}$` pattern of the three branding colours, and `0 < token_expiry_minutes <= 1440` |
| Descriptions | 5 identical, 13 near-identical, 75 different on purpose: the model describes the field ("User email address"), the tool describes the argument ("New email (optional)") |

So the problem is not the one this issue was opened for. It is:

> **The published MCP contract is more permissive than the domain contract that is then enforced.** A client cannot know before the call that a colour must be hexadecimal, or that `token_expiry_minutes` has a range; it learns from the tool's answer.

**Scope: derive the value shape, not the tool schema.**

```text
domain model                      tool definition
  type                              description   (argument-specific)
  enum                              requiredness  (operation-specific)
  minimum / maximum                 tool semantics (create / update / ...)
  minLength / maxLength
  pattern
  items
        \                            /
         MCP input_schema property
```

- The domain model is authoritative for the **shape of the value**; the tool stays authoritative for its description and for what the operation requires. `required` is not derived: `update_user` takes a partial patch of a model whose full representation demands more.
- Hand-written schema stays for tool-only arguments (`admin_secret`, `expected_*_revision`, `limit`, `event_type`, the `generate_token` lists) and for anything a tool means differently from the field.
- Two closed sets become closed in the domain too, so the constraint is visible in the type rather than in a validator or only in the MCP schema: `login_mode` (`password` | `persona`, today a `str` with a validator) and `saml_c14n_algorithm` (`exc_c14n` | `c14n` | `c14n11`, today a `str` where `routes/saml.py` silently falls back to Exclusive C14N for an unknown value, so a typo changes the algorithm without a word).
- A parity test covers the properties declared 1:1 and compares the canonical shape only: type, enum, minimum/maximum, minLength/maxLength, pattern, items. Not descriptions, titles, examples or requiredness, which the census shows are legitimately different.

**Accepted consequence.** The schemas become stricter, so some arguments the domain rejects today (a tool result with `success: false`) will be rejected by the schema instead (`MCP_INVALID_ARGUMENTS`, the dispatch layer of the error model in #287). That is validation moving to the right boundary; it is a change to the MCP error surface, so it belongs in the CHANGELOG and in the MCP reference next to the error table.

**Why before #190.** RFC 7591 adds a ninth client shape. On a derivation it inherits type, enum, pattern and range from the model; hand-written, it is a tenth copy to keep in step.

cdelmonte-zg:
Closes #297.

## What the census found

Before writing any code I compared all 26 tool schemas against the models they describe (106 properties):

| dimension | result |
| --- | --- |
| types | 89 of 89 agree, no drift |
| enums | 4 duplicated and equal, 2 declared only in the tool schema |
| constraints | 30 properties carry one; 3 agree, 27 are enforced by the model and advertised nowhere |
| descriptions | 5 identical, 13 near, 75 different on purpose |

So the duplication was real but harmless in the direction the issue assumed, and harmful in the other one: the published contract was more permissive than the enforced one.

## What changes

`mcp_server/field_schemas.py` derives a **value shape** from a model field: `type`, `enum`, `items`, `minimum`/`maximum`, `exclusiveMinimum`/`exclusiveMaximum`, `minLength`/`maxLength`, `pattern`. Nothing else. The input schema itself is still written per tool, because `required` is the operation's (`update_user` patches a model whose full representation demands more) and the description is the argument's, not the field's.

83 properties in `schemas.py` now read `_domain(User, "username", "Username to look up")`; shared blocks such as the common user fields put them in 92 argument slots across the tools.

Ten arguments are exceptions to a pure derivation: **nine carry an explicit override** because the tool accepts more than the field does (an empty `client_secret` on `create_client`/`update_client` means "no secret", an empty `background_color`/`header_color`/`footer_color` on both means "clear it", an empty `get_audit_log` `username` means "no filter"), and **one stays hand-written** because it is not a field's value at all - `verify_token`'s `audience` is the audience the caller wants a token tested against, not the IdP's own.

**Measured delta against main: 16 constraints added, 0 removed, 1 enum reordered.** The 16 are `minLength` on the names and secrets without a wider vocabulary, plus the range of `token_expiry_minutes`.

## One of them closes a real hole

`_tool_update_settings` writes its arguments onto `Settings` with `setattr`, and `Settings` has no `validate_assignment`, so its tool schema was the only check between an argument and the running configuration. That schema carried two enums and no bounds. Measured on `main`:

```
update_settings(token_expiry_minutes=0)     -> success=True,  effective value 0
update_settings(token_expiry_minutes=99999) -> success=True,  effective value 99999
```

Both are refused before dispatch now, and the setting keeps its value. The handler's comment used to justify the gap ("the schema declares the enum for `login_mode`"); it now states the general rule.

For the other rejections this is a visible change for a client: they move from the tool's answer to `MCP_INVALID_ARGUMENTS`, the dispatch layer of the two-layer error model (#287). What the tool answered varied more than I first wrote - `delete_user` reported `{"success": false}`, `get_user` `{"found": false}`, `create_user` returned the raw pydantic text - so the docs now say that rather than claiming one shape.

## Two closed sets move into the domain

`login_mode` and `saml_c14n_algorithm` were plain `str` in the models and `enum` only in the tool schemas. They become `Literal` in `models.py` and in `config_documents.py`, written once and shared by both.

This fixes a real defect: a typo in `saml.c14n_algorithm` used to pass the load and reach `routes/saml.py`, which silently fell back to Exclusive C14N. It now fails the load.

Closing the set needed two guards, both added after the review:

- **Blank keeps meaning "the default".** `config.py` expands `${VAR}` before the document is validated and an unset variable expands to `""`, so `c14n_algorithm: ${SAML_C14N}` used to load and would otherwise now abort. `normalize_saml_c14n_algorithm` is the same treatment `roles_attr_name`/`groups_attr_name` already get.
- **The writer checks the value before it replaces the file.** `_atomic_write` replaces `settings.yaml` first and reloads after, so an unchecked typo left behind a file the next process could not load at all - reproduced through `update_saml_settings` and through `POST /settings`. `validate_saml_c14n_algorithm` is the same pre-write guard `login_mode` already had, and the `saml.c14n_algorithm` row becomes `omit_when_falsy` so a blank form field clears the key instead of persisting `""`.

(`login.mode` was already rejected by `Settings.validate_login_mode` on `main`; it only moves into the document type here. The earlier draft of this text credited the PR with more than it does.)

## Tests

`tests/test_mcp_field_schemas.py`, 24 tests: every derived property still matches its model; the description stays the tool's; `required` stays the operation's; the new constraints are published; the overrides are kept; the closed sets are rejected in the model and in the document; a blank algorithm still resolves to the default; `field_shape` refuses a field it cannot describe (a union of value types, or nothing publishable) instead of quietly publishing an unconstrained argument. A property named like a domain field that is neither derived nor declared tool-only fails the tripwire, and a derived one must be linked to the field it is named after - the parity test compares a derivation with itself, so it cannot see `"username": _domain(User, "email", ...)`, and the names are what closes that gap. `RENAMED_DERIVATIONS` is the explicit place for a future derivation that deliberately does not follow the name. Together these are what should catch #190 when it adds a ninth client shape.

`tests/test_settings_plumbing_parity.py` covers the writer guard (both entry points, file untouched) and the blank-clears-the-key contract; `tests/test_mcp.py` covers the out-of-range `update_settings` refusal.

Each new rule was checked by mutation: disabling it makes exactly one test fail. Full suite 2348 passed, ruff/mypy/import-linter clean, the book builds, `docs/schema/config.v1.json` regenerates unchanged. Verified live on a running server (the SAML settings form: typo refused and file untouched, `c14n11` persisted and effective, blank clears the key and the process reloads) and over MCP stdio (`get_audit_log` and `verify_token` accept `""`, `get_user` with `""` comes back as `MCP_INVALID_ARGUMENTS` with `isError`).


cdelmonte-zg:
Follow-up filed for the pre-existing family this PR did not widen to cover: #366 (blank `oauth.issuer`, `oauth.audience` and `server.rate_limit_token_endpoint` are written before the load that rejects them).

Issue #404, opened 2026-09-19
Runtime repository contract: entries with instance identity and one atomic decision primitive (prerequisite of #354)


cdelmonte-zg:
#354 is scoped as "`SqliteRuntimeStore` implementing the #235 contract and passing the same contract tests". That is not sufficient: the contract is five primitives, and the atomicity the callers rely on is not in it. It lives in process-local locks around compositions of those primitives. A SQLite backend that passes today's contract tests would be correct per call and wrong per operation as soon as two processes share the file.

This issue defines the contract that replaces those compositions. It was designed together with #405, because object identity changes its shape. #408 (closes #403) is the interim, single-process answer to one of the compositions; its 20 tests are part of the acceptance here.

## The contract today

`RuntimeRepository` (`services/runtime_identities.py`): `create`, `get`, `list`, `delete`, `delete_all`. Its docstring states the rule: "A change is a delete and a create."

## Census: operations composed above the contract

Each row is an operation that must be atomic and is made so by a `threading` lock that a second process does not share.

| Owner | Operation | Composition | Lock |
| --- | --- | --- | --- |
| `authorization_transactions.py` | `create` | prune expired, count against `MAX_PENDING_TRANSACTIONS` (1000), create | `_store_lock` |
| | `mark_primary_verified`, `reset_login` | read, check state, `_replace` = delete + create | `_store_lock` |
| | `consume` | read, check state and username, delete, exactly once | `_store_lock` |
| | every read | taken under the lock because `_replace` has a gap where a live transaction is absent | `_store_lock` |
| `pending_second_factors.py` | `create`, `consume`, `discard` | same shapes: capped create (1000), read-check-delete | `_store_lock` |
| `client_metadata.py` | `remember` | sweep, read old entry, delete, evict oldest unprotected (cap 100, refuse when all are protected), create carrying `protected_until` over | `cache_lock` |
| | `retain_until` | read, delete, create with a later `protected_until` | `cache_lock` |
| | `cached_client` | read, and delete when expired, re-read under the lock | `cache_lock` |
| `routes/registration.py` | `register` | sweep, count against `max_clients`, create client, create record | the lifecycle scope of #408 |
| `services/identities.py` | `create_runtime_user/client` | check the declared configuration, create | `holding_loads()` |
| | `reset_runtime_identities` | read the promotion marks, delete all but the marked | `holding_loads()` + `_promoting_lock` |
[...]

Issue #363, opened 2026-09-15
Protocol state behind one runtime boundary: authorization codes, device codes, revocations and the audit


cdelmonte-zg:
Step 1 of 3 of the runtime state chain **#363 -> #354 -> #193** ([VISION](https://github.com/cdelmonte-zg/nanoidp/blob/main/VISION.md#what-comes-next)). Phase 2 of #235: the rest of the execution state moves behind the boundary that holds runtime users and clients, so that #354 can add a backend behind one boundary instead of five singletons, and #193's separate process can see what the IdP sees, the audit included.

The contract it moves onto is #404's (#409 to #412): entries with an instance identity, holds, one atomic decision primitive. **The census below changes what this issue asks for.** It is not "put four stores into `RuntimeRepository`". It is: bring four shapes of state behind **one runtime boundary**, using the repository abstraction only where the shape of the data earns it.

## Census

On `main` at 2158505. Four stores, four shapes.

| Store | Shape | What decides the design |
| --- | --- | --- |
| `services/auth_code.py` | one dict, code -> dataclass with datetimes and `amr` as a tuple; no cap | `consume_code` is a decision on one record whose predicate is client, redirect URI and PKCE. It marks the code `used` and keeps it; a second redemption deletes it; a mismatch does not burn it. `get_code_info` hands out the stored object by reference (test and debug reads only). |
| `services/device_code.py` | **two maps**: device_code -> grant, user_code -> device_code; cap 10,000, pruned on create | `verify(user_code)` finds the device_code inside the atomic step. `poll` calls `get_user` inside the lock. An expired grant is **reported** (`expired_token`), not treated as absent. A `user_code` collision silently overwrites the older mapping today. |
| `services/revocation.py` | two dicts, id -> expiry; `float("inf")` for a verified token without `exp`; extension is monotonic | every mutating call sweeps everything; `check_and_claim_refresh` touches both dicts in one step; `is_revoked` is a lock-free hot path and does no cleanup |
| `services/audit.py` | **not objects by name**: a bounded append-only deque of 1000, and counters | logging and `on_audit_event` hooks run after the append |

Callers: 13 in `routes/oauth.py`, 4 in `routes/oauth_grants.py`. The MCP tools read the audit log and its stats, of their own process, which is exactly what #193 cannot live with.

### Measured

| | |
| --- | --- |
| creating a device code with 10,000 pending, today | 0.17 ms |
[...]

PR #415, opened 2026-09-20
refactor: device codes live in the runtime store (#363, step 3)

cdelmonte-zg:
Step 3 of 5 of #363 (it does not close the issue). Device codes move behind the runtime boundary, on the contract of #404 and the envelope expiry of #413. The hardest of the five: two collections that must agree, and a poll that needs a user from somewhere a decision may not reach.

## What

`DeviceCodeStore` kept two dictionaries and a lock. It is now a view, with no state, over two repositories the runtime store lends: the grants, by device code, and an index from the user's code to the grant, which names the grant's **instance**. `get_device_code_store()` returns a view each time; the module singleton and its reset in `conftest.py` are gone. The call sites do not change.

There is no lock around the two collections and no transaction across them. They agree the way dynamic registration does since #411: instance identity, conditional operations, a compensation and a postcondition.

```text
create   grant, under the cap, dropping what is past its time     one decision
         index entry naming the grant's instance                  one decision
           user code taken  -> withdraw the grant, another pair
         look back at the grant
           gone             -> withdraw the index entry, another pair
lookup   index -> grant, and the grant must be the instance named; otherwise nothing
verify   lookup, then one decision on the grant, which must still be that instance
poll     classify (one decision) -> resolve the user OUTSIDE -> claim that instance
```

A decision reaches no other repository and no service, so the poll cannot look the user up while it holds the grant, as it did under the old lock. It classifies, has the user resolved outside, and then claims with a `consume` whose condition is: the same instance, still authorized, for the same user, still in time. A grant that changed in between is classified again instead of its successor being consumed. A user who cannot be found still costs nothing: the grant stays as it was. That condition is one function, `_is_the_one_seen`, and everything the poll concludes after the lookup goes by it, the claim and the missing-user answer alike.

A grant past its time answers `expired_token`, to the device and to the user alike, for as long as it is there: expiry on the envelope means removable by a cleanup, never invisible.

## Deliberate corrections, each pinned

- **A user code that is already taken is refused and another pair is made.** The second grant used to take the code over in silence and leave the first device's user approving somebody else's device.
- **When no pair can be made, the caller is told to come back.** `create()` retries on a device code that is taken as it does for a user code, and after a few attempts raises `DeviceCodeStoreBusy`, which is a `DeviceCodeStoreFull`: `/device_authorization` answers the same plain `503` with `Retry-After`, and the message and the audit entry say which of the two it was. The first commit ended in a bare `RuntimeError`, a `500`. Found in review.
- **A poll that keeps losing its claim answers "pending", not "not found".** The latter is false, and final: a device stops polling on `invalid_grant`. Found in review.
- **The two cleanups are of one moment.** Grants and user codes are cleaned in two decisions; at two readings of the clock, one on each side of a grant's time, the user code would go and the grant stay ("invalid code" to the user, "expired" to the device, about one authorization). One reading for the pair, and a full store still drops the user codes past their time. Found in review.
- **"User not found" is said of the grant that is there.** When the lookup found nobody, the poll answered `USER_NOT_FOUND` at once, of a grant that might by then be gone, run out, or replaced by a pending one under the same device code; the token endpoint makes a `500` of it. Under the old lock the lookup ran inside the critical section, so this window came with the step. The poll now confirms the grant it saw and otherwise looks again. Found in review of the PR.
- An expired grant is marked once, not rewritten by every poll.
- `DeviceCodeGrant.device_code` is keyword only and has no default: a grant kept under the empty name would be found by whoever polls for nothing. `nanoidp.services` exports the class, so the changelog says so; nothing outside the store constructs one.

## Looked at and left, with the reason in the code

- The user is whoever `get_user` found at that moment, and may be deleted a moment later. Nothing ever excluded that: the old lock was this store's, not the users', and the token is minted from the same snapshot after `poll()` has returned.
- A grant about to be withdrawn counts towards the cap while it is there, so a store one short of full can refuse a creation during a user-code clash. That is the transient of a saga with a compensation; it costs one early "come back". Creating the index entry first was considered and is the more complicated of the two.

## Tests

- `tests/test_runtime_services_atomicity.py::TestDeviceCodes`: claimed once under a race, a grant that changed or ran out under the poll, a user who is gone, a missing user with the grant replaced by a pending one (`PENDING`), deleted (`NOT_FOUND`) or run out (`EXPIRED`) during the lookup, a user code whose grant is gone or whose device code changed hands, a taken user code, a taken device code, no pair possible, a vanished grant, the cap under concurrency, the two cleanups, expiry for device and user, the shape rule, plain JSON, two views being one store.
- `tests/test_device_flow_complete.py`: the endpoint's `503` for a store that cannot make a pair, with its message and audit reason. Two tests that reached into the old dictionaries now move a grant's time through the repository.
- 35 hand-written mutants, 32 killed on three consecutive runs. The three that survive are the instance, status and user conditions of the claim taken one at a time: each is masked by the others, and removed in pairs they are killed. All three are kept, and the code says why.
- Full suite 3594 passed, 5 skipped, identical with every decision run twice; ruff, mypy, import-linter clean.
- Live, on the head commit: `e2e/test_agent.py` 70/70. By hand, with every user code forced to clash: the second request answers `503` with `Retry-After: 5` and its own message, the first grant still answers `authorization_pending`, a garbage device code `invalid_grant`, no traceback in the debug log.

Measured with ten thousand pending: creating one takes 0.81 ms (0.17 ms before: two repositories and their cleanups instead of one dictionary), a poll 4 us.


cdelmonte-zg:
Thank you, the race is real and it came with this step: under the old lock the user lookup ran inside the critical section, so a missing user was always said of the grant that was there.

Reproduced first, with three deterministic tests that failed on `165be47` (all three answered `USER_NOT_FOUND`):

```text
during get_user("alice") -> None
  A deleted, pending B under the same device code   -> PENDING
  A deleted                                         -> NOT_FOUND
  A run out                                         -> EXPIRED
```

Fixed in `f4a2fb1`, as suggested: the claim's condition is now one function, `_is_the_one_seen(current, seen)` (that instance, still authorized, for that user, still in time), asked by the `consume` and by the missing-user branch alike. When it no longer holds, the poll classifies again; the attempts are bounded by `_POLL_ATTEMPTS` and running out of them still answers `PENDING`.

Two mutants added (the confirmation skipped; a grant that is gone counted as confirmed), both killed. 32 of 35 on three consecutive runs; the three survivors are unchanged, the claim's conditions masking one another, killed in pairs. Full suite 3594 passed, identical with every decision run twice.

On the constructor: `services` is internal and nothing outside the store constructs a `DeviceCodeGrant`, but it is exported, so the changelog entry now says that `device_code` is a required keyword.

Not verified live: the window needs a hook inside the process between the two decisions, so it is pinned by the tests above rather than probed over HTTP.

Issue #420, opened 2026-09-21
Generated signing keys are not coherent across local processes, and a rotation invalidates nanoidp's own earlier tokens


cdelmonte-zg:
Several nanoidp processes on one host are about to become a supported shape (#354, then #193). The signing keys nanoidp generates for itself are process state that happens to be written to a directory: each `CryptoService` loads them once and never looks again, and the files are written one after the other with no exclusivity. Measured on main (`6a78d09`), three things follow. The first two only show with more than one process; the third shows with one, and belongs here because the fix for the second rests on it.

## Measured

**1. Two processes cold-starting on one empty keys directory end with different signing keys: 12 of 12.** `_ensure_keys()` is check-then-create (`if not (priv_path.exists() and ...)`), then writes the private key, the public key, the `kid`, the certificate and the metadata in sequence. Both processes find nothing, both generate, the second overwrites the first's files, and the first goes on signing with a key that is no longer on disk.

**2. After one process rotates, it and a running peer refuse each other's tokens, for as long as the peer runs.** `rotate_keys()` writes the same files in sequence and updates the `CryptoService` of the process that rotated, and nothing else:

```text
after rotation 1 by A: A signs with 5e1d0a67, B still signs with bfbeb079; B's JWKS has A's key: False
  a token A mints now: REFUSED by B (Signature verification failed)
  a token B mints now: REFUSED by A (Signature verification failed)
a process started now signs with A's key
```

Same after a second and a third rotation. B's JWKS never learns A's key, so a resource server that fetches the JWKS from B cannot verify a token minted by A either.

**3. In a single process, a token minted before a rotation is refused after it.** `verify_jwt()` decodes against `self.pub_pem`, the active key, whatever `kid` the token names. The previous key is kept and published in the JWKS (two keys after one rotation), and the documentation says it stays valid: "old key stays valid for verification" (`reference/mcp.md`), "Tokens signed with previous keys remain valid until those keys are rotated out" (`guides/SECURITY.md`). That holds for a resource server reading the JWKS; nanoidp's own endpoints refuse such a token. Through the HTTP surface, with a token pair from the password grant and then `POST /api/keys/rotate`: `/userinfo` goes from `200` to `401`, `/introspect` answers `active: false`, and the refresh grant answers `invalid_grant`. It is why B's token is refused by A in (2) even though A still holds B's key as a previous one.

[...]

PR #421, opened 2026-09-21
fix: generated signing keys are one bundle for every process that shares the directory (#420, part 1)

cdelmonte-zg:
Part 1 of 2 of #420 (it does not close the issue): the keys directory as state several processes share. Part 2 is a running peer noticing another process's rotation, in `get_crypto_service()`, and verification by `kid`. Both come before #354.

Four commits: `766534c` the change; `52f6f14` what a review of it found; `34d72c9` two cases found in review of this pull request; `648c366` what a review of those fixes found. Each fix was reproduced before it was made.

## Measured on main before this

- **Two processes cold-starting on one empty keys directory ended with different signing keys, 12 of 12.** `_ensure_keys()` was check-then-create, then five files written one after the other; the second process overwrote the first's files, and the first went on signing with a key that was no longer on disk.
- **A rotation wrote over the same fixed names**, so a process killed after the first file left `kid.txt` saying OLD over a private key that was NEW. "Files first, marker last" does not fix that: replacing the first live file already destroys the old bundle.

## What

`services/key_directory.py`: bytes in, bytes out; it knows nothing of JWT or SAML. The layout and the documented file names are unchanged.

- **One lock across processes**, held for a cold start, a rotation and a load, and for nothing else. It is `config_writer.directory_lock`, the advisory file lock the configuration directory already uses (POSIX and Windows, a bounded wait), with a thread lock of the keys directory's own, so that key I/O and configuration saves in one process do not wait for each other. The RSA key is generated before the lock is taken.
- **A cold start has one winner.** Under the lock, look again; whoever finds no bundle writes the whole of it and `kid.txt` last; the others load that bundle whole. A marker with no key behind it is not a bundle, and is removed before the files are written.
- **A rotation is a small journaled transaction, and replacing `kid.txt` is its commit point.**

  ```text
  rollback of OLD written down        the bytes of the live files as they are
  journal published                   until then nothing live is touched
  previous keys NEW needs, added      none deleted yet: OLD still refers to its own
  live files replaced                 private, public, certificate, keys.json
  kid.txt replaced                    <- commit
  previous keys nothing refers to, removed; journal removed first and alone; the rest
  ```

  Whoever next takes the lock recovers: a journal with the marker still OLD means not committed, and OLD is restored; with the marker NEW it means committed, and the leftovers go. **A restore copies from the rollback and never moves it**, so a recovery that dies is recovered again. There is no cleanup on the way out of a failure, on purpose: a process that is killed runs none.
- **A rotation retires the key that is published**, not the one the rotating service remembers, so a rotation on top of a peer's orphans nothing. On a directory emptied under a running service it retires the service's own key, as it always did.
- **A directory this process cannot write still boots**, as documented (a read-only mount; a volume created by the image that ran as root). It has no lock to take and nothing it could repair: it loads what is published without the lock, checking that the marker did not move and that no rotation is under way. One that does not end is an error that says what to do. A lock that is there and busy is never read around.
- `POST /api/keys/rotate`: `409` for a directory this process can never write, `503` with `Retry-After` when the lock cannot be had in time; the MCP `rotate_keys` tool answers `success: false`. A timeout names the keys directory, not the configuration.
- **The private key is written with mode `0600`** (it followed the umask). A write that fails takes its temporary file with it, and whoever takes the lock sweeps the ones a killed writer left.

## From the review (`52f6f14`)

Eight findings, all applied; two were reproduced defects of the first commit. The lock made a **load** need a writable directory, so a read-only keys volume no longer started: that is the read-only concession above, the one `config_store` already makes. And `shutil.rmtree` on `.rotation/` could remove the rollback before the journal, after which every start failed on a journal pointing at nothing: the journal now goes first and alone, with a checkpoint and a death exactly there. One correction of prose: the first commit said the keys needed their own thread lock because a load would otherwise wait for itself. That is not true of today's code (the whole suite passes with the shared lock); the lock stays, for the reason given above.

## From the review of this pull request (`34d72c9`, `648c366`)

- **Read-only with the lock file there.** A lock file that exists can be taken through a read-only view, and a directory NanoIDP has used has one; the first test had removed it. Every mutation now finds out by writing a probe under the lock whether it can write at all, and says `KeysDirectoryNotWritable` (a `LockNamespaceUnavailable`, `409`), told by `errno` (`EACCES`, `EPERM`, and `EROFS`, which is not a `PermissionError`).
- **A certificate two processes repair together**: made outside the lock, installed after looking again under it; a peer's is adopted if it belongs to the signing key. The marker does not move for a certificate, so nothing would ever have realigned them. Pinned with a seam: the real-process test passes on the defective code too, the window being a millisecond against a lock asked for again after 50.
- **What is read without the lock is one bundle**: the marker and the journal are looked at before the read and after it and have to agree; a recovery that ended meanwhile used to leave a NEW private key next to an OLD public key.
- **A published key that cannot be read is an error, never "no keys"**, which with the key now `0600` would have started cold over another user's bundle.
- **Nothing fails a rotation past its commit**; a committed rotation that was not tidied up does not keep a read-only process out; a peer that rotated while a starting service made its certificate is followed; the certificate of external keys is installed under the lock too; one atomic bytes writer, `serialization.atomic_write_bytes`.

## Tests

- `tests/test_key_directory.py`, 67 tests. **Real processes** (spawn): four starting together on an empty directory, four rounds, end with one `kid`, one private key, equal to what is on disk, none raising; two rotating together both succeed, and the second retires the first's key. **A death at every step**: each checkpoint of a cold start (with and without a stale marker), each step of a rotation before the commit (the old bundle byte for byte, its previous keys too, with `max_previous_keys=1` so that the key the rotation would drop must still be there), each after it, each step of a recovery, and the recovery's own tidying up. **A process really killed**, `os._exit` in a subprocess, before and after the commit. A directory from before the metadata existed; temporary files; a failed write; an emptied directory; a read-only directory (loads, refuses a rotation, refuses a rotation that never ends); a busy lock at the service, the endpoint and the MCP tool.
- 44 hand-written mutants, 44 killed on three consecutive runs; across the rounds nine survived a first run, each a test of mine that did not reach the branch it was written for, and each was closed. One of my own tests found a defect of the protocol as first written: a rolled-back rotation left the previous-key file it had added.
- Full suite 3793 passed, 5 skipped, identical with every decision run twice; ruff, mypy, import-linter clean.
- Live: two real servers started together on an empty directory serve one `kid`, and a token minted by one is `active` at the other; a rotation leaves no work directory; a server on a directory of mode `555` boots, issues tokens, answers `409` to a rotation and touches no file. `e2e/test_agent.py` 70/70. After A's rotation B still serves the old key: that is part 2.


cdelmonte-zg:
Thank you. Both cases are real, both were reproduced before anything was changed, and both are fixed on this PR. A `/code-review high` of the fixes then found nine more things, all applied; two of them could destroy a published bundle.

**1. Read-only with `.nanoidp-write.lock` there.** The test did remove the lock file, which selects the easy case. With it there the lock is had, the rotation begins, and a `PermissionError` comes out as a `500`. Every mutation (cold start, rotation, certificate) now first finds out **by writing a probe under the lock** whether it can write at all, and raises `KeysDirectoryNotWritable`, a `LockNamespaceUnavailable`: `409`. `os.access` was not used: it answers for the mode bits and not for a read-only mount. A recovery that cannot write what it has to settle says the same, and a start then reads around it as it does around a missing lock. Three twin tests leave the lock file there.

**2. Concurrent repair of `idp-cert.pem`.** Reproduced deterministically, with a seam on the certificate's making: the real-process test passes even on the defective code, because the window is about a millisecond against a lock that is asked for again after 50. The candidate is still made outside the lock and installed after looking again under it: a certificate a peer installed meanwhile is adopted if it belongs to the signing key, replaced if it does not. One function, `_install_certificate`, now serves the generated keys and the external keys' certificate alike.

**From the second review, the ones that matter most:**

- **A mixed bundle read without the lock.** The unlocked read looked at the journal only after the files. A rotation that had died was there when the reading began, a peer settled it before the reading ended: no journal afterwards, the marker as it was, and a NEW private key next to an OLD public key in hand. The journal is now looked at before the read as well, and the two observations have to agree.
- **A published key that cannot be read was taken for "no keys".** `load()` mapped every `OSError` to `None`, so with the key now `0600` another user who could write the directory would have started cold over somebody's bundle. Only `FileNotFoundError` means "not a bundle"; anything else is an error, and the bundle stays as it was.
- `EROFS` is a plain `OSError`, not a `PermissionError`: a volume mounted `:ro` escaped the classification that `chmod` tests exercise. Classified by `errno` now (`EACCES`, `EPERM`, `EROFS`), and an `EIO` is not called "not writable".
- **Nothing fails a rotation past its commit.** A failure to tidy up was reported as a failed rotation, which the caller would have repeated over a key in use while this process went on signing with the one it had retired. What cannot be tidied up is left to whoever next takes the lock.
- A rotation committed and not tidied up no longer keeps a read-only process out: once the marker names the journal's NEW kid the commit point is past and every live file is NEW's.
- A peer that rotated while a starting service was making its certificate is followed: the service adopts the published bundle instead of finishing its start on the retired key.
- The certificate of external keys is written under the lock, since the lock's owner sweeps temporary files in that directory.
- One atomic bytes writer, `serialization.atomic_write_bytes`, next to `atomic_write_text`, instead of a third copy.

44 hand-written mutants, 44 killed on three consecutive runs (three survived their first run: tests of mine that did not reach the branch they were written for, corrected). Full suite green in both modes. Live, a server on a `555` directory that has its lock file boots, issues tokens and answers `409` to a rotation.

The journal protocol and the part 1 / part 2 split are unchanged.

PR #422, opened 2026-09-21
fix: a token is verified against the key its kid names, and a process notices a peer's key rotation (#420, part 2)

### LINKED ISSUE #420: Generated signing keys are not coherent across local processes, and a rotation invalidates nanoidp's own earlier tokens

cdelmonte-zg:
Part 2 of 2 of #420. Part 1 (#421) made the keys directory one bundle for every process that shares it; this makes the service that signs and verifies follow it, and verify by `kid`.

Closes #420.

Four commits: `9fe24f0` the change, `3ce0d79` what a review of it found, `32d0d21` what a second review found, `087502b` a race found in review of this pull request. Each fix was reproduced before it was made.

## Measured on main before #420

- **In a single process, a token minted before a rotation was refused after it.** `verify_jwt` decoded against the active key whatever `kid` the token named. With a token pair from the password grant and then `POST /api/keys/rotate`: `/userinfo` went from `200` to `401`, `/introspect` to `active: false`, the refresh grant to `invalid_grant`. Meanwhile the JWKS went on publishing the previous key, and the documentation said "old key stays valid for verification".
- **After one process rotated, it and a running peer refused each other's tokens for as long as the peer ran**, and the peer's JWKS never showed the new key.

## What

**Verification is by `kid`**, among the active key and the previous ones kept, no wider than `jwt.max_previous_keys`, so the same set the JWKS shows:

```text
no kid                     the active key, as it has always been
the active kid             the active key
a previous kid kept        that key
anything else              invalid token
```

The `kid` chooses the key and vouches for nothing: a token signed by a stranger under a `kid` this issuer keeps fails on its signature. **External keys have no history**: they verify with their one key whatever `kid` a token names, as before, so changing `jwt.external_key_id` for the same pair invalidates nothing.

**The keys of a service are one value, replaced whole.** `SigningKeys` (frozen) holds the kid, the private key, the public key, the certificate and the previous keys; a rotation or the adoption of a peer's assigns a new one. Before, a rotation set those one after the other on the service every request was using, and a token minted between two of the assignments named one key and was signed with the other. That was harmless while every token was checked against the active key, and becomes a refusal for good once verification is by `kid`. Minting, verifying, the JWKS, SAML signing, the keys page and the keys information each read the keys once (`crypto.keys` for a reader that needs more than one of them).

**A peer's rotation is noticed in one place, `get_crypto_service()`**, through which JWT, JWKS, SAML (which reads the private key and the certificate directly) and the key information all come, so none has a freshness of its own.

- It looks at `kid.txt` without a lock: a `stat` first, a read only when the file is a new one.
- Only when the marker names another key is the bundle loaded, whole, under the directory's lock. **Loading and nothing else**: the constructor would generate a key pair where it finds none, and a marker with no key behind it would then have had the next request replace the signing key and every previous one. Such a marker is reported in the log and left.
- A **new service is published**; the one in hand is never changed, so a request that already holds it keeps a whole one. A signature made with the old key just before the peer's commit is correct, that key being a previous one now.
- The refresh has **a lock of its own**, so the lock services are published under is never held while the directory's is waited for. A request waits for another's refresh at most half a second, counted from when that refresh began, and not at all for one that is stuck. A refresh that failed is not tried again for five seconds; the key in hand is used meanwhile.
- **This process's own rotation** adopts its keys before it lets the directory go. A refresh that finds the service caught up, a newer service published, or the service's kid moved while the bundle was loading (its own rotation landed on top) publishes nothing.
- **What a refresh read is about the service it read it for.** If another service was published before the refresh's lock is had (a reload of the configuration: external keys, another keys directory), that service is returned as it is, and nothing is loaded for it.
- A marker that cannot be read (`EMFILE`, `EIO`, `EACCES`) is said in the log, not taken for "no rotation"; `active_kid()` is `None` only for a marker that is not there. A directory with no marker at all is not a rotation either: a running service goes on, and its next rotation puts the directory right (part 1).
- External keys are not looked for in the directory.

## From the reviews

- **`3ce0d79`**, seven findings, all applied:
  - the keys changed in place during a rotation (the snapshot above);
  - a failed refresh retried at every request under the publication lock (the refresh's own lock, the bounded wait, the retry interval);
  - a refresh running the constructor (loading only);
  - this process's own rotation seen as a peer's (adopting inside the lock);
  - the marker read and trusted at every call (the `stat`, the errors said);
  - external keys refused under a changed `kid` (their one key);
  - a real-peer fixture that could hang the run (a daemon, stopped whatever happens, reporting its exceptions).
- **`087502b`**, from review of this pull request: a marker read for the service in hand was used for another service published meanwhile by a reload of the configuration, and the refresh loaded for the wrong directory (for external keys, the generated bundle next to them). Reproduced with two tests, external keys and generated keys elsewhere; both fail on `32d0d21`.
- **`32d0d21`**, one finding. A peer rotates; a request loads that bundle; before it publishes, this process rotates on top, which makes its own key the newest. The request then published the older bundle over it, until the next request put it right. The refresh now remembers the kid the service had when it began.

## Tests

- `tests/test_key_freshness.py`, 47 tests:
  - **Verification by kid:** earlier tokens after a rotation, until rotated out; the retention applied from the start; an unknown kid; the active key under a kid that is not its own; a stranger under a known kid; no kid and an empty kid; kids that are no text (signed by hand, since PyJWT will not write them); garbage; external keys under a changed kid.
  - **HTTP pins** at `/userinfo`, `/introspect` and the refresh grant, after one rotation and after `max_previous_keys + 1`.
  - **One reading:** a trap makes any read of a single field rotate the service right after it answers, so a second read answers for the other key. Minting, verifying and the JWKS pass under it.
  - **Noticing:**
    - nothing loaded, and not even the refresh's lock asked for, while the marker stays;
    - no lock taken to look (all three held while another thread looks);
    - a peer's rotation adopted whole;
    - a reader of the old service keeping a whole one;
    - the settings kept;
    - own rotation: no reload, keys in hand before the directory is let go, one landing during a refresh not undone;
    - a caught-up service publishes nothing;
    - an emptied directory, and a marker with no key behind it (no key made);
    - the marker read only for a new file; one that cannot be read, said;
    - external keys not looked for;
    - a failed refresh: not retried at every look, and nobody waits for a stuck one nor for the publication lock;
    - eight threads noticing together, one reload counted.
  - **A real peer process**, driven over queues, in both directions: it signs, verifies and publishes as this process does, and its SAML certificate is the same file.
- The tests of each review fail on the commit before it (11 failures and 3 errors on `9fe24f0`; the last one on `3ce0d79`).
- 39 hand-written mutants, 39 killed on three consecutive runs; across the rounds six survived a first run, each a test of mine that did not reach its branch or that measured only the result where the finding was a cost, and each was closed. The 44 mutants of part 1 are killed on this code as well.
- Full suite 3840 passed, 5 skipped, identical with every decision run twice; ruff, mypy, import-linter clean.
- **Live, on the head commit, two real servers on one keys directory.**
  - A rotates; B's JWKS equals A's.
  - New tokens are valid at both.
  - A token from before the rotation is valid at B: `/introspect` active, `/userinfo` `200`, refresh `200`.
  - The SAML certificate is the same bytes at A, at B and on disk.
  - After two more rotations (`max_previous_keys` 2) the first token is refused at both (`/userinfo` `401`), and the JWKS agree.
  - With the marker naming a key that is not there, B goes on answering, generates nothing and warns once.
  - No traceback. `e2e/test_agent.py` 70/70.


cdelmonte-zg:
Thank you, the race is real and it is fixed as suggested.

Reproduced first, with the two tests you asked for. A wrapper around the refresh's lock publishes another service right before the lock is had, and `reloaded` is made to fail the test if it is called. Both failed on `32d0d21` by loading for the wrong service:

```text
generated G sees a peer rotation, then external keys E are published   -> loaded for E's directory
generated G sees a peer rotation, then generated keys elsewhere        -> loaded for the other directory
```

Fixed right after the refresh's lock is taken:

```python
current = _crypto_service or service
if current is not service:
    return current
if current.kid == published:
    return current  # this process rotated meanwhile
```

The observation belongs to `service`, and as soon as the published service is another one it is no longer usable. `_fresh` is not called again under the refresh's lock; the new service is checked on its own terms at its next look. Both tests now return the other service unchanged (the same `keys` object), and nothing is loaded.

The first mutant run showed one more thing worth pinning: the `current.kid == published` branch, for a service that caught up before the refresh began (this process's own rotation adopts its keys inside the directory's lock), only saved a load, so removing it changed no result. It is now pinned by a test that fails if anything is loaded.

39 hand-written mutants, 39 killed on three consecutive runs; the two new ones are "an observation of one service is used for another" and "this process's own rotation is loaded again". Full suite green in both modes.

PR #423, opened 2026-09-21
feat: the runtime store is chosen by the configuration and activated with it (#354, step 1)

cdelmonte-zg:
Step 1 of 5 of #354 (it does not close the issue): the runtime store gets a contract and is chosen by the configuration, activated with it. No second backend yet: the SQLite store is step 2.

Three commits: `1f4ca3f` the step, `d2ea878` what a review of it found, and a third that corrects the prose after that fix (preparing does not activate; it is not that it touches nothing global), with no change of behaviour.

## Before

`get_runtime_store()` built an in-memory store on first use. It took no settings, and nobody published it. So no backend could ever be chosen, nothing said which store a process was using, and eight service modules named the in-memory backend in their annotations.

## What

**Configuration.** A `runtime:` section in `settings.yaml`, YAML-only like `jwt:`:

```yaml
runtime:
  store: memory     # the default, and in this step the one value
```

The schema names no backend NanoIDP does not have. `store` is `Literal["memory"]`, and the published schema artifact says `const: "memory"`. `sqlite` and `sqlite_path` arrive together with the backend in step 2, so no commit of `main` promises a store that is not there. A value that is not a store is an invalid file. An unknown key under `runtime:` is what any unknown key is (#175): a warning in `warn` mode, a refusal under `strict`.

**The contract.** `RuntimeStore` is a Protocol with `users`, `clients` and `audit` as read-only properties, plus `repository(name, key_of, codec)`. The services are annotated against `RuntimeStore`, `RuntimeRepository` and `AuditStore` only. An AST test pins that no consumer names a `Memory*` backend (the backend modules themselves are exempt).

**Activation**, next to the signing service (#359), through the step the `ConfigManager` already has:

```text
stage the directory, validate
activate_services(candidate)
    prepare_runtime_store(candidate)      activates nothing; may refuse
    prepare_crypto_service(candidate)
    return publish_both
configure hooks                           may still fail
publish the runtime store, then the signing service
assign the settings
after_load
```

- `services.activate_services` is what both the app and the MCP server now pass to `init_config`. `activate_crypto_service` is still there.
- The runtime store comes first: it is the cheaper of the two, and the one that can refuse outright.
- A failure in either publishes nothing. **Preparing does not activate**: only the publication records the inputs a store was chosen with (`_runtime_store_inputs`), so a load that fails leaves no configured choice behind. The one global effect of preparing: for memory, when nobody has asked for a store yet, it brings the provisional store into existence, the same effect any `get_runtime_store()` has, so that what is recorded between preparing and publishing lands in the store that is published (`d2ea878`). For another kind, the candidate is built aside and published by nobody.

**Restart required.** The store of the process and the inputs it was activated with are kept apart (`_runtime_store`, `_runtime_store_inputs`). Being the activated store is the lifecycle's business, not a property of a store. A reload whose inputs differ from the ones in force is refused with `RuntimeStoreRestartRequired` before anything is built for it. That surfaces as a `ConfigurationRejected` of kind `activation`, a `422` at the endpoint, with a message that says to restart, and none of the new configuration takes effect. The same inputs mean the same store.

**Before the first activation**, `get_runtime_store()` hands out a provisional memory store; the audit of a plugin's load hook is the case. The first activation that asks for memory adopts that very instance, so what was recorded before the configuration was read is kept. If nobody asked yet, preparing makes the provisional store and adopts it, so what is recorded between preparing and publishing is kept too (`d2ea878`).

**The getter never reads the configuration.** It is called while the configuration is being built. A `ConfigManager` built without `activate=` governs nothing, and whoever wants it to passes `activate_services`.

**Visibility.** `GET /api/config` and the MCP `get_settings` tool report `runtime.store`.

## Owed to step 2, written down

When the first activation asks for another backend than memory, the provisional store is replaced, which is no restart: nothing had chosen it. What was recorded in it before that is then not carried over. There is no second backend in this step, so nothing is lost today. Step 2 has to decide what happens to it before SQLite is usable.

## From the review (`d2ea878`)

One finding, reproduced. When nobody had asked for a store before the first load, preparing built one aside. An event recorded between preparing and publishing (a plugin's `configure()`, while the hooks are configured) made a provisional store that the publication then replaced. Fixed as described above, and pinned by a test that prepares, records and publishes.

## Tests

- `tests/test_runtime_store_activation.py`, 29 tests:
  - **Chosen by the configuration:** the default; the same said in the file; `/api/config` and MCP report it; the store used is the one activated.
  - **Schema:** a store the schema does not offer (`sqlite`, `redis`, empty, a bare string) is an invalid file; a key it does not offer follows #175.
  - **Other backends:** a backend named by the settings is the one built and published; a kind nanoidp has no store for is refused, never taken for memory.
  - **Preparing does not activate:** a prepared store is not published; an activation that fails after the store was prepared leaves it provisional.
  - **Restart required:** an unrelated reload keeps the store and its contents; other inputs are refused before anything is built; end to end, a reload that changes them is a `422` of kind `activation` and nothing of it applies; a required restart is said before a signing service is prepared; the activation's publication publishes both; the same inputs are the same store.
  - **Before the first activation:** a store for whoever asks; adopted by the first memory activation; what is recorded between preparing and publishing is kept; another backend replaces it and is no restart; the getter never reads the configuration; a manager without the activation step governs nothing; both compositions activate the store.
  - **Contracts:** no consumer names a backend; the activation state is no part of a store.
- A second backend in the tests is a private factory for a kind no file can name, with candidate settings made by `model_copy` (which does not validate).
- 23 hand-written mutants, 23 killed on three consecutive runs. One survived its first run: a kind nanoidp has not, built as memory. It is now pinned.
- Full suite 3869 passed, 5 skipped, identical with every decision run twice; ruff, mypy, import-linter clean. The published schema artifact was regenerated (`nanoidp config-schema --write`).
- **Live, on the head commit:**
  - `/api/config` reports the store, and a runtime client is created;
  - a reload asking for `sqlite` is refused (`invalid`) and the runtime client is still there;
  - an unknown `runtime.sqlite_path` is warned about and the reload succeeds;
  - a start with `store: sqlite` is refused with "runtime.store: Input should be 'memory'";
  - no traceback, and `e2e/test_agent.py` 70/70.


cdelmonte-zg:
Agreed, and done in `16fb908`, text only.

`prepare_runtime_store` now says what holds: preparing activates nothing, and only the publication records the inputs a store was chosen with, so a load that fails leaves no configured choice behind. For memory, when nobody has asked for a store yet, it may bring the provisional store into existence, which is the same effect any `get_runtime_store()` has. For another kind it builds a candidate aside that nobody publishes.

The same correction is in `activation.py`, in the test module's docstring, in the CHANGELOG entry and in the PR body. `TestPreparingChangesNothing` is now `TestPreparingDoesNotActivate`.

No line of logic changed; the full suite passes as before.

